PR #2, opened 2019-06-14
Add spec attributes

baijum:


pkg/apis/apps/v1alpha1/servicebinding_types.go
```diff
@@ -4,21 +4,21 @@ import (
...
-// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // ServiceBindingSpec defines the desired state of ServiceBinding
 // +k8s:openapi-gen=true
 type ServiceBindingSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+	CSVName          string `json:"csvName"`
```
baijum: The name will be something like this: `postgresql-operator.v0.1.0` (For more examples run `kubectl get csv --all-namespaces` in your OpenShift 4 cluster). You may also look at CSV files in this repo: https://github.com/operator-framework/community-operators

pkg/apis/apps/v1alpha1/servicebinding_types.go
```diff
@@ -4,21 +4,21 @@ import (
...
+	CSVNamespace     string `json:"csvNamespace"`
+	ApplicationLabel string `json:"ApplicationLabel"`
```
baijum: Do we need multiple labels? I mean should we make it `[]string` as type here?

Avni-Sharma: shall we make ApplicationLabel as applicationLabel, just to maintain the consistency like csvNamespace and csvName?

baijum: @Avni-Sharma Yes, I will change that. Thanks!

Avni-Sharma: We just want to bind to a backing service right. (No message queue or cache, I suppose). If it's only a db then I guess one label should suffice.

pkg/apis/apps/v1alpha1/servicebinding_types.go
baijum: How about using `BackingOperatorName`? Which one would be in-abmiguous?

DhritiShikhar: `CSVName` is ok I guess

baijum: CSV is the meta data for the operator. So, the CSVname is pointing to the operator name including the version number.

baijum:
Based on the discussion in the Slack, the CRD will be renamed to `ServiceBindingRequest`. I will do it in a separate PR.

pkg/apis/apps/v1alpha1/servicebinding_types.go
baijum: What if there are multiple applications? Create separate CRs?

baijum: After some thinking, I realize it's better to create separate CRs for different applications. Otherwise deployments will be re-deployed across for all labels.

baijum: Multiple labels can be used to qualify service binding request to narrow down the scope and target a specific application development. This is the reason why there are multiple labels for label selectors in Kubernetes. I think we can follow the same approach. For example:
```
connects-to: postgres
environment: production
```
Another could be like this:
```
connects-to: postgres
environment: stage
```

baijum: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors

baijum: How about something like this:
```
applicationSelector:
  matchLabels:
    connects-to: postgres
    environment: stage
```

baijum: How about introducing resource "kind" also in the application selector?

For Development kind:
```
applicationSelector:
  matchLabels:
    connects-to: postgres
    environment: stage
  kind: Deployment
```

For StatefulSet kind:
```
applicationSelector:
  matchLabels:
    connects-to: postgres
    environment: stage
  kind: StatefulSet
```

baijum: Instead of `kind`, probably we could qualify it further with a prefix like this: `resourceKind`. Or is it too verbose? ;-)

Avni-Sharma: I think so that resourceKind would cause less confusion because kind would be already there. Wdyt @baijum ?

baijum: Done. 71aec326ad6ab64d910eec26628b29e7e4826a4b

pkg/apis/apps/v1alpha1/servicebinding_types.go
baijum: Done. 71aec326ad6ab64d910eec26628b29e7e4826a4b

baijum:
> Based on the discussion in the Slack, the CRD will be renamed to `ServiceBindingRequest`. I will do it in a separate PR.

Renamed.

baijum:
@DhritiShikhar @Avni-Sharma @sbose78 
I want to start work on the controller. But this PR need to be merged before that. So, please take a look again.

PR #5, opened 2019-06-18
Create Makefile

pratikjagrut:
This PR adds Makefile with some basic lints and other useful targets.
[Jira task: ODC-1096](https://jira.coreos.com/browse/ODC-1096)

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,33 @@
+rules:
+  braces: enable
+  brackets: enable
+  colons: enable
+  commas: enable
+  comments:
+    level: warning
+  comments-indentation:
+    level: warning
+  document-end: disable
+  document-start: disable
+  empty-lines: enable
+  empty-values: enable
+  hyphens: enable
+  key-duplicates: enable
+  key-ordering: disable
+  line-length:
+    level: warning
+  new-line-at-end-of-file: disable
+  new-lines: enable
+  octal-values: enable
+  quoted-strings: disable
+  trailing-spaces: enable
+  truthy:
+    level: warning
+
+  # accept both     key:
+  #                   - item
+  #
+  # and             key:
+  #                 - item
+  indentation:
+    indent-sequences: whatever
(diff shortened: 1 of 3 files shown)


.yamllint
```diff
@@ -0,0 +1,33 @@
...
+  document-end: disable
+  document-start: disable
+  empty-lines: enable
+  empty-values: enable
+  hyphens: enable
+  key-duplicates: enable
+  key-ordering: disable
+  line-length:
+    level: warning
+  new-line-at-end-of-file: disable
```
Avni-Sharma: enable would be better I suppose

Makefile
```diff
@@ -0,0 +1,95 @@
...
+	  } \
+          else { \
+	    lastLine = $$0 \
+          } \
+        }' $(MAKEFILE_LIST)
+
+
+
+# By default the project should be build under GOPATH/src/github.com/<orgname>/<reponame>
+GO_PACKAGE_ORG_NAME ?= $(shell basename $$(dirname $$PWD))
```
Avni-Sharma: This won't be required as we are using go modules.

pratikjagrut: Its not standard go project path, if you look at it `goapth` is missing, we can add any path before it, it could be `gopath` or `dir_xyz` but for base or as default I took same path as pkg path in go. The variable GO_PACKAGE_PATH is currently not used anywhere but could be used for `Dockerfile` or for build location.

.yamllint
pratikjagrut: Enable is not required as as it will force for `\n` character at the end of file. '\n` character is must for files used by unix tools because of POSIX standards, and I don't think that our yaml files will be handled by any unix tool. But still I'll look into it to confirm.

Makefile
akashshinde: @Avni-Sharma Have we decided that we are going to use gomod for service-binding operator ?

.yamllint
baijum: In GitHub look at the end of this file. You can see a red icon with a circle and arrow. And if you mouse over, you can see a message: `No newline at end of file` I would prefer not to see a message like that.

Makefile
baijum: @akashshinde Yes, it is Go modules.

baijum:
@pratikjagrut Don't force push anything to your remote branch.

PR #6, opened 2019-06-18
Create environment variables from OLM descriptors

baijum:
Story: https://jira.coreos.com/browse/ODC-1057 
The backing operator cluster service version (CSV) resource is identified using the name of custom resource and optionally the version of the API. The CSV resource is used to extract the descriptors relevant for service binding. Descriptors which has URN starting with `urn:alm:descriptor:servicebinding:secret:` and `urn:alm:descriptor:servicebinding:configmap:` is considered as the relevant fields. The correspoding path value should be pointing to a Secret and ConfigMap objects respectively. Later, these objects is injected into application containers as environment variable.

The application container resources are identified using label selectors for the given kind of resource. In the absence resource kind, it will fall back to Deployment.

The Secret and ConfigMap resources are injected as envrironment variables in to the application containers.

Here is an example service binding request CR:

```
kind: ServiceBindingRequest
metadata:
spec:
  backingSelector:
    resourceName: database.example.org
    resourceVersion: v1alpha1
  applicationSelector:
    matchLabels:
      connects-to: postgres
      environment: production
    resourceKind: Deployment 
```

## TODO List

- [ ] Watch for deployments based on the label selectors


pkg/controller/servicebindingrequest/servicebindingrequest_controller.go
```diff
@@ -97,5 +100,47 @@ func (r *ReconcileServiceBindingRequest) Reconcile(request reconcile.Request) (r
...
+		return reconcile.Result{}, err
+	}
+
+	evList := []corev1.EnvVar{}
+
+	for _, crd := range csv.Spec.CustomResourceDefinitions.Owned {
+		for _, spec := range crd.SpecDescriptors {
+			pt := spec.Path
+			for _, xd := range spec.XDescriptors {
+				if strings.HasPrefix(xd, "urn:alm:descriptor:servicebinding:secret:") {
```
Avni-Sharma: Preferably it should  be servicebindingrequest instead of servicebinding

pkg/controller/servicebindingrequest/servicebindingrequest_controller_test.go
```diff
@@ -0,0 +1,90 @@
...
+			Namespace: namespace,
+		},
+		Spec: olmv1alpha1.ClusterServiceVersionSpec{
+			CustomResourceDefinitions: olmv1alpha1.CustomResourceDefinitions{
+				Owned: []olmv1alpha1.CRDDescription{
+					{
+						Name: "some name",
+						SpecDescriptors: []olmv1alpha1.SpecDescriptor{
+							{
+								XDescriptors: []string{"urn:alm:descriptor:servicebinding:secret:password", "aaa:ccc:aa"},
```
Avni-Sharma: servicebinding -> servicebindingrequest

pkg/controller/servicebindingrequest/servicebindingrequest_controller.go
baijum: Done. 414d93276e54cf11e225c11efbc514dc9dd067ac

pkg/controller/servicebindingrequest/servicebindingrequest_controller_test.go
baijum: Done. 414d93276e54cf11e225c11efbc514dc9dd067ac

PR #10, opened 2019-06-19
Clarify the API

baijum:
The resource details are looked up inside the CSV spec. For example, consider this CSV file:
https://github.com/operator-framework/community-operators/blob/0230c07c33ae36272151f918ad7c87d62ae13ee4/upstream-community-operators/etcd/etcdoperator.v0.9.4.clusterserviceversion.yaml#L232

Line no. 232 to 234 is what is used to identify the CSV.

### DIFF
--- a/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
+++ b/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
@@ -13,7 +13,6 @@ type ServiceBindingRequestSpec struct {
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
 
 	// BackingSelector is used to identify the backing service operator.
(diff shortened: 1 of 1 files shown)


pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -25,12 +24,12 @@ type ServiceBindingRequestSpec struct {
...
 	//	backingSelector:
 	//		resourceKind: Database
-	//		resourceName: my-db
+	//		resourceName: database.example.org
 	// Example 2:
 	//	backingSelector:
 	//		resourceKind: Database
-	//		resourceName: my-db
-	//		version: v0.1.2
+	//		resourceName: database.example.org
```
Avni-Sharma: I guess there should be one resourceKind here. Apart from that LGTM :)

PR #21, opened 2019-06-21
Remove unnecessary vendor directory and ignore it.

pmacik:
The `vendor` directory is generated and filled by the `make vendor` target so we don't really need to have it in the repository.

This PR:
* Removes the `vendor` directory, it made to be ignored by git.
* Update the `travis.yaml` to use make targets instead of direct commands to run tests.


baijum:
I can see most of the projects under openshift has the vendor directory:
https://github.com/openshift

pmacik:
> I can see most of the projects under openshift has the vendor directory:
> https://github.com/openshift

So, there is a pattern... OK, I'm dropping this PR

PR #32, opened 2019-06-25
Add Dockerfile 

sbose78:
- Adding this Dockerfile until we have automated mirroring of builds happening from openshift CI
- I plan to integrate this with quay.io/redhat-developer
- Added `make build`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,83 @@
+FROM centos:7 as build-tools
+LABEL maintainer "Devtools <devtools@redhat.com>"
+LABEL author "Shoubhik Bose <shbose@redhat.com>"
+ENV LANG=en_US.utf8
+ENV GOPATH /tmp/go
+ARG GO_PACKAGE_PATH=github.com/redhat-developer/service-binding-operator
+
+RUN yum install epel-release -y \
+    && yum install --enablerepo=centosplus install -y --quiet \
+    findutils \
+    git \
+    golang \
+    make \
+    procps-ng \
+    tar \
+    wget \
+    which \
+    bc \
+    kubectl \
+    yamllint \
+    python36-virtualenv \
+    httpd-tools \
+    && yum clean all
+
+# install dep
+RUN mkdir -p $GOPATH/bin && chmod a+rwx $GOPATH \
+    && curl -L -s https://github.com/golang/dep/releases/download/v0.5.1/dep-linux-amd64 -o dep \
+    && echo "7479cca72da0596bb3c23094d363ea32b7336daa5473fa785a2099be28ecd0e3  dep" > dep-linux-amd64.sha256 \
+    && sha256sum -c dep-linux-amd64.sha256 \
+    && rm dep-linux-amd64.sha256 \
+    && chmod +x ./dep \
+    && mv dep $GOPATH/bin/dep
+
+ENV PATH=$PATH:$GOPATH/bin
+
+# download, verify and install openshift client tools (oc and kubectl)
+WORKDIR /tmp
+RUN curl -L -s https://github.com/openshift/origin/releases/download/v3.11.0/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz -o openshift-origin-client-tools.tar.gz \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,83 @@
+FROM centos:7 as build-tools
```
sbose78: Haven't figured out how to use UBI here.

pkg/controller/servicebindingrequest/servicebindingrequest_controller.go
```diff
@@ -157,7 +157,7 @@ outerLoop:
 					evs := &corev1.EnvVarSource{
 						SecretKeyRef: sks,
 					}
-					evn := strings.ToUpper(strings.ReplaceAll(instance.Name, "-", "_")) + "_" + strings.ToUpper(strings.ReplaceAll(key, "-", "_"))
+					evn := strings.ToUpper(strings.Replace(instance.Name, "-", "_", -1)) + "_" + strings.ToUpper(strings.Replace(key, "-", "_", -1))
```
baijum: This is an unrelated change. Please use Go 1.12

pkg/controller/servicebindingrequest/servicebindingrequest_controller.go
```diff
@@ -173,7 +173,7 @@ outerLoop:
 					evs := &corev1.EnvVarSource{
 						ConfigMapKeyRef: cmks,
 					}
-					evn := strings.ToUpper(strings.ReplaceAll(instance.Name, "-", "_")) + "_" + strings.ToUpper(strings.ReplaceAll(key, "-", "_"))
+					evn := strings.ToUpper(strings.Replace(instance.Name, "-", "_", -1)) + "_" + strings.ToUpper(strings.Replace(key, "-", "_", -1))
```
baijum: This is an unrelated change. Please use Go 1.12

Makefile
```diff
@@ -141,3 +141,9 @@ test-unit:
 ./vendor: go.mod go.sum
 	$(Q)GOCACHE=$(shell pwd)/out/gocache GO111MODULE=on go mod vendor ${V_FLAG}
 
+
+.PHONY: build
+build: ./out/operator
+
+./out/operator:
+	$(Q)CGO_ENABLED=0 GOARCH=amd64 GOOS=linux go build  ${V_FLAG} -o ./out/operator  cmd/manager/main.go
```
baijum: `operator-sdk build` produce docker image with UBI as the base image.
Earlier we used `go build` as there was no way pass build flags. But that is fixed in SDK: https://github.com/operator-framework/operator-sdk/issues/1435

pkg/controller/servicebindingrequest/servicebindingrequest_controller.go
sbose78: Ah ok! That explains it. Thanks.

sbose78:
Ok, where could I find the operator image built off master?

baijum:
> Ok, where could I find the operator image built off master?

As soon as you run `operator-sdk build <image>` command, it produces the image. The image will be in the same system where this command was run.

But if the intent of this Makefile target is to use inside OpenShift CI like how we did it in DevConsole operator, then this will be required.
https://github.com/openshift/release/blob/872b1e09f6e5c3cca821b8c8bac90b54f9355573/ci-operator/config/redhat-developer/devconsole-operator/redhat-developer-devconsole-operator-master.yaml#L23

baijum:
@sbose78 You can use the `openshift-ci/Dockerfile.tools` for the Dockerfile here. It has code Go 1.12 and other required toold.

sbose78:
Eventually the intent is to build it on ci and mirror images to Quay. 

But till we have that, I wish to setup master builds on Quay so that it could be used in the OperatorSource without having to locally build it and push it.

sbose78:
( I will address the go1.12 comments )

otaviof:
If we really need a Dockerfile to be able to run this under OpenShift CI, we should find a automated way to extract the data from a `operator-sdk build` command output. By using automation we can mitigate the risk of deviating from what `operator-sdk` provides for the images we will be consuming in upstream.

baijum:
@otaviof I have sent a PR to introduce building without OCI image: https://github.com/operator-framework/operator-sdk/pull/1600

PR #35, opened 2019-06-26
Add Makefile target for CI

DhritiShikhar:
Adds makefile target for Ci


manifests/service-binding-operator/service-binding-operator.package.yaml
```diff
@@ -0,0 +1,4 @@
+packageName: service-binding-operator
+channels:
+- name: stable
+  currentCSV: service-binding-operator.v0.0.1
```
pratikjagrut: By adding new line at the end this symbol can be removed.

DhritiShikhar: but I dont want a new line

pratikjagrut: @baiju told me the same in my previous PR of Makefile.

baijum: @DhritiShikhar So, you want to remove newline character (`\n`) at the end of the file from all source files?

As per POSIX standard, a line is defined like this:
`A sequence of zero or more non- <newline> characters plus a terminating <newline> character.`
Refer to this link: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206

So, without a newline character, it's not a line.

I am fine with not following POSIX standard as long as all the source files are in the same way. That means removing newline at the end of file from all source files. We need consistency. Otherwise, this will make unnecessary automatic changes depending on the editor configuration.

Please read these two links also:
- https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file
- https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

baijum:
> The manifests are already present in `deploy/olm-catalog..`
> Could you ensure we have only one copy?

In the `deploy/olm-catalog` directory the operator image location is hardcoded. This will not work for CI.

DhritiShikhar:
@baijum @sbose78 So, let's have just one copy - under `manifests/` ?

baijum:
@DhritiShikhar I am merging this now. We will fix CI for e2e and lint. You can send another PR to fix remaining issues.

DhritiShikhar:
@baijum okay

PR #36, opened 2019-06-26
Small makefile fixes

pratikjagrut:
Renaming `test-e2e` to `test-unit`, as incorrect name was given to unit test target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,7 @@ test-unit:
 # Build and vendor tarets
 #---------------------------------------------------------
 
+.PHONEY: ./vendor
 # Vendor target: "go mod vendor" resets the main module's vendor directory to include all packages needed to build and 
 # test all of the module's packages based on the state of the go.mod files and Go source code.
 ./vendor: go.mod go.sum
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ test-unit:
 # Build and vendor tarets
 #---------------------------------------------------------
 
-.PHONEY: ./vendor
+.PHONY: ./vendor
 # Vendor target: "go mod vendor" resets the main module's vendor directory to include all packages needed to build and 
 # test all of the module's packages based on the state of the go.mod files and Go source code.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -113,7 +113,7 @@ get-test-namespace: ./out/test-namespace
 
 # E2E test
 .PHONY: e2e-setup
-e2e-setup: e2e-cleanup
+e2e-setup: ./vendor e2e-cleanup
```
baijum: Vendor directory is already committed to git. So this is not required.

Makefile
```diff
@@ -80,17 +80,17 @@ GO_PACKAGE_PATH ?= github.com/${GO_PACKAGE_ORG_NAME}/${GO_PACKAGE_REPO_NAME}
 GOLANGCI_LINT_BIN=./out/golangci-lint
 .PHONY: lint
 ## Runs linters on Go code files and YAML files
-lint: lint-go-code lint-yaml
+lint: ./vendor lint-go-code lint-yaml
```
baijum: Vendor directory is already committed to git. So this is not required.

Makefile
pratikjagrut: Ok, I'll remove it.

Makefile
pratikjagrut: Ok, I'll remove it.

Makefile
```diff
@@ -136,6 +136,7 @@ test-unit:
 # Build and vendor tarets
 #---------------------------------------------------------
 
+.PHONY: ./vendor
```
baijum: The `./vendor` need not be a PHONY target. It's an output directory. See the introduction: https://www.gnu.org/software/make/manual/html_node/Rule-Introduction.html#Rule-Introduction

sbose78:
Add a description, please.

Makefile
pratikjagrut: Thanks @baijum I'll remove it.

PR #37, opened 2019-06-27
[ODC-1165] Refactoring Reconcile and E2E Bootstrap

otaviof:
# Description

Refactoring `Reconciler` logic to be split up into components. The idea is to have a `Planner` to inspect the namespace and locate the backend CRD instance, and later a `Retriever` to load all the interesting data into a single secret, to be mounted as `envFrom` in the applications matching label selector.

## Testing

To run from source-code, run:

``` bash
operator-sdk up local --verbose --namespace openshift-operators
```

Assuming this operator and [postgresql operator](https://github.com/baijum/postgresql-operator) are already running, execute the following:

### Create a Database

``` yaml
---
apiVersion: postgresql.baiju.dev/v1alpha1
kind: Database
metadata:
  name: demo-database
spec:
  dbName: postgres
  image: docker.io/postgres:latest
  imageName: postgres
```

### Create a Deployment

``` yaml
---
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: busybox
  annotations:
    connects-to: demo-database
    environment: e2e
spec:
  replicas: 1
  template:
    metadata:
      labels:
        connects-to: demo-database
        environment: e2e
    spec:
      containers:
        - name: busybox
          image: busybox:latest
          command: [ "sleep", "3600" ]
```

### Create a `ServiceBindingRequest`

``` yaml
---
kind: ServiceBindingRequest
metadata:
  name: binding-request
spec:
  applicationSelector:
    matchLabels:
      connects-to: demo-database
      environment: e2e
    resourceKind: Deployment
  backingSelector:
    objectName: demo-database
    resourceName: postgresql.baiju.dev
    resourceVersion: v1alpha1
```

### Assert

The target deployment should have `envFrom` directive:

```
$ k get deploymentconfigs demo-nodejs-crud -o yaml |grep -A2 -i envfrom
      - envFrom:
        - secretRef:
            name: binding-request
```

And assert the items in a secret with the same name than `ServiceBindingRequest` has all interesting items:

```
$ k get secret binding-request -o json |jq --raw-output '.data'
{
  "SERVICE_BINDING_DATABASE_DBCONNECTIONIP": "",
  "SERVICE_BINDING_DATABASE_DBCONNECTIONPORT": "MA==",
  "SERVICE_BINDING_DATABASE_DBCREDENTIALS": "ZGVtby1kYXRhYmFzZS1wb3N0Z3Jlc3Fs",
  "SERVICE_BINDING_DATABASE_DBNAME": "cG9zdGdyZXM=",
  "SERVICE_BINDING_DATABASE_SECRET_PASSWORD": "cGFzc3dvcmQ=",
  "SERVICE_BINDING_DATABASE_SECRET_USER": "cG9zdGdyZXM="
}
```


test/e2e/servicebindingrequest_test.go
```diff
@@ -0,0 +1,191 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/redhat-developer/service-binding-operator/pkg/apis"
+	v1alpha1 "github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1"
+)
+
+var (
+	retryInterval        = time.Second * 5
+	timeout              = time.Second * 120
+	cleanupRetryInterval = time.Second * 1
```
golangcibot: `cleanupRetryInterval` is unused (from `deadcode`)

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -0,0 +1,226 @@
...
+			return &csv
+		}
+	}
+
+	return nil
+}
+
+// intermediarySecret create a secret to be used as a intermediary place beteween operator descriptor
+// fields and applications interested to have them.
+func (r *ReconcileServiceBindingRequest) intermediarySecret(csv *olmv1alpha1.ClusterServiceVersion) *corev1.Secret {
```
golangcibot: U1000: func `(*ReconcileServiceBindingRequest).intermediarySecret` is unused (from `unused`)

baijum:
@otaviof The `Makefile` and `.gitignore` changes could be a separate PR. So that we can merge it right away.

test/e2e/servicebindingrequest_test.go
```diff
@@ -0,0 +1,210 @@
...
+	cleanupTimeout       = time.Second * 5
+	replicas             = 1
+	operatorKind         = "ServiceBindingRequest"
+	operatorAPIVersion   = "apps.openshift.io/v1alpha1"
+	olmKind              = "CustomResourceDefinition"
```
golangcibot: `olmKind` is unused (from `deadcode`)

test/e2e/servicebindingrequest_test.go
```diff
@@ -0,0 +1,210 @@
...
+	olmAPIVersion        = "apiextensions.k8s.io/v1beta1"
```
golangcibot: `olmAPIVersion` is unused (from `deadcode`)

pkg/controller/servicebindingrequest/controller.go
```diff
@@ -0,0 +1,56 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	v1alpha1 "github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1"
+)
+
+var log = logf.Log.WithName("controller_servicebindingrequest")
```
golangcibot: `log` is unused (from `deadcode`)

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+				continue
+			}
+			return &csv
+		}
+	}
+
+	return nil
+}
+
+func (r *ReconcileServiceBindingRequest) retrieveData() {
```
golangcibot: U1000: func `(*ReconcileServiceBindingRequest).retrieveData` is unused (from `unused`)

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -0,0 +1,252 @@
...
+
+			logger.Info("CRD matches backing-selector!")
+			return &csv
+		}
+	}
+
+	return nil
+}
+
+func (r *ReconcileServiceBindingRequest) retriveTargetServiceCRD(crdName, crdVersion, name string) {
```
golangcibot: U1000: func `(*ReconcileServiceBindingRequest).retriveTargetServiceCRD` is unused (from `unused`)

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -0,0 +1,527 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	v1alpha1 "github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1"
+)
+
+const (
+	connectsToLabel                = "connects-to"
```
Avni-Sharma: It is currently hard coded right?

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -0,0 +1,527 @@
...
+
+// pathValue read value of a given "path" inside of informed custom resource definition instance.
+func (r *Reconciler) pathValue(crd *unstructured.Unstructured, path string) (string, error) {
+	object := crd.Object
+	status, exists := object["status"].(map[string]interface{})
+	if !exists {
+		return "", fmt.Errorf("unable to find 'status' field in object '%s'", crd.GetName())
+	}
+
+	value, exists := status[path]
```
Avni-Sharma: Sometimes the value can be as a nested struct. For example - https://github.com/operator-framework/community-operators/blob/f980b60e6862cccb3f7eaf8bc2d9baff23a847e0/community-operators/etcd/etcdoperator.v0.9.2.clusterserviceversion.yaml#L298
Corresponding source code - https://github.com/coreos/etcd-operator/blob/8347d27afa18b6c76d4a8bb85ad56a2e60927018/pkg/apis/etcd/v1beta2/restore_types.go#L65
(For this sprint maybe we can omit this, but this should be under consideration)

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -0,0 +1,385 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	v1alpha1 "github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1"
+)
+
+const (
+	connectsToLabel                = "connects-to"
+	almDescriptorPrefix            = "urn:alm:descriptor"
+	almKubernetesPrefix            = "io.kubernetes"
```
golangcibot: `almKubernetesPrefix` is unused (from `varcheck`)

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -0,0 +1,385 @@
...
+	sections := strings.Split(xDescriptor, ":")
+	kind := sections[len(sections)-2]
+	attribute := sections[len(sections)-1]
+
+	return kind, attribute
+}
+
+// generateFieldName based on the x-descriptor value, generating a string with the same standards
+// than environment variable naming.
+func (r *Reconciler) generateFieldName(xDescriptor string) string {
```
Avni-Sharma: Do we really need the complete Xdescriptor or can we take the last segment after the ":". For eg- <CR NAME>_<keyvalue>

Avni-Sharma: For eg if CR is `postgres` and `key` is password then the env variable will become `POSTGRES_PASSWORD`

Avni-Sharma:
### Descriptor specification for retrieving and consuming the configuration from CSV.

The first case is to retrieve the configuration specified as a spec/status attribute.
`urn:alm:descriptor:servicebindingrequest:env:attribute: ...`

The second case is to retrieve a specific key from a secret object.
`urn:alm:descriptor:servicebindingrequest:env:object:secret:<key>: ...`

The third case is to retrieve a specific key from a configmap object.
`urn:alm:descriptor:servicebindingrequest:env:object:configmap:<key>: ...`

Note 1 - An `env` section is added in the descriptor because there can be a future scope wherein we need to consume configuration as a volume mount. So when this situation arises we can replace `env` with `volumemount`.

pkg/controller/servicebindingrequest/reconciler.go
otaviof: @Avni-Sharma The intention is to not have hardcoded label names, and use a constant to define it.

pkg/controller/servicebindingrequest/reconciler.go
otaviof: @Avni-Sharma indeed, that's a limitation we need a better approach to overcome than it's currently implemented. I think we could bring a library to aid in the dot-notation to extract data from a generic `map[string]interface{}` recursively.

pkg/controller/servicebindingrequest/reconciler.go
otaviof: @Avni-Sharma that's a good question, I think @sbose78 and @baijum can share more information.

cmd/manager/main.go
```diff
@@ -7,6 +7,7 @@ import (
 	"os"
 	"runtime"
 
+	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.
```
baijum: This line is not required. Next line content is the same.

go.mod
```diff
@@ -1,47 +1,26 @@
 module github.com/redhat-developer/service-binding-operator
 
 require (
-	contrib.go.opencensus.io/exporter/ocagent v0.4.9 // indirect
-	github.com/Azure/go-autorest v11.5.2+incompatible // indirect
-	github.com/appscode/jsonpatch v0.0.0-20190108182946-7c0e3b262f30 // indirect
+	github.com/baijum/postgresql-operator v0.0.0-20190627061224-9ebf4814fa66
```
baijum: This is not a good dependency!

pkg/controller/servicebindingrequest/reconciler.go
baijum: @otaviof What is the purpose of this label? To identify the backing CRD, it's name and version is sufficient, right?

baijum: @otaviof On a closer look, I realized that we need label selectors for backing CR also. Something like this:
`MatchLabels  map[string]string `json:"matchLabels"``
I will send a separate PR for the same, right away.

baijum: I send PR: https://github.com/redhat-developer/service-binding-operator/pull/43
Please review and approve!

go.mod
otaviof: Good point, let me fix it.

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -0,0 +1,180 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	v1alpha1 "github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1"
+)
+
+const (
+	almDescriptorPrefix            = "urn:alm:descriptor"
```
golangcibot: `almDescriptorPrefix` is unused (from `deadcode`)

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -0,0 +1,180 @@
...
+	almKubernetesPrefix            = "io.kubernetes"
+	almServiceBindingRequestPrefix = "io.servicebindingrequest"
```
golangcibot: `almServiceBindingRequestPrefix` is unused (from `deadcode`)

cmd/manager/main.go
otaviof: I think it would not hurt if we leave, since that's a part of every Kubernetes client. So if we have developers working against those clouds, our project is already compatible, maybe we have more reasons to let it stay than to remove it.

baijum: I was not referring to the import statement. The comment above that is (line 10 and 11) are the same content.

baijum:
@otaviof Are you still working on attribute access.

There are a few attributes here: 
https://github.com/baijum/postgresql-operator/blob/f72b43994f54ea1c543978b4d2e449b93db9cf64/manifests/0.0.3/postgresql-operator.v0.0.3.clusterserviceversion.yaml#L166

sbose78:
https://github.com/redhat-developer/service-binding-operator/pull/37/files#diff-2a5a5377bb5993112bb22eedda855700 . 

The file deploy/olm-catalog should be gone

sbose78:
manifests/0.0.1/service-binding-operator.v0.0.1.clusterserviceversion.yaml isn't in master. 
`manifests/` is flat.

otaviof:
@sbose78  just removed the file.

otaviof:
> @otaviof The `Makefile` and `.gitignore` changes could be a separate PR. So that we can merge it right away.

I've made changes to the file based in this comment, please consider.

pkg/controller/servicebindingrequest/planner.go
```diff
@@ -0,0 +1,166 @@
...
+	CRDDescription *olmv1alpha1.CRDDescription // custom resource definition description
+	CRD            *ustrv1.Unstructured        // custom resource definition
+}
+
+const (
+	connectsToLabel = "connects-to"
+)
+
+// extractConnectTo inspect ServiceBindingRequest to extract connects-to value.
+func (p *Planner) extractConnectsTo() string {
```
sbose78: @otaviof 

```
matchLabels:
      connects-to: postgres
      environment: production
```

The hard requirement for having a specific label should not be done. That would make it a strong constraint. Any label should be supported here for selecting objects. We can take care of this in a separate PR though.

Makefile
```diff
@@ -136,8 +146,45 @@ test-unit:
 # Build and vendor tarets
 #---------------------------------------------------------
 
-# Vendor target: "go mod vendor" resets the main module's vendor directory to include all packages needed to build and 
-# test all of the module's packages based on the state of the go.mod files and Go source code.
+.PHONY: build
+## Build: using operator-sdk to build a new image
+build:
```
sbose78: @otaviof , Could you please fix the duplication here?
there's already a `build` target which does a typical `go build`. 

Sync-ing this with master shows up the conflict.

We need to have both targets
- `make build` : Compiles and does a simple `go build` as we have in `master`
-  and another target which does a lot more than that with operator-sdk

Makefile
```diff
@@ -144,11 +154,48 @@ test-e2e-olm-ci:
...
+	operator-courier --verbose flatten $(MANIFESTS_DIR) $(MANIFESTS_TMP)
+	cp -vf deploy/crds/*_crd.yaml $(MANIFESTS_TMP)
+
+#---------------------------------------------------------
+# Deploy
+#---------------------------------------------------------
+
+## Prepare-CSV: using a temporary location copy all operator CRDs and metadata to generate a CSV.
+prepare-csv: build-image
+	$(eval ICON_BASE64_DATA := $(shell cat ./assets/icon/red-hat-logo.png | base64))
```
baijum: Probably it's not a good idea to use Red Hat old logo here.
For now, we can use some placeholder image.
Check this site to other product logos usage: https://operatorhub.io/
BTW, here is the new Red Hat logo details:
https://www.redhat.com/en/about/brand/standards/logo

Makefile
otaviof: Sure! It's now having both targets, `build` and `build-image`, please consider.

pkg/controller/servicebindingrequest/planner.go
otaviof: Indeed, I think we can address this in a stand-along PR, which could also include E2E coverage started on this PR.

Makefile
otaviof: Good point. I just wanted a temporary logo, and now I'm picking up one from the link you shared. Please consider.

cmd/manager/main.go
otaviof: Check! I missed that one, but it's gone now.

sbose78:
I can confirm that `make build` works and `docker build..` works as well.
This is good enough to use the operator image.

PR #38, opened 2019-06-27
Improving Makefile Help and Git Ignore

otaviof:
As a spin-off PR, improving the help message of build target and the message display itself. Also ignoring `.env` file in local project, and the whole `build` directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # Temporary Build Files
-build/_output
-build/_test
+build/
+
+# dot-env files, local project environment variables
+.env
+
 
 #gocache
 out/
--- a/Makefile
+++ b/Makefile
@@ -39,15 +39,13 @@ $(shell mkdir -p ./out);
 # HELP target
 #----------------------------------------------------------------
 
-## This target is used to display all other target
-.PHONY: help
 # Based on https://gist.github.com/rcmachado/af3db315e31383502660
 ## Display this help text
 help:/
 	$(info Available targets)
 	$(info -----------------)
 	@awk '/^[a-zA-Z\-%\_0-9]+:/ { \
-		helpMessage = match(lastLine, /^## (.*)/); \
+		helpMessage = match(lastLine, /^## (.*)/);
 		helpCommand = substr($$1, 0, index($$1, ":")-1); \
 		if (helpMessage) { \
 			helpMessage = substr(lastLine, RSTART + 3, RLENGTH); \
@@ -57,13 +55,12 @@ help:/
 		} \
 	} \
 	{ hasComment = match(lastLine, /^## (.*)/); \
-          if(hasComment) { \
+		if(hasComment) { \
             lastLine=lastLine$$0; \
-	  } \
-          else { \
-	    lastLine = $$0 \
-          } \
-        }' $(MAKEFILE_LIST)
+		} else { \
+			lastLine = $$0 \
+		} \
+	}' $(MAKEFILE_LIST)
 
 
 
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1,6 +1,9 @@
 # Temporary Build Files
-build/_output
-build/_test
+build/
```
baijum: There are few files under the `build` which need to be tracked, right?

Makefile
```diff
@@ -126,7 +123,7 @@ test-e2e: e2e-setup
 	$(info Running E2E test: $@)
 	$(Q)GO111MODULE=on operator-sdk test local ./test/e2e --namespace $(TEST_NAMESPACE) --up-local --go-test-flags "-v -timeout=15m"
 
-.PHONY: test-unit
+.PHONY: test-e2e
```
baijum: It's `unit`. Please revert.

pratikjagrut: @otaviof This target is to run unit test. You can see `test-e2e` target just above it.

.gitignore
pratikjagrut: We have Dockerfile inside build. I think this file need to be tracked.

Makefile
```diff
@@ -39,15 +39,13 @@ $(shell mkdir -p ./out);
...
-## This target is used to display all other target
-.PHONY: help
 # Based on https://gist.github.com/rcmachado/af3db315e31383502660
 ## Display this help text
 help:/
 	$(info Available targets)
 	$(info -----------------)
 	@awk '/^[a-zA-Z\-%\_0-9]+:/ { \
-		helpMessage = match(lastLine, /^## (.*)/); \
+		helpMessage = match(lastLine, /^## (.*)/);
```
pratikjagrut: We need `\` here otherwise it will give `unexpected EOF while looking for matching `''` error.

I'm trying to run `make help` throwing error:
```
Available targets
-----------------
/bin/bash: -c: line 0: unexpected EOF while looking for matching `''
/bin/bash: -c: line 2: syntax error: unexpected end of file
make: *** [Makefile:47: help] Error 1
```

Makefile
ldimaggi: Yes - they are 2 separate tests.

.gitignore
ldimaggi: We need that Dockerfile - yes!

Makefile
```diff
@@ -136,14 +133,13 @@ test-unit:
...
+build:
+	$(Q)GO111MODULE=on operator-sdk build "$(GO_PACKAGE_ORG_NAME)/$(GO_PACKAGE_REPO_NAME):latest"
+
+# Vendor target: "go mod vendor" resets the main module's vendor directory to include all packages needed to build and
 # test all of the module's packages based on the state of the go.mod files and Go source code.
 ./vendor: go.mod go.sum
 	$(Q)GOCACHE=$(shell pwd)/out/gocache GO111MODULE=on go mod vendor ${V_FLAG}
 
-
-.PHONY: build
```
pratikjagrut: We need this `build` target as it generates binaries of this operator and we use it in opeshift-ci. Am I right @baijum ? 
As `operator-sdk build` will give you an image so you can rename that target as `build-operator-image` or something similar and keep this target.

baijum: Yes, this target is used for CI.

sbose78: I find this way of building a Golang project a bit more transparent instead of depending on a library :)

pratikjagrut: As operator-sdk does not provide any significant way to produce binary of operator we need to use this way.

pratikjagrut:
I think this PR is not of any use now as many of the changes has already took place and made this PR outdated. @otaviof can you please close this if there is nothing to merge in master from this PR.

baijum:
To close a PR type:
/close

PR #50, opened 2019-07-03
Change channel from stable to alpha

Avni-Sharma:
This PR provides two channels - `dev` and `stable`. In the future, we can have these channels pointing to different ClusterServiceVersion respectively. For development purpose, we can use the dev channel and a stable release can be under the stable channel.  By default the cannel is the `dev` channel.

### DIFF
--- a/manifests/service-binding-operator.package.yaml
+++ b/manifests/service-binding-operator.package.yaml
@@ -1,4 +1,7 @@
 packageName: service-binding-operator
 channels:
-- name: alpha
+- name: dev
   currentCSV: service-binding-operator.v0.0.10
+- name: stable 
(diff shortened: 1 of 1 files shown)


Avni-Sharma:
> Is this is the only place channel is specified?

yes.
Also https://github.com/redhat-developer/service-binding-operator/blob/master/test/e2e/catalog_source.yaml and subscription.yaml is not needed anymore, if we are using operator source. wdyt

sbose78:
I would prefer to have two channels

Avni-Sharma:
cc @pmacik

pmacik:
LGTM

PR #51, opened 2019-07-04
Create a more readable help target

baijum:
- The help target shows sections of the target. Colorization for targets
  in the output helps to recognize targets.
- Remove the `./` prefix from targets where a path needs to be specified. This causes some target
  not properly rendered in the help output.
- Remove `vendor` as a dependency and it is already part of the source tree.

Current help output:

Screenshot from 2019-07-04 11-24-24

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+.DEFAULT_GOAL := help
+
 # It's necessary to set this because some environments don't link sh -> bash.
 SHELL := /bin/bash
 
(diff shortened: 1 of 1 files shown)


pratikjagrut:
`./` prefix targets are anyway used in as dependencies to other target so help massage for them won't be of  much use as those are not suppose run individually. And `./` prefix was according to the directory structure/path on which they will be working or storing there output.

baijum:
> `./` prefix targets are anyway used in as dependencies to other target so help massage for them won't be of much use as those are not suppose run individually. And `./` prefix was according to the directory structure/path on which they will be working or storing there output.

The `./vendor` was used as a target that needs to run sometimes. I changed the other one for consistency.

baijum:
I will not attempt to fix the unit test & e2e test in this PR. We can work on it in a separate PR.

PR #52, opened 2019-07-04
Add operator-courier lint, run operator locally and cleaning targets

pratikjagrut:
This PR adds `courier` target a sub target in lint, `local` target to run operator locally and `clean` target to clean cached directory. `local` target has some dependency targets added in this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,6 +220,8 @@ push-image: build-image
 	docker push "$(OPERATOR_IMAGE):$(OPERATOR_TAG_LONG)"
 	docker push "$(OPERATOR_IMAGE):latest"
 
+## -- Local deployment targets --
+
 .PHONY: local
 ## Run operator locally
 local: deploy-clean deploy-rbac deploy-crds deploy-cr
@@ -253,9 +255,7 @@ deploy-clean:
 	$(Q)-kubectl delete -f deploy/service_account.yaml
 
 
-#---------------------------------------------------------
-# Cleaning targets
-#---------------------------------------------------------
+## -- Cleanup targets --
 
 .PHONY: clean
 ## Removes temp directories
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,7 @@ courier:
 	$(Q)python3 -m venv ./out/venv3
 	$(Q)./out/venv3/bin/pip install --upgrade setuptools
 	$(Q)./out/venv3/bin/pip install --upgrade pip
-	$(Q)./out/venv3/bin/pip install operator-courier==2.0.2
+	$(Q)./out/venv3/bin/pip install operator-courier
 	$(Q)./out/venv3/bin/operator-courier flatten ./manifests ./out/manifests
 	$(Q)./out/venv3/bin/operator-courier verify ./out/manifests
 
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ MANIFESTS_TMP ?= ./tmp/manifests
 GOLANGCI_LINT_BIN=./out/golangci-lint
 .PHONY: lint
 ## Runs linters on Go code files and YAML files
-lint: lint-go-code lint-yaml
+lint: lint-go-code lint-yaml courier
 
 YAML_FILES := $(shell find . -path ./vendor -prune -o -type f -regex ".*y[a]ml" -print)
 .PHONY: lint-yaml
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -210,3 +219,45 @@ push-image: build-image
 	docker tag "$(OPERATOR_IMAGE):$(OPERATOR_TAG_LONG)" "$(OPERATOR_IMAGE):latest"
 	docker push "$(OPERATOR_IMAGE):$(OPERATOR_TAG_LONG)"
 	docker push "$(OPERATOR_IMAGE):latest"
+
```
baijum: Add a line like this here:
```
## -- Local deployment and cleanup targets --
```

This will become the section title.

Makefile
```diff
@@ -210,3 +219,45 @@ push-image: build-image
...
+
+#---------------------------------------------------------
+# Cleaning targets
+#---------------------------------------------------------
+
+.PHONY: clean
+## Removes temp directories
+clean:
+	$(Q)-rm -rf ${V_FLAG} ./out
```
baijum: Do we really need to silence the command itself here?
The hyphen here (`$(Q)-`) makes the command invisible, right?

Makefile
```diff
@@ -121,6 +121,15 @@ lint-go-code: $(GOLANGCI_LINT_BIN)
 $(GOLANGCI_LINT_BIN):
 	$(Q)curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ./out v1.17.1
 
+.PHONY: courier
+## Validate manifests using operator-courier
+courier:
+	$(Q)python3 -m venv ./out/venv3
+	$(Q)./out/venv3/bin/pip install --upgrade setuptools
+	$(Q)./out/venv3/bin/pip install --upgrade pip
+	$(Q)./out/venv3/bin/pip install operator-courier==2.0.2
```
baijum: There is a new version of operator-courier: https://pypi.org/project/operator-courier/2.1.2/

Makefile
```diff
@@ -121,6 +121,15 @@ lint-go-code: $(GOLANGCI_LINT_BIN)
...
+	$(Q)./out/venv3/bin/operator-courier flatten ./manifests ./out/manifests
```
baijum: The `manifests` directory is already in the flat structure, right?
Can we directly run the `verify` command against this directory?

Makefile
```diff
@@ -121,6 +121,15 @@ lint-go-code: $(GOLANGCI_LINT_BIN)
 $(GOLANGCI_LINT_BIN):
 	$(Q)curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ./out v1.17.1
 
+.PHONY: courier
+## Validate manifests using operator-courier
+courier:
+	$(Q)python3 -m venv ./out/venv3
```
baijum: I like the approach of creating a virtual environment for installing operator-courier. Can we do a similar thing for `yamllint` inside `lint-yaml` target. Then we can remove `yamllint` from the container image.
Refere: https://github.com/redhat-developer/service-binding-operator/blob/e21d5b1f2e9b874c9172296699ed4cb54a3d572c/openshift-ci/Dockerfile.tools#L20

baijum: This change need not be in this PR.

Makefile
pratikjagrut: Its not necessary it was just a precaution, if due to some reason it gives any error then it should halt any execution(if clean target is added as dependency to some other target in future).

Makefile
pratikjagrut: Oh ok, as consistency to other targets you merged in your help target PR. Will add.

Makefile
pratikjagrut: Yes we can, but our `manifests` directory structure has changed so many times in last few days. And if somebody again try to make it according to operator-registry's manifests format this code will break. I already experienced this few time in last week. So its better to keep flatten command here. And anyway its creating flatten dir in out dir. So won't matter much.

Makefile
pratikjagrut: Thank you, I'll create another PR for this.

Makefile
pratikjagrut: They are quite fast, on June 22 they released `2.1.1` and on June 24 `2.1.2`. Anyways I'll update it to latest one.

pratikjagrut: I'll just put `pip install operator-courier` it should fetch latest one.

Makefile
baijum: Once this is addressed, we can merge this PR.

PR #53, opened 2019-07-04
Adding Unit-Testing on Retriever (Controller)

otaviof:
Covering with unit-tests the `Retriever` type inside controller. Also, modifying make target `test-unit` to allow `TEST_EXTRA_ARGS`. For example:

``` bash
make test-unit TEST_EXTRA_ARGS='-run "^TestRetriever.*"
```

Additionally, renaming `subscription.yaml` and `catalog_source.yaml` out of `test/e2e` directory, since in there we expect to only find Go files.


baijum:
@otaviof How about the renaming of YAML files into another PR?

pkg/controller/servicebindingrequest/retriever_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func TestRetrieverNew(t *testing.T) {
+	ns := "testing"
+	crdName := "db-testing"
+
+	crdDescription := mocks.CRDDescriptionMock()
+	crd := mocks.DatabaseCRDMock(ns, crdName)
+
+	genericCRDObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&crd)
+	assert.Nil(t, err)
```
baijum: If this is an absolute requirement to proceed further, you can use `require` package: `github.com/stretchr/testify/require`
Same in other places for the initial assertions.

otaviof: Good one! I didn't know about the package, but that's exactly what we need over there. I'm gonna change it.

otaviof:
> @otaviof How about the renaming of YAML files into another PR?

As we spoke, I'm improving the description of the PR. Please consider.

PR #54, opened 2019-07-05
First draft at expanded README file - to describe goals and usage of operator

ldimaggi:
At some point we may want to expand this further into a formal operator user guide document. As a first step, we should describe the goals of the operator and how it works for prospective users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,123 +1,150 @@
-# Service Binding Operator: Connect Applications with Services
+# Connect Applications with Operator-backed Services
 
-## Operator Goal - Binding Imported Applications with Operator-Backed Services
+## Introduction
 
 The goal of the Service Binding Operator is to enable application authors to 
-import an application and run it on OpenShift with backing services such as 
+import an application and run it on OpenShift with operator-backed services such as 
 databases, without having to perform manual configuration of secrets, 
 configmaps, etc. 
 
 In order for the Service Binding Operator to bind an application to a backing
 service, the backing service operator must specify the information required 
-to for the application to bind to the operator's service. The information must
+by the application to bind to the operator's service. The information must
 be specified in the operator's OLM descriptor from which it will be extracted
-to bind the application to the operator. The information can be specified in a
-secret, or in the "status" and/or "spec" section of the OLM.
+to bind the application to the operator. The information could be specified in
+the "status" and/or "spec" section of the OLM in plaintext or as a reference 
+to a secret.
 
-In order to make an imported application (for example, as a NodeJS 
-application) connect to a backing services (for example as a database):
+In order to make an imported application (for example, a NodeJS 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1 +1,121 @@
...
+    - description: Database connection port
+      displayName: DB port
+      path: dbConnectionPort
+      x-descriptors:
+    - urn:alm:descriptor:servicebindingrequest:env:attribute
+```
+
+## Quick Start
+
+To build:
```
otaviof: Should we point the audience to run Makefile targets instead? All the `kubectl apply -f` commands are covered after PR #52. :-)

README.md
```diff
@@ -1 +1,121 @@
...
+* The app author (developer) creates a ServiceBindingRequest and specifies a label in the form of "connects-to." For example, to bind to an operator-backed database service created using a CR named "my-db", the label will be "connects-to=my-db".
+
+* The Service Binding Operator then:
+  * Reads backing service operator OLM descriptor to discover the binding attributes
+  * Creates a binding secret for the database
+  * Injects environment variables into the applications's DeploymentConfig, Deployment, Replicaset
+
+Here is an example of the "bind-able" operator OLM Descriptor - in ths case for a PostgreSQL database backing operator:
+
+```
```
otaviof: After the triple backtick you can add a ` yaml` (space followed by the format), therefore GitHub will render colorized syntax. For instance:
```
yaml:
  entry: value
  int: 1
```
Versus:
``` yaml
yaml:
  entry: value
  int: 1
```

README.md
pratikjagrut: Yes right, we have make targets for basic purposes and we can add more as we get to know other purposes. So it’s good idea to point audience towards makefile target. Maybe in this process @ldimaggi could also identify new targets which can be part of makefile.

README.md
```diff
@@ -1 +1,121 @@
...
+To run the operator locally:
+
+	kubectl apply -f deploy/service_account.yaml
+	kubectl apply -f deploy/role.yaml
+	kubectl apply -f deploy/role_binding.yaml
+	kubectl apply -f deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml
+
+	And then you can: 
+	operator-sdk up local --verbose --namespace default
+
```
pratikjagrut: To run operator locally we have ‘make local’ target you can use it.

baijum:
Please wrap each line at 80 characters.
Ref: https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#general-file-guidelines

sbose78:
Since the `lint` is happy, I'll go ahead and merge this so that we have a README which describes what it does. Let's keep adding more to the docs.

PR #58, opened 2019-07-09
Replace docker by podman.

pmacik:
This PR:
* Replaces `docker` by `podman` (and `buildah`) tools. The reason is that OpenShift do not ship with Docker. The `podman` (and `buildah`) are used by OpenShift instead, as it is part of it, so we should, too.


Avni-Sharma:
Hi @pmacik. Can you rebase the PR.

pmacik:
> Hi @pmacik. Can you rebase the PR.

@Avni-Sharma Just rebased.

PR #62, opened 2019-07-12
Refactor test to use sub tests

baijum:
- Avoid global variable
- Make each sub test to run independently

### DIFF
--- a/pkg/controller/servicebindingrequest/retriever_test.go
+++ b/pkg/controller/servicebindingrequest/retriever_test.go
@@ -47,7 +47,7 @@ func TestRetriever(t *testing.T) {
 		assert.NotEmpty(t, retriever.data)
 	})
 
-	t.Run("GetCRKey", func(t *testing.T) {
+	t.Run("getCRKey", func(t *testing.T) {
 		imageName, err := retriever.getCRKey("spec", "imageName")
 		assert.Nil(t, err)
 		assert.Equal(t, "postgres", imageName)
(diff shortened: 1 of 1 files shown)


pkg/controller/servicebindingrequest/retriever_test.go
```diff
@@ -41,65 +40,66 @@ func TestRetrieverNew(t *testing.T) {
...
-}
+		assert.Contains(t, retriever.data, "SERVICE_BINDING_DATABASE_SECRET_USER")
+		assert.Contains(t, retriever.data, "SERVICE_BINDING_DATABASE_SECRET_PASSWORD")
+	})
 
-func TestRetrieverSaveDataOnSecret(t *testing.T) {
-	err := retriever.saveDataOnSecret()
-	assert.Nil(t, err)
-}
+	t.Run("Store", func(t *testing.T) {
```
otaviof: Should the title string follow the same capitalization than the method? So in this case it would become `store`?

pkg/controller/servicebindingrequest/retriever_test.go
```diff
@@ -41,65 +40,66 @@ func TestRetrieverNew(t *testing.T) {
...
 
-func TestRetrieverExtractSecretItemName(t *testing.T) {
-	assert.Equal(t, "user", retriever.extractSecretItemName(
-		"urn:alm:descriptor:servicebindingrequest:env:object:secret:user"))
-}
+	})
 
-func TestRetrieverReadSecret(t *testing.T) {
-	retriever.data = make(map[string][]byte)
+	t.Run("", func(t *testing.T) {
```
otaviof: Here we are missing a tittle string, which would be `extractSecretItemName`.

baijum: Done. a829d8538772192f7c58ac44ea3074ee0867e2c5

pkg/controller/servicebindingrequest/retriever_test.go
baijum: I made it lowe case. a829d8538772192f7c58ac44ea3074ee0867e2c5

ldimaggi:
Nice use of subtests!

PR #64, opened 2019-07-12
Support dotted notation in attribute path

baijum:
Fixes https://jira.coreos.com/browse/ODC-1243

### DIFF
--- a/pkg/controller/servicebindingrequest/retriever.go
+++ b/pkg/controller/servicebindingrequest/retriever.go
@@ -28,7 +28,8 @@ const (
 	secretPrefix  = "urn:alm:descriptor:servicebindingrequest:env:object:secret"
 )
 
-func getNestedValue(key string, sectionMap interface{}) (string, error) {
+// getNestedValue retrieve value from dotted key path
+func (r *Retriever) getNestedValue(key string, sectionMap interface{}) (string, error) {
 	if !strings.Contains(key, ".") {
 		value, exists := sectionMap.(map[string]interface{})[key]
 		if !exists {
@@ -41,7 +42,7 @@ func getNestedValue(key string, sectionMap interface{}) (string, error) {
 	if !exists {
 		return "", fmt.Errorf("Can't find '%v' section in CR", attrs)
 	}
-	return getNestedValue(attrs[1], newSectionMap.(map[string]interface{}))
+	return r.getNestedValue(attrs[1], newSectionMap.(map[string]interface{}))
 }
 
 // getCRKey retrieve key in section from CR object, part of the "plan" instance.
@@ -56,7 +57,7 @@ func (r *Retriever) getCRKey(section string, key string) (string, error) {
 		return "", fmt.Errorf("Can't find '%s' section in CR named '%s'", section, objName)
 	}
 
-	return getNestedValue(key, sectionMap)
+	return r.getNestedValue(key, sectionMap)
 }
 
 // read attributes from CR, where place means which top level key name contains the "path" actual
--- a/pkg/controller/servicebindingrequest/retriever.go
+++ b/pkg/controller/servicebindingrequest/retriever.go
@@ -29,10 +29,8 @@ const (
 )
 
 func getNestedValue(key string, sectionMap interface{}) (string, error) {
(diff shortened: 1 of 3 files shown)


pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -28,6 +28,24 @@ const (
 	secretPrefix  = "urn:alm:descriptor:servicebindingrequest:env:object:secret"
 )
 
+func getNestedValue(key string, sectionMap interface{}) (string, error) {
```
otaviof: I would expect to have `getNestedValue` as parte of the `Retriever` type. Is there a reason for standalone func?

pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -28,6 +28,24 @@ const (
...
+	var value interface{}
+	var exists bool
+	if !strings.Contains(key, ".") {
+		value, exists = sectionMap.(map[string]interface{})[key]
```
otaviof: You can just use inline declaration here (`:=`), since after this `if` you declare them again (line 42).

pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -28,6 +28,24 @@ const (
...
+		if !exists {
+			return "", fmt.Errorf("Can't find key '%s'", key)
+		}
+		return fmt.Sprintf("%v", value), nil
+	}
+	attrs := strings.SplitN(key, ".", 2)
```
otaviof: Do you need to check for `len()`, before doing the `SplitN`?

pkg/controller/servicebindingrequest/retriever.go
baijum: Retriever object is not accessed anywhere within the function. So, the standalone function would be fine.

pkg/controller/servicebindingrequest/retriever.go
baijum: That's not really required. This function will not return any error.

pkg/controller/servicebindingrequest/retriever.go
baijum: Done. 7351fde1f36e5fe43f8847745714b967842bd9dd

pkg/controller/servicebindingrequest/retriever_test.go
```diff
@@ -102,3 +102,44 @@ func TestRetriever(t *testing.T) {
...
+	}
+
+	dbSecret := mocks.SecretMock(ns, "db-credentials")
+	objs := []runtime.Object{&dbSecret}
+	fakeClient := fake.NewFakeClient(objs...)
+
+	retriever = NewRetriever(context.TODO(), fakeClient, plan)
+	require.NotNil(t, retriever)
+
+	t.Run("Second level", func(t *testing.T) {
```
ldimaggi: Is it worth adding an edge case here - where the path does not include a "." char?

Or - maybe a test of an invalid path (that should raise an error) such as "first..second"?

baijum: There is a test without any dot in line 50.

Added test for two dot scenario: b0f80a0c68a422e9d941f3b62f2fb9eb48c8f93a

pkg/controller/servicebindingrequest/retriever.go
otaviof: Indeed it does not access, but it's also not used anywhere else. Furthermore, is not ready to be used anywhere else, since it belongs to `Retriever` type. Since it does belong to Retriever and is only ready to be used in that context, I think it either deserves a standalone type, or join Retriever.
But leaving a global method with local scope is not something we should do in this case.

otaviof: By the way, we are also missing a doc-comment on this function, please.

baijum: Done. 991bb922eb12a0d1658ecc69521e64b51d12bbfc

otaviof: Thanks :-)

PR #83, opened 2019-07-22
Bind secret values as files into the container

baijum:
Based on x-descriptor value in the CSV file, the secret values will be mounted as files inside the container.

Ref: https://jira.coreos.com/browse/ODC-1324

### DIFF
--- a/pkg/controller/servicebindingrequest/reconciler.go
+++ b/pkg/controller/servicebindingrequest/reconciler.go
@@ -42,6 +42,49 @@ func (r *Reconciler) appendEnvFrom(envList []corev1.EnvFromSource, secret string
 	})
 }
 
+// appendVolumeMounts append volume mounts pointing to volumes created using secret
+func (r *Reconciler) appendVolumeMounts(vmList []corev1.VolumeMount, volumeName, mountPath string) []corev1.VolumeMount {
+	for _, vm := range vmList {
+		if vm.Name == volumeName {
+			// volume name already referenced
+			return vmList
+		}
+	}
+
+	return append(vmList, corev1.VolumeMount{
+		Name:      volumeName,
+		MountPath: mountPath,
+	})
+}
+
+// appendVolumes append volumes
+func (r *Reconciler) appendVolumes(volumeList []corev1.Volume, data map[string][]byte, volumeKeys []string, volumeName, secretName string) []corev1.Volume {
+	for _, vm := range volumeList {
+		if vm.Name == volumeName {
+			// volume name already referenced
+			return volumeList
+		}
+	}
+
+	items := []corev1.KeyToPath{}
+	for _, k := range volumeKeys {
+		items = append(items, corev1.KeyToPath{
+			Key:  k,
+			Path: k,
+		})
+	}
+
+	return append(volumeList, corev1.Volume{
+		Name: volumeName,
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: secretName,
+				Items:      items,
+			},
+		},
+	})
+}
+
 // Reconcile a ServiceBindingRequest by the following steps:
 // 1. Inspecting SBR in order to identify backend service. The service is composed by a CRD name and
 //    kind, and by inspecting "connects-to" label identify the name of service instance;
@@ -113,11 +156,24 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 				Info("Inspecting DeploymentConfig object...")
 
 			for i, c := range deploymentConfigObj.Spec.Template.Spec.Containers {
-				logger.Info("Adding EnvFrom to container")
-				deploymentConfigObj.Spec.Template.Spec.Containers[i].EnvFrom = r.appendEnvFrom(
-					c.EnvFrom, instance.GetName())
+				if len(retriever.data) > 0 {
+					logger.Info("Adding EnvFrom to container")
+					deploymentConfigObj.Spec.Template.Spec.Containers[i].EnvFrom = r.appendEnvFrom(
+						c.EnvFrom, instance.GetName())
+				}
+				if len(retriever.volumeKeys) > 0 {
+					logger.Info("Adding VolumeMounts to container")
+					mountPath := "/var/data"
+					if instance.Spec.MountPathPrefix != "" {
+						mountPath = instance.Spec.MountPathPrefix
+					}
+					deploymentConfigObj.Spec.Template.Spec.Containers[i].VolumeMounts = r.appendVolumeMounts(
+						c.VolumeMounts, instance.GetName(), mountPath)
+					logger.Info("Adding Volumes to pod")
+					deploymentConfigObj.Spec.Template.Spec.Volumes = r.appendVolumes(
+						deploymentConfigObj.Spec.Template.Spec.Volumes, retriever.data, retriever.volumeKeys, instance.GetName(), instance.GetName())
+				}
 			}
-
 			logger.Info("Updating DeploymentConfig object")
 			err = r.client.Update(ctx, &deploymentConfigObj)
 			if err != nil {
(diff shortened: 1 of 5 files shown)


sbose78:
@baijum @otaviof @pmacik @Avni-Sharma 

Please confirm that this is the implementation we are heading towards:

Based on,

```
urn:alm:descriptor:servicebinding:volumemount:secret:username
urn:alm:descriptor:servicebinding:volumemount:secret:username
```

The DeploymentConfig's container spec would look like:

```
  containers:
  - name: mypod
    image: redis
    volumeMounts:
    - name: foo
      mountPath: "/servicebindingCRname"
      readOnly: true
  volumes:
  - name: foo
    secret:
      secretName: service-binding-secret
      items:
      - key: username
        path: username
      - key: password
        path: password
```

This will enable these paths to be readable inside the container 
- /servicebindingCRname/username
- /servicebindingCRname/password

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -42,6 +43,40 @@ func (r *Reconciler) appendEnvFrom(envList []corev1.EnvFromSource, secret string
...
+	for _, vm := range volumeList {
+		if vm.Name == volumeName {
+			// volume name already referenced
+			return volumeList
+		}
+	}
+
+	return append(volumeList, corev1.Volume{
+		Name: volumeName,
+		VolumeSource: corev1.VolumeSource{
```
sbose78: What is the corresponding yaml/json structure for the `VolumeSource` , and why do we need it?

baijum: The YAML structure will be like this:
```
  volumes:
    - name: secret-volume
      secret:
        secretName: test-secret
```
This volume is referred from the volumeMount.

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -42,6 +43,40 @@ func (r *Reconciler) appendEnvFrom(envList []corev1.EnvFromSource, secret string
...
+	for _, vm := range vmList {
+		if vm.Name == volumeName {
+			// volume name already referenced
+			return vmList
+		}
+	}
+
+	return append(vmList, corev1.VolumeMount{
+		Name:      volumeName,
+		MountPath: "/" + volumeName,
```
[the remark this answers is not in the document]
baijum: yes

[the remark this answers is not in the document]
baijum: Hmm.. not exactly, it will be `/serviceBindingCRName-volumemount`

pkg/controller/servicebindingrequest/reconciler.go
baijum: I have added items. 935a50dc49bc80180f09481339f56367567c479a

pkg/controller/servicebindingrequest/reconciler.go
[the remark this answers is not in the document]
sbose78: @baijum  

May I suggest we improve the path prefix a bit ?

Let's add a an API attribute in `spec` to let the user specify the base mount path, because this is an information that the user of the ServiceBindingRequest can specify best.
 
```
kind: ServiceBindingRequest
metadata:
  name: binding-request
  namespace: shoubhik
spec:
  baseVolumePath:"/var/data"
  applicationSelector:
    matchLabels:
      connects-to: postgres
      environment: demo
    resourceKind: DeploymentConfig
  backingServiceSelector:
    resourceKind: postgresql.baiju.dev
    resourceName: Database
    resourceRef: db-demo
    resourceVersion: v1alpha1
```

Because the user has specified `baseVolumePath:"/var/redhat/credentials"`, any binding-driven volume paths would be 
- /var/redhat/credentials/username
- /var/redhat/credentials/password

If the user doesn't specify anything for `baseVolumePath`, then go for a default like
- /var/data/username
- /var/data/password . ( if there are any conventions on how a default data path should look like, go for that. Else, start with `/var/data/*` as a default.

pkg/controller/servicebindingrequest/volumemount/retriever.go
```diff
@@ -0,0 +1,198 @@
...
+	attrs := strings.SplitN(key, ".", 2)
+	newSectionMap, exists := sectionMap.(map[string]interface{})[attrs[0]]
+	if !exists {
+		return "", fmt.Errorf("Can't find '%v' section in CR", attrs)
+	}
+	return r.getNestedValue(attrs[1], newSectionMap.(map[string]interface{}))
+}
+
+// getCRKey retrieve key in section from CR object, part of the "plan" instance.
+func (r *Retriever) getCRKey(section string, key string) (string, error) {
```
sbose78: Isn't everything here a duplicate of `pkg/controller/servicebindingrequest/retriever.go`  ?

pkg/controller/servicebindingrequest/reconciler.go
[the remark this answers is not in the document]
baijum: Done. ff7be37a8cceb8949dc2fecfa4d06cdfcaa05160

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -116,8 +165,13 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 				logger.Info("Adding EnvFrom to container")
 				deploymentConfigObj.Spec.Template.Spec.Containers[i].EnvFrom = r.appendEnvFrom(
 					c.EnvFrom, instance.GetName())
+				logger.Info("Adding VolumeMounts to container")
+				deploymentConfigObj.Spec.Template.Spec.Containers[i].VolumeMounts = r.appendVolumeMounts(
+					c.VolumeMounts, instance.GetName(), instance.Spec.MountPathPrefix)
```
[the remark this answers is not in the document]
baijum: Done. ff7be37a8cceb8949dc2fecfa4d06cdfcaa05160

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -116,8 +165,13 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+				logger.Info("Adding Volumes to pod")
+				deploymentConfigObj.Spec.Template.Spec.Volumes = r.appendVolumes(
```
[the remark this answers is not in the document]
baijum: Done. 4a42c7415d3d6c7dd083711ced0e7efda53714f9

pkg/controller/servicebindingrequest/volumemount/retriever_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	t.Run("store", func(t *testing.T) {
+		retriever.store("test", []byte("test"))
+		assert.Contains(t, retriever.Data, "SERVICE_BINDING_DATABASE_TEST")
+		assert.Equal(t, []byte("test"), retriever.Data["SERVICE_BINDING_DATABASE_TEST"])
+	})
+
+	t.Run("saveDataOnSecret", func(t *testing.T) {
+		err := retriever.saveDataOnSecret()
+		assert.Nil(t, err)
```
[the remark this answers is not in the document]
baijum: This is for volume mount which gets data from a different secret object.

pkg/controller/servicebindingrequest/volumemount/retriever.go
Avni-Sharma: Yes, but here a new intermediary secret object is being created

baijum: Can we address this in a separate PR?

baijum:
/hold

I will wait for Shoubhik's approval

pkg/controller/servicebindingrequest/volumemount/retriever.go
sbose78: Sure!

sbose78:
@baijum Why are we using a different intermediary secret? If we used
```
volumes:
  - name: foo
    secret:
      secretName: service-binding-secret
      items:
      - key: username
        path: username
      - key: password
        path: password
```

We could ensure everything doesn't get mounted, and that should let us use a single intermediate secret for both volume mounts and `envFrom`... ?

baijum:
> @baijum Why are we using a different intermediary secret? If we used
> 
> ```
> volumes:
>   - name: foo
>     secret:
>       secretName: service-binding-secret
>       items:
>       - key: username
>         path: username
>       - key: password
>         path: password
> ```
> 
> We could ensure everything doesn't get mounted, and that should let us use a single intermediate secret for both volume mounts and `envFrom`... ?

Ok, I will change the code based on this suggestion here itself.

sbose78:
Thank you! Let me know if you bump into any design issues there.

baijum:
> Thank you! Let me know if you bump into any design issues there.

I have done this here: 6c768412f4c4f9bc271bf7cacd7552ae3101fddd

When creating env vars from secret there is no way to exclude certain values. So all values in the secret object will create an env var including that was meant for volume mount.

sbose78:
> When creating env vars from secret there is no way to exclude certain values. So all values in the secret object will create an env var including that was meant for volume mount.

Unwanted byproduct :)
but we can live with that for now!

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -144,9 +200,24 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+					logger.Info("Adding VolumeMounts to container")
+					mountPath := "/var/data"
+					if instance.Spec.MountPathPrefix != "" {
+						mountPath = instance.Spec.MountPathPrefix
+					}
+					deploymentObj.Spec.Template.Spec.Containers[i].VolumeMounts = r.appendVolumeMounts(
+						c.VolumeMounts, instance.GetName(), mountPath)
+					logger.Info("Adding Volumes to pod")
+					deploymentObj.Spec.Template.Spec.Volumes = r.appendVolumes(
+						deploymentObj.Spec.Template.Spec.Volumes, retriever.Data, retriever.VolumeKeys, instance.GetName(), instance.GetName()+"-volumemount")
```
sbose78: Isn't the name of the created intermediate secret `instance.GetName()` instead of `instance.GetName()+"-volumemount"` ?

baijum: Done. 32517eaefaab24f44ce41fc73bbe85604aac6faa

PR #82, opened 2019-07-22
Add test coverage via Codecov

pmacik:
This PR adds a code coverage reporting into tests.

At the end of test output it will print reports to the stdout similar to the following:

```
go tool cover -func=./out/coverage.test-unit.txt
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/common.go:9:                  RequeueOnNotFound       0.0%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/common.go:17:                 Requeue                 0.0%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/common.go:22:                 Done                    100.0%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/controller.go:16:             Add                     0.0%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/controller.go:21:             newReconciler           0.0%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/controller.go:26:             add                     0.0%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/planner/planner.go:40:        searchCRDDescription    77.3%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/planner/planner.go:87:        searchCR                88.9%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/planner/planner.go:111:       Plan                    71.4%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/planner/planner.go:128:       NewPlanner              100.0%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/reconciler.go:28:             appendEnvFrom           50.0%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/reconciler.go:55:             Reconcile               55.9%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/retriever.go:33:              getNestedValue          90.0%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/retriever.go:50:              getCRKey                87.5%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/retriever.go:67:              read                    88.2%
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/retriever.go:106:             extractSecretItemName   100.0%
[...]

PR #84, opened 2019-07-22
Install jq to root container for OpenShift CI.

pmacik:
This PR adds `jq` among the tools installed in the `root` container for OpenShift CI.
The `jq` is a dependency for https://github.com/redhat-developer/service-binding-operator/pull/82.

The tool needs to be added and merged in a separated PR prior merging the actual PR that depends on it, because of how OpenShift CI and Prow works.


sbose78:
Add a description please so that we know why :)

PR #92, opened 2019-07-30
Add environment variable prefix

baijum:
Fixes: https://jira.coreos.com/browse/ODC-1425

If prefix is not given, there won't be any prefix.

### DIFF
--- a/pkg/controller/servicebindingrequest/reconciler.go
+++ b/pkg/controller/servicebindingrequest/reconciler.go
@@ -116,7 +116,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return RequeueOnNotFound(err)
 	}
 
-	retriever := NewRetriever(ctx, r.client, plan, instance.Spec.MountPathPrefix)
+	retriever := NewRetriever(ctx, r.client, plan, instance.Spec.EnvVarPrefix)
 	if err = retriever.Retrieve(); err != nil {
 		return RequeueOnNotFound(err)
 	}
--- a/pkg/controller/servicebindingrequest/reconciler.go
+++ b/pkg/controller/servicebindingrequest/reconciler.go
@@ -116,7 +116,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return RequeueOnNotFound(err)
 	}
 
-	retriever := NewRetriever(ctx, r.client, plan)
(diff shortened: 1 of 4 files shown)


DhritiShikhar:
@baijum Can you add test for `No prefix` scenario?

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -116,7 +116,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return RequeueOnNotFound(err)
 	}
 
-	retriever := NewRetriever(ctx, r.client, plan)
+	retriever := NewRetriever(ctx, r.client, plan, instance.Spec.MountPathPrefix)
```
DhritiShikhar: @baijum Shouldn't this be `EnvVarPrefix` instead of `MountPathPrefix`?
considering
`MountPathPrefix is the prefix for volume mount` and `EnvVarPrefix is the prefix for environment variables`

or should it be both?

DhritiShikhar:
@baijum `envVarPrefix` should be removed from required criteria here -> 

https://github.com/redhat-developer/service-binding-operator/blob/master/deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml#L80

pkg/controller/servicebindingrequest/reconciler.go
baijum: Yes

baijum: Done. 01d0ba5ecdb225b20d6686d20af74bb02d5eac8d

baijum:
> @baijum `envVarPrefix` should be removed from required criteria here ->
> 
> https://github.com/redhat-developer/service-binding-operator/blob/master/deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml#L80

Should I send a separate PR for that?

baijum:
> > @baijum `envVarPrefix` should be removed from required criteria here ->
> > https://github.com/redhat-developer/service-binding-operator/blob/master/deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml#L80
> 
> Should I send a separate PR for that?

I sent a separate PR: #97

baijum:
> @baijum Can you add test for `No prefix` scenario?

Done. e4c1b9956049297f81b3f1bdf7838857835a544d

PR #98, opened 2019-08-05
Exposes service binding request status

DhritiShikhar:
Partially solves: https://jira.coreos.com/browse/ODC-1327

This PR exposes the binding status - `success`, `failure`, `inProgress`

To Do:
- [x] Add tests

Note:
This PR exposes only one status field `BindingStatus` . There will be follow up PR for exposing other fields.


sbose78:
To ensure that update of status don't trigger reconcile, you need to filter out that event by using Predicates

https://github.com/operator-framework/operator-sdk/blob/master/doc/user/event-filtering.md#using-predicates

sbose78:
.. ie, this https://github.com/redhat-developer/service-binding-operator/pull/99/files

PR #101, opened 2019-08-07
Add wait in reconcile loop.

akashshinde:
This PR adds utility functions to wait for k8s/Openshift resources.

**Why**
There are some gitOps use cases where `servicebindingrequest` would need to wait for resources(such as Deployment, backend Operator CR). 

**What is added**
This PR adds following two wait functions
1. `WaitUntilResourceFound` -> Function to wait for particular resource to be created. 
2. `WaitUntilResourcesFound` -> Function to wait for list of resources to be created.

**To Do**

- [x] Remove utils package (we don't need `utils` package as per @sbose78's suggestion).

### DIFF
--- a/pkg/controller/servicebindingrequest/reconciler.go
+++ b/pkg/controller/servicebindingrequest/reconciler.go
@@ -137,7 +137,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 
 	// FIXME: find a way to DRY this block, and then add statefulsets and other kinds back again;
-		switch resourceKind {
+	switch resourceKind {
 	case "deploymentconfig":
 		logger.Info("Searching DeploymentConfig objects matching labels")
 
--- a/pkg/controller/servicebindingrequest/reconciler.go
+++ b/pkg/controller/servicebindingrequest/reconciler.go
@@ -2,7 +2,7 @@ package servicebindingrequest
 
 import (
 	"context"
-	"github.com/redhat-developer/service-binding-operator/pkg/utils"
+	"github.com/redhat-developer/service-binding-operator/pkg/resourcepoll"
 	"strings"
 
 	osappsv1 "github.com/openshift/api/apps/v1"
@@ -142,7 +142,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		logger.Info("Searching DeploymentConfig objects matching labels")
 
 		deploymentConfigListObj := &osappsv1.DeploymentConfigList{}
-		err = utils.WaitUntilResourcesFound(r.client, &searchByLabelsOpts, deploymentConfigListObj)
+		err = resourcepoll.WaitUntilResourcesFound(r.client, &searchByLabelsOpts, deploymentConfigListObj)
 		if err != nil {
 			return RequeueOnNotFound(err)
 		}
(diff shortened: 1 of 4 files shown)


pkg/utils/wait_utils.go
```diff
@@ -0,0 +1,37 @@
+package utils
```
sbose78: A `utils` package is big NO :) because it will soon become a kitchen sink of unrelated things. Feel free to create a  `pkg/resourcepoll/wait.go`

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -141,6 +142,10 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		logger.Info("Searching DeploymentConfig objects matching labels")
 
 		deploymentConfigListObj := &osappsv1.DeploymentConfigList{}
+		err = utils.WaitUntilResourcesFound(r.client, &searchByLabelsOpts, deploymentConfigListObj)
```
DhritiShikhar: shouldn't this be added to `default` case as well in addition to `deploymentconfig` switch case?

pkg/utils/wait_utils.go
akashshinde: @sbose78 makes sense

pkg/controller/servicebindingrequest/reconciler.go
akashshinde: @DhritiShikhar yes need to add wait to the `default` case

akashshinde: Done https://github.com/redhat-developer/service-binding-operator/pull/101/commits/78eb64a5588760d876c29374da182335e4a030b0

pkg/utils/wait_utils.go
akashshinde: Done https://github.com/redhat-developer/service-binding-operator/pull/101/commits/dafb96619aa2a0fc10b3bc2e2637222a85cd673e

DhritiShikhar:
@akashshinde can you add tests for the same?

DhritiShikhar:
As per discussion with @akashshinde there will be another PR with tests. So, approving this one.

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -166,6 +167,10 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		logger.Info("Searching DeploymentConfig objects matching labels")
 
 		deploymentConfigListObj := &osappsv1.DeploymentConfigList{}
+		err = resourcepoll.WaitUntilResourcesFound(r.client, &searchByLabelsOpts, deploymentConfigListObj)
```
isutton: I assume that the reason this was put in place is because it indeed can take some time for the deployment config objects to appear, but now every time this gets executed it'll wait at least the first interval for each worker, which seems to block the whole process for each reconcile attempt.

A somehow easier approach would be a delayed re-enqueue at this point, pretty much moving the waiting process from inside the process to the queue itself (as in re-enqueue the ServiceBindingRequest object in 5 seconds if related objects are not present). 

Yet another approach would be listen for changes in DeploymentConfig at the operator level, and enqueue the related ServiceBindingRequest once a relevant object comes from the stream. This is more work, but also means that all the events of related objects are processed, and work on ServiceBindingRequest reconciliation would be re-attempted automatically.

DhritiShikhar: +1 for the delayed re-enqueue approach

PR #103, opened 2019-08-08
Adds secret name in service binding status

DhritiShikhar:
Partially solves: https://jira.coreos.com/browse/ODC-1327

This PR exposes the secret name in service binding status.

ToDo:
- [x] Update README

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,7 @@ spec:
 
 | Field | Description |
 |-------|-------------|
-| Status | the binding status of Service Binding Request |
+| BindingStatus | The binding status of Service Binding Request |
+| Secret | The name of the intermediate secret |
 
 That's it, folks!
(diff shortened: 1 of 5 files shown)


baijum:
Do we need this as it is always the name of CR.

sbose78:
Yes
- to inform the user that we created a `secret`
- we might change the naming convention in future to accommodate conflicts.

README.md
```diff
@@ -276,6 +276,7 @@ spec:
 
 | Field | Description |
 |-------|-------------|
-| Status | the binding status of Service Binding Request |
+| BindingStatus | The binding status of Service Binding Request |
+| Secret | The name of the intermediate secret |
```
sbose78: Thanks for this @DhritiShikhar . Could you also include a CR yaml in the readme which has the `status` populated?

DhritiShikhar: @sbose78 Yes. I am planning to do that. But I don't want to do it as a part of this PR. I will raise another PR for the changes in README.

PR #105, opened 2019-08-09
Adds ApplicationObjects in status

DhritiShikhar:
This status field has all the objects that were found for that label.

Partially fixes ODC-1327.

Note:
Changes in README will be in a followup PR.

### DIFF
--- a/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
+++ b/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
@@ -84,6 +84,8 @@ const (
 type ServiceBindingRequestStatus struct {
 	// BindingStatus is the status of the service binding request. Possible values are Success, Failure, InProgress.
 	BindingStatus BindingStatus `json:"BindingStatus"`
+	// LabelObjects contains all the objects with the label
+	LabelObjects []string `json:LabelObjects`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
+++ b/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
@@ -87,7 +87,7 @@ type ServiceBindingRequestStatus struct {
 	// Secret is the name of the intermediate secret
 	Secret string `json:"Secret"`
 	// LabelObjects contains all the objects with the label
-	LabelObjects []string `json:LabelObjects`
+	LabelObjects []string `json:"LabelObjects"`
 }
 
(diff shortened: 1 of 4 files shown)


pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -84,6 +84,8 @@ const (
 type ServiceBindingRequestStatus struct {
 	// BindingStatus is the status of the service binding request. Possible values are Success, Failure, InProgress.
 	BindingStatus BindingStatus `json:"BindingStatus"`
+	// LabelObjects contains all the objects with the label
+	LabelObjects []string `json:LabelObjects`
```
golangcibot: structtag: struct field tag `json:LabelObjects` not compatible with reflect.StructTag.Get: bad syntax for struct tag value (from `govet`)

akashshinde: @DhritiShikhar not sure which json key format we have been following but keys should be `camel-case` or `snake-case`, I think that is the reason `govet` is also not happy 😃

DhritiShikhar: @akashshinde `govet` is not happy because the struct tag was wrong. I already fixed it here a86de53

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -86,6 +86,8 @@ type ServiceBindingRequestStatus struct {
 	BindingStatus BindingStatus `json:"BindingStatus"`
 	// Secret is the name of the intermediate secret
 	Secret string `json:"Secret"`
+	// LabelObjects contains all the objects with the label
+	LabelObjects []string `json:"LabelObjects"`
```
baijum: How about naming it as `ApplicationObjects`?

DhritiShikhar: @baijum hmm.. that's a better word..

DhritiShikhar: done here 
5fe2f13

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
akashshinde: ah okay

sbose78:
Whenever we update the API spec or status , please include an example in the Description of the PR and preferably the Readme as well :)

DhritiShikhar:
Sure @sbose78 I will add a description in the PR. And I am working on the README changes.

PR #120, opened 2019-08-14
Add lint back

sbose78:
- https://bugzilla.redhat.com/show_bug.cgi?id=1739804
- Pushed branch to redhat-developer/.. so that it triggers a build on https://quay.io/repository/redhat-developer/app-binding-operator/build which is driven from the `/Dockerfile`


sbose78:
Quay is green 

image

PR #123, opened 2019-08-14
Use Group/Version/Kind for both ApplicationSelector & BackingServiceSelector

sbose78:
The modified API would look like:

```
---
kind: ServiceBindingRequest
metadata:
spec:
  mountPathPrefix: "/var/credentials"
  backingServiceSelector:
    resourceRef: pg-instance
    group: postgresql.example.dev
    kind: Database
    version: v1alpha1
  applicationSelector:
    matchLabels:
      connects-to: postgres
      environment: production
    group: apps
    kind: Deployment
    version: v1
```

**What and Why**

- Made the API consistent with the conventional use of the terms 'group', 'version' and 'kind'
- made `Group` optional for ease of use.
- Commented out the CR creation on `make local` as it is unnecessary.

fixes https://jira.coreos.com/browse/ODC-1497


Makefile
```diff
@@ -284,7 +284,7 @@ push-image: build-image
 
 .PHONY: local
 ## Local: Run operator locally
-local: deploy-clean deploy-rbac deploy-crds deploy-cr
+local: deploy-clean deploy-rbac deploy-crds #deploy-cr
```
otaviof: I think you can really remove this target, instead of just commenting it out.

sbose78: True 🤖

sbose78: Done https://github.com/redhat-developer/service-binding-operator/pull/123/commits/a550fa1817e6f8d293f2354344619909fac1e445

PR #124, opened 2019-08-15
Add a Binder which uses a dynamic client to CRUD deployment objects

otaviof:
* Introducing Kubernetes dynamic API client;
* Making generic support for application formats in `Binder` (new component); 
* refactoring code and tests to reduce code-duplication, introducing `test/mocks/fake.go`;
* Refactoring `Planner` to only use dynamic client;
* Removing `wait.go` to use native approach based on `RequeueAfter`;

### Using `resource` in `ApplicationSelector`

Updating CR to have `resource` instead of `kind` in `ApplicationSelector`. Therefore `Binder` would not have to guess anything anymore. For instance, please consider the following example:

``` yaml
---
kind: ServiceBindingRequest
metadata:
spec:
  [...]
  applicationSelector:
    matchLabels:
      connects-to: postgres
      environment: production
    group: apps
    version: v1
    resource: deployments
```

### Next Steps

Migrate all components to rely in the dynamic client and remove previous client references.


sbose78:
After adding a `watch` on the secret

```
secretResource := schema.GroupVersionResource{corev1.GroupName, corev1.SchemeGroupVersion.Version, "Secret"}
secretClient := r.dynClient.Resource(secretResource).Namespace(r.plan.Ns)
secretWatcher, err := secretClient.Watch(metav1.ListOptions{Watch: true, FieldSelector: fmt.Sprintf("metadata.name=%s", name)})
ch := secretWatcher.ResultChan()
for event := range ch {
	_, ok := event.Object.(*corev1.Secret)
	if !ok {
	    logger.Error("could not figure out what triggered this event")
	}
}
// add something to the queue
```

I'm curious, what do you plan on doing here?  :)
- would you be calling r.reconcile(..) directly after getting the object into this..
- would you be causing a dirty update to the ServiceBindingRequest CR to force a reconcile..
- would you be using the workqueue api directly?

sbose78:
Wondering how to ensure we don't keep adding watches - because we aren't keeping a track of the initial watch being added.

pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/dynamic"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	"github.com/redhat-developer/service-binding-operator/test/mocks"
+)
+
+var binderFakeClient client.Client
```
golangcibot: `binderFakeClient` is unused (from `deadcode`)

pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -0,0 +1,60 @@
...
+var binderFakeDynClient dynamic.Interface
```
golangcibot: `binderFakeDynClient` is unused (from `deadcode`)

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	return append(volumes, u), nil
+}
+
+// updateSpecContainers extract containers from object, and trigger update.
+func (b *Binder) updateSpecContainers(
+	logger logr.Logger,
+	obj *ustrv1.Unstructured,
+) (*ustrv1.Unstructured, error) {
+	containersPath := []string{"spec", "template", "spec", "containers"}
```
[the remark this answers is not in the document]
otaviof: @sbose78 yep, those kind are now supported since the internal exchange format is `Unstructured` and thus we can add a simple `if` block to address resources that would not follow the same structured.

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -0,0 +1,295 @@
...
+func (b *Binder) getResourceKind() string {
+	return strings.ToLower(b.sbr.Spec.ApplicationSelector.ResourceKind)
+}
+
+// getListGVR returns a GVR for a list of resources.
+func (b *Binder) getListGVR() (schema.GroupVersionResource, error) {
+	kind := b.getResourceKind()
+	switch kind {
+	case "deploymentconfig":
+		return osappsv1.Resource("deploymentconfigs").WithVersion("v1"), nil
```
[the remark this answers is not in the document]
sbose78: Is there no way to not hardcode `"deploymentconfigs"` ? :(

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -21,76 +21,40 @@ type ServiceBindingRequestSpec struct {
...
 // BackingServiceSelector defines the selector based on resource name, version, and resource kind
 // +k8s:openapi-gen=true
 type BackingServiceSelector struct {
-	Group       *string `json:"group,omitempty"`
-	Version     string  `json:"version"`
-	Kind        string  `json:"kind"`
-	ResourceRef string  `json:"resourceRef"`
+	Group       string `json:"group,omitempty"`
+	Version     string `json:"version"`
+	Kind        string `json:"kind"`
```
sbose78: You could replace this with `Resource` here as well, since you are going to need it for https://github.com/sbose78/service-binding-operator/pull/2/files#diff-42835fd3b25b196b1bbc21495e0ac5b7R149 . I  could do in a different PR as well - as you wish :)

pkg/controller/servicebindingrequest/planner.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	p.logger.Info("Warning: not able to find a CRD description!")
+	return nil, errors.NewNotFound(olmv1alpha1.Resource(csvResourceName), "")
+}
+
+// extractCRDDescription auxiliary method to identify the CRD-Description object the
+// BackingServiceSelector is looking for, otherwise returns nil.
+func (p *Planner) extractCRDDescription(
+	logger logr.Logger,
```
golangcibot: SA4009: argument logger is overwritten before first use (from `staticcheck`)

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
otaviof: @sbose78 please check the approach in `planner.go` in order to exchange a GVK for a GVR. Both scenarios, on going from GVK to GVR or vice-versa would be the same effort. (cc @isutton)

sbose78: We are going to need the `Resource for `crResource := schema.GroupVersionResource{plan.CR.GroupVersionKind().Group, plan.CR.GroupVersionKind().Version,"databases"` 

`Kind` unfortunately doesn't help. But we can skip making that change in this PR as it is not needed here :)

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -21,76 +21,40 @@ type ServiceBindingRequestSpec struct {
...
+	ResourceRef string `json:"resourceRef"`
 }
 
-// ApplicationSelector defines the selector based on labels and resource kind
+// ApplicationSelector defines the selector based on labels and GVR
 // +k8s:openapi-gen=true
 type ApplicationSelector struct {
 	MatchLabels map[string]string `json:"matchLabels"`
-	Group       *string           `json:"group,omitempty"`
+	Group       string            `json:"group,omitempty"`
```
sbose78: Why did we remove the pointer? :) It's typically a good way to say if a specific field is optional hence nulllable?
Or, did you make `Group` mandatory in this? I guess the latter.. ah Yes.
( sorry thought out aloud :) )

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -112,7 +76,8 @@ type ServiceBindingRequest struct {
 type ServiceBindingRequestList struct {
 	metav1.TypeMeta `json:",inline"`
 	metav1.ListMeta `json:"metadata,omitempty"`
-	Items           []ServiceBindingRequest `json:"items"`
+
+	Items []ServiceBindingRequest `json:"items"`
```
sbose78: What can we do on our vscode to ensure it doesn't change again? :)

otaviof: @sbose78 I'm afraid because I added a new line, to follow the same standard than on Spec field, the Go-Format took care of setting up the code again. So as long as we don't change the surroundings anymore, that should not change. :-)

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
otaviof: @sbose78 if the field is nullable or empty, should we then use `omitempty`? I thought the pointer reference over there were a typo, since no other fields were doing the same, and examples using Operator-SDK does not show that in a String.

sbose78: The pointer reference is for use in the code - During deserialization if the `Group` in the yaml is empty, the ` Group` in the struct would be nil if we specify it as a pointer  :) 
That along with the `omitempty` is to tell `operator-sdk generate openapi` that the specific field is to be marked as non-required in the openapi validation spec.

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -0,0 +1,280 @@
...
+	sbr        *v1alpha1.ServiceBindingRequest // instantiated service binding request
+	volumeKeys []string                        // list of key names used in volume mounts
+	logger     logr.Logger                     // logger instance
+}
+
+// search objects based in Kind/APIVersion, which contain the labels defined in ApplicationSelector.
+func (b *Binder) search() (*ustrv1.UnstructuredList, error) {
+	ns := b.sbr.GetNamespace()
+	gvr := schema.GroupVersionResource{
+		Group:    b.sbr.Spec.ApplicationSelector.Group,
```
sbose78: @otaviof , is the value of `b.sbr.Spec.ApplicationSelector.Group` mandatory ?

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -0,0 +1,280 @@
...
+	matchLabels := b.sbr.Spec.ApplicationSelector.MatchLabels
+	opts := metav1.ListOptions{
+		LabelSelector: labels.Set(matchLabels).String(),
+	}
+
+	return b.dynClient.Resource(gvr).Namespace(ns).List(opts)
+}
+
+// updateSpecVolumes execute the inspection and update "volumes" entries in informed spec.
+func (b *Binder) updateSpecVolumes(
```
sbose78: @baijum Could you please review ?

pkg/controller/servicebindingrequest/planner.go
```diff
@@ -0,0 +1,177 @@
...
+
+	logger := p.logger.WithValues(
+		"CSV.Group", gvr.Group,
+		"CSV.Version", gvr.Version,
+		"CSV.Resource", gvr.Resource,
+		"CSV.Namespace", ns,
+	)
+	logger.Info("Searching for ClusterServiceVersion...")
+
+	// FIXME: usually the CSV resources are in the "openshift-operator-lifecycle-manager" namespace;
```
sbose78: Umm. no. The CSVs which a user wants to use would be copied to the user's namespace. Most CSVs get copied over to all namespaces.

sbose78:
Why do we have the requirement of Group Version Resource while looking up deploymentConfig / deployments ?

```

// search objects based in Kind/APIVersion, which contain the labels defined in ApplicationSelector.
func (b *Binder) search() (*ustrv1.UnstructuredList, error) {
	ns := b.sbr.GetNamespace()
	gvr := schema.GroupVersionResource{
		Group:    b.sbr.Spec.ApplicationSelector.Group,
		Version:  b.sbr.Spec.ApplicationSelector.Version,
		Resource: b.sbr.Spec.ApplicationSelector.Resource,
	}
	matchLabels := b.sbr.Spec.ApplicationSelector.MatchLabels
	opts := metav1.ListOptions{
		LabelSelector: labels.Set(matchLabels).String(),
	}

	return b.dynClient.Resource(gvr).Namespace(ns).List(opts)
}
```

but for looking up CRs, we do use Group,Version & Kind ?
```
func (p *Planner) searchCR(kind string) (*ustrv1.Unstructured, error) {
	bss := p.sbr.Spec.BackingServiceSelector
	gvk := schema.GroupVersionKind{Group: bss.Group, Version: bss.Version, Kind: bss.Kind}
	gvr, _ := meta.UnsafeGuessKindToResource(gvk)
	opts := metav1.GetOptions{}

	logger := p.logger.WithValues("CR.GVK", gvk.String(), "CR.GVR", gvr.String())
	logger.Info("Searching for CR instance...")

	cr, err := p.client.Resource(gvr).Namespace(p.sbr.GetNamespace()).Get(bss.ResourceRef, opts)
```


Can we use GVR in both places ( ie ApplicationSelector & BackingServiceSelector in the API )  ?

sbose78:
OK, we should not use `UnsafeGuessKindToResource` https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/restmapper.go#L123

sbose78:
Ah because the csv has a `Kind`. Damn :)

```
customresourcedefinitions:
    owned:
    - name: apicuritos.apicur.io
      version: v1alpha1
      kind: Apicurito
      displayName: Apicurito CRD
      description: CRD for Apicurito
```

sbose78:
@pmacik 

So, this is how the minimal set of fields look like :


```
kind: ServiceBindingRequest
metadata:
  name: binding-request
  namespace: sbose
spec:
  applicationSelector:
    matchLabels:
      connects-to: postgres
      environment: demo
    group: apps.openshift.io
    version: v1
    resource: deploymentconfigs
  backingServiceSelector:
    group: postgresql.baiju.dev
    kind: Database
    version: v1alpha1
    resourceRef: demo-database
```

sbose78:
Tested end to end by doing a 

1. `make local`
2. Import app
3. Created a Database CR
3. Created a ServiceBindingRequest
```
kind: ServiceBindingRequest
metadata:
  name: binding-request
  namespace: sbose
spec:
  applicationSelector:
    matchLabels:
      connects-to: postgres
      environment: demo
    group: apps.openshift.io
    version: v1
    resource: deploymentconfigs
  envVarPrefix: "shoubhik"
  backingServiceSelector:
    group: postgresql.baiju.dev
    kind: Database
    version: v1alpha1
    resourceRef: demo-database
 ```
And verified that 
1. Secret was created with
2. Binding request secret was part of DC

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
otaviof: Agreed. It's a good idea making this change when we are touching that part.

PR #126, opened 2019-08-17
Reconcile if the BackingService-owned secret/configmap changes

sbose78:
This PR is an alternate way to achieve the goal of triggering a reconcile when something on the BackingService side changes.

Example, if the db credentials stored in a secret or a configmap owned by the Database CR changes, a reconcile is triggered. 

:white_check_mark: Support for watching the database CR's secret

:white_check_mark: Support for watching the database CR's configmap

::heavy_exclamation_mark: Support for watching changes in the database CR itself.

The above non-supported item is because the following isn't supported - ie, one can't pass an Unstructured without defining the `Kind`. In our case, the `Kind` of a CR isn't known at the controller boot time.
```
c.Watch(&source.Kind{Type: &ustrv1.Unstructured{Object: map[string]interface{}{}, customHandler)
```

Here's how I've tested this PR so far:
1. Install the PostgreSQL operator.

2. Create a DB 
```
apiVersion: postgresql.baiju.dev/v1alpha1
kind: Database
metadata:
  name: db-demo
spec:
  image: docker.io/postgres
  imageName: postgres
  dbName: db-demo
```

3. Import from Git : https://github.com/sclorg/nodejs-ex and edit labels in the DeploymentConfig to have `connects-to: postgres` , `environment: demo`

4. Do a `make local` and then create a `ServiceBindingRequest`

```
kind: ServiceBindingRequest
metadata:
  name: binding-request
  namespace: sbose
spec:
  applicationSelector:
    matchLabels:
      connects-to: postgres
      environment: demo
    kind: DeploymentConfig
    
    version: v1
  backingServiceSelector:
    kind: Database
    version: v1alpha1
    resourceRef: demo-database
    group: postgresql.baiju.dev
```

5. Wait for success, and ensure that the intermediate secret named `binding-request` is created by the controller.

6. Modify the `secret` named `demo-database-postgresql` and watch the  `binding-request` secret get updated automatically. :white_check_mark:

7. Modify the `configmap` named `demo-database` and watch the  `binding-request` secret get updated automatically. :white_check_mark:


ldimaggi:
Looks good - note that that Postgres DB's operator was moved to the samples repo - see: 
 https://github.com/redhat-developer/service-binding-operator/pull/137

PR #127, opened 2019-08-18
Force triggering of a deployment whenever reconcile causes change in bound data

sbose78:
### Why
If the binding secret is updated, the event is not propagated to the DeploymentConfig/Deployment which reads from it. Hence, without a trigger to re-deploy the imported app, the modified binding data doesn't get into the app or the pod.

### What & How
On reconcile, add/update an environment variable called `lastboundtime` 

```
containers:
        - resources: {}
          env:
            - name: lastbound
              value: '2019-08-19 18:01:23.010807 -0400 EDT m=+105.168405731'
          envFrom:
            - secretRef:
                name: binding-request
```
Given that this time is always going to change whenever a reconcile tries to update the deploymentconfig/deployment. 

### Future improvements
Currently, it always updates on reconcile. In the near future, trigger redeploy only if data has really changed. It is fine for now, because a reconcile is triggered only if specific objects ( SBR ) change.

### Merge readiness

- [x] Manually tested
- [x] New tests


pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -229,6 +231,38 @@ func (r *Retriever) Retrieve() error {
...
+	for _, key := range sortedEnvKeys {
+		concatString = concatString + string(r.data[key])
+	}
+
+	return strings.TrimSpace(md5Hash(concatString))
+}
+
+func md5Hash(s string) string {
+	h := sha1.New()
+	h.Write([]byte(s))
```
golangcibot: Error return value of `h.Write` is not checked (from `errcheck`)

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -239,15 +253,15 @@ func (b *Binder) update(objList *ustrv1.UnstructuredList) ([]string, error) {
...
 
-		logger.Info("Updating object...")
+		logger.Info("Updating object in Kube...")
 		if err := b.client.Update(b.ctx, updatedObj); err != nil {
 			return nil, err
 		}
 
 		// recording object as updated
 		updatedObjectNames = append(updatedObjectNames, name)
+		logger.Info(updatedObjectNames[0])
```
akashshinde: If there are no updated objects, `updatedObjectNames` would be empty in that case. In that case here we are accessing the first element from the slice which would cause panic. 
cc: @sbose78

sbose78: True!

sbose78: Removed it :) Was pretty unnecessary.

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -189,10 +201,12 @@ func (b *Binder) updateContainer(container interface{}) (map[string]interface{},
 	}
 	// effectively binding the application with intermediary secret
 	c.EnvFrom = b.appendEnvFrom(c.EnvFrom, b.sbr.GetName())
+	c.Env = b.appendEnvVar(c.Env, lastboundparam, time.Now().Format(time.RFC3339))
```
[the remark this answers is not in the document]
sbose78: yes!

PR #132, opened 2019-08-20
Make the annotations shorter

sbose78:
This was a feedback we received from Jessica during an API review. We don't really need the long strings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,13 +51,13 @@ statusDescriptors:
     path: dbCredentials
     x-descriptors:
       - urn:alm:descriptor:io.kubernetes:Secret
-      - urn:alm:descriptor:servicebindingrequest:env:object:secret:user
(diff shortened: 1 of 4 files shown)


sbose78:
CC @pmacik @otaviof

sbose78:
Added the hold so that we could get Pavel's CR updated first :)

pmacik:
> Added the hold so that we could get Pavel's CR updated first :)

@sbose78 I merged the updates. You can go ahead and un-hold this :)

sbose78:
@pmacik Would the latest `OperatorSource` of the two sample operators reflect these changes?

pmacik:
> @pmacik Would the latest `OperatorSource` of the two sample operators reflect these changes?

@sbose78 Yes, both sample operators were updated and released to quay

PR #133, opened 2019-08-20
Fix PascalCase in constant values stored in ServiceBindRequest's Status field

isutton:
Convert texts like `inProgress` to `InProgress`


sbose78:
thank you @isutton !

/ok-to-test

Issue #148, opened 2019-08-23
Trigger re-binding if the BackingService CR's "associated" secret changes


sbose78:
### Description

A BackingService like a Database managed by a Database CR might _use_/_refer to_ a `secret` or a `configmap` which stores credentials or connection information on how to connect to the database. When such objects change, a rebinding ( "reconcile" ) should be triggered.

### What does a BackingService CR associated object imply ?

By "Database CR being **associated** with a secret", it is implied that the `secret` or `configmap` has an **annotation** containing the details of the ServiceBindingRequest which consumed it as part of binding an app to a specific Backing Service CR.

### Acceptance Criteria

Ensure the following scenario works:

- Create a Database CR.
- Import App
- Create a Binding request
- Validate the the imported app's CRUD works.
- Validate that the ServiceBindingRequest **has set an annotation** on all secrets/configmaps that were to referred to, in the Database CR
- Modify credentials in the backing service using **an update to the secret associated** by the Database CR.
- Ensure that reconcile is triggered and the imported app's CRUD works with the new credentials.

### Hint
- Today, annotations in the Database CR's secrets/configmaps are not set in the `binder`. We should be doing so as part of this story.

Issue #147, opened 2019-08-23
Trigger rebinding if the BackingServiceCR-owned secret/configmap changes


sbose78:
### Description

A BackingService like a Database managed by a Database CR might own a secret or a configmap which stores credentials or connection information on how to connect to the database. 

### What does a BackingService CR owned object imply ?

By "Database CR **owning** a secret", it is implied that the secret or configmap has an **ownerReferences** containing the details of the Database CR which created it.

If such objects change, a rebinding ( "reconcile" ) should be triggered. 

### Acceptance Criteria

Ensure the following scenario works:

- Create a Database CR.
- Import App
- Create a Binding request
- Validate the the imported app's CRUD works.
- Modify credentials in the backing service using an update to the secret **owned** by the Database CR.
- Ensure that reconcile is triggered and the imported app's CRUD works with the new credentials.

sbose78:
Partially done in https://github.com/redhat-developer/service-binding-operator/pull/126

PR #149, opened 2019-08-23
Watch Resources to Trigger Rebinding

otaviof:
This PR will create a workflow where the controller watches for all "interesting" types of GVKs, and using `EnqueueRequestsFromMapFunc` (`controller-runtime/handler`) to always extract a `ServiceBindingRequest` request from those.

To identify the namespaced-name of an `ServiceBindingRequest`, this PR rely in annotations, and also checks if the given resource is actually a SBR. At the end of a binding workflow, after actions are executed successfully, related resources are annotated to identify from which `ServiceBindingRequest` instance it belongs to.

#### TODOs

- [ ] Review controller predicate, do we want to reconcile when annotations are found, or when generation has changed (default behavior);
- [ ] Improve getWatchingGVKs` to allow more GVKs, review the logic we would like to apply;

#### Fixes

- #147 
- #148 

### Collateral Effects

During coding some components were refactored to use a dynamic client, since during testing the objects were not visible by "static" and dynamic clients at the same time. Since they do not share the same cache.

### DIFF
--- a/pkg/controller/servicebindingrequest/annotations.go
+++ b/pkg/controller/servicebindingrequest/annotations.go
@@ -1,0 +1,116 @@
+package servicebindingrequest
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
+
+	"github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1"
+)
+
+const (
+	sbrNamespaceAnnotation = "service-binding-operator.apps.openshift.io/binding-namespace"
+	sbrNameAnnotation      = "service-binding-operator.apps.openshift.io/binding-name"
+)
+
+// extractNamespacedName returns a types.NamespacedName if the required service binding request keys
+// are present in the given data
+func extractNamespacedName(data map[string]string) types.NamespacedName {
+	namespacedName := types.NamespacedName{}
+	ns, exists := data[sbrNamespaceAnnotation]
+	if !exists {
+		return namespacedName
+	}
+	name, exists := data[sbrNameAnnotation]
+	if !exists {
+		return namespacedName
+	}
+	namespacedName.Namespace = ns
+	namespacedName.Name = name
+	return namespacedName
+}
+
+// GetSBRNamespacedNameFromObject returns a types.NamespacedName if the required service binding
+// request annotations are present in the given runtime.Object, empty otherwise. When annotations are
+// not present, it checks if the object is an actual SBR, returning the details when positive. An
+// error can be returned in the case the object can't be decoded.
+func GetSBRNamespacedNameFromObject(obj runtime.Object) (types.NamespacedName, error) {
+	namespacedName := types.NamespacedName{}
+	data, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		return namespacedName, err
+	}
+
+	u := &unstructured.Unstructured{Object: data}
+
+	namespacedName = extractNamespacedName(u.GetAnnotations())
+	logger := log.WithValues(
+		"Resource.GVK", u.GroupVersionKind(),
+		"Resource.Namespace", u.GetNamespace(),
+		"Resource.Name", u.GetName(),
+		"Extracted.NamespacedName", namespacedName.String(),
+	)
+	if !IsSBRNamespacedNameEmpty(namespacedName) {
+		logger.Info("Not able to define SBR namespaced-name based on annotations!")
+		return namespacedName, nil
+	}
+
+	if u.GroupVersionKind() != v1alpha1.SchemeGroupVersion.WithKind("ServiceBindingRequest") {
+		logger.Info("Object is also not a SBR resource type.")
+		return namespacedName, nil
+	}
+
+	logger.Info("Creating namespaced-name for a actual SBR object.")
+	namespacedName.Namespace = u.GetNamespace()
+	namespacedName.Name = u.GetName()
+	return namespacedName, nil
+}
+
+// IsSBRNamespacedNameEmpty returns true if any of the fields from the given namespacedName is empty.
+func IsSBRNamespacedNameEmpty(namespacedName types.NamespacedName) bool {
+	return namespacedName.Namespace == "" || namespacedName.Name == ""
+}
+
+// SetSBRAnnotations update existing annotations to include operator's. The annotations added are
+// referring to a existing SBR namespaced name.
+func SetSBRAnnotations(
+	ctx context.Context,
+	client dynamic.Interface,
+	namespacedName types.NamespacedName,
+	objs []*unstructured.Unstructured,
+) error {
+	for _, obj := range objs {
+		annotations := obj.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+
+		annotations[sbrNamespaceAnnotation] = namespacedName.Namespace
+		annotations[sbrNameAnnotation] = namespacedName.Name
+		obj.SetAnnotations(annotations)
+
+		gvk := obj.GroupVersionKind()
+		gvr, _ := meta.UnsafeGuessKindToResource(gvk)
+		opts := metav1.UpdateOptions{}
+
+		logger := log.WithValues(
+			"SBR.Namespace", namespacedName.Namespace,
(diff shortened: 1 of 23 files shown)


pkg/controller/servicebindingrequest/annotations_test.go
```diff
@@ -0,0 +1,51 @@
...
+	assert.Nil(t, err)
+	assert.Equal(t, types.NamespacedName{Namespace: "ns", Name: "name"}, namespacedName)
+
+	sbr := &unstructured.Unstructured{}
+	sbr.SetGroupVersionKind(v1alpha1.SchemeGroupVersion.WithKind("ServiceBindingRequest"))
+	sbr.SetNamespace("ns")
+	sbr.SetName("name")
+
+	// it should also understand a actual SBR as well, so return not empty
+	namespacedName, err = GetSBRNamespacedNameFromObject(sbr.DeepCopyObject())
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/controller/servicebindingrequest/annotations.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	namespacedName.Namespace = ns
+	namespacedName.Name = name
+	return namespacedName
+}
+
+// GetSBRNamespacedNameFromObject returns a types.NamespacedName if the required service binding
+// request annotations are present in the given runtime.Object, empty otherwise. An error can be
+// returned in the case the object can't be decoded.
+func GetSBRNamespacedNameFromObject(obj runtime.Object) (types.NamespacedName, error) {
```
isutton: Perhaps we can change the signature to return `*types.NamespacedName` instead, since it is expected that no error has happened and the object doesn't refer to any ServiceBindingRequest. This means that `return nil, nil` is a valid return indicating that no error has happened but no value has been found in the given `obj`.

This would influence the `IsSBRNamespaceNameEmpty` method, where a pointer would be used as input and the check for null-ability performed in addition to the existing checks.

pkg/controller/servicebindingrequest/controller.go
```diff
@@ -36,70 +34,58 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
...
-				return true
-			}
-
-			// Ignore updates to CR status in which case metadata.Generation does not change
+			// ignore updates to CR status in which case metadata.Generation does not change
 			return e.MetaOld.GetGeneration() != e.MetaNew.GetGeneration()
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
-			// Evaluates to false if the object has been confirmed deleted.
+			// evaluates to false if the object has been confirmed deleted
```
isutton: Do we have any guidelines for code comments?

pkg/controller/servicebindingrequest/annotations.go
otaviof: True, that can be introduced again. I changed it back during the process, but now it's not needed anymore.

pkg/controller/servicebindingrequest/controller.go
```diff
@@ -36,70 +34,58 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
...
 
-func addServiceBindingRequestWatches(c controller.Controller) error {
+// add adds a new Controller to mgr with r as the reconcile.Reconciler.
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	opts := controller.Options{Reconciler: r}
+	c, err := controller.New("servicebindingrequest-controller", mgr, opts)
+	if err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
otaviof: Well, `getWatchingGVKs` is already returning a list, and that's all we need in that context. If we need to influence this list, and I believe we do in the near feature, then we can make `getWatchingGVKs` return a different list instead of breaking this logic in the middle of `add` method. Add method would only be reponsible by "adding" it does not need to differentiate which GVK.

pkg/controller/servicebindingrequest/controller.go
otaviof: Simple answer, no.

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -19,6 +21,8 @@ type Reconciler struct {
 	scheme    *runtime.Scheme   // api scheme
 }
 
+type watchObjects map[types.NamespacedName]*unstructured.Unstructured
```
golangcibot: `watchObjects` is unused (from `deadcode`)

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -23,14 +23,15 @@ import (
 type Binder struct {
 	ctx        context.Context                 // request context
 	client     client.Client                   // kubernetes API client
+	objects    []*unstructured.Unstructured    // list of objects employed
```
golangcibot: `objects` is unused (from `structcheck`)

pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -135,18 +140,32 @@ func (r *Retriever) extractConfigMapItemName(xDescriptor string) string {
...
+	gvr := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "secrets"}
+	u, err := r.client.Resource(gvr).Namespace(r.plan.Ns).Get(name, metav1.GetOptions{})
 	if err != nil {
 		return err
 	}
-	logger.Info("Inspecting secret data...")
-	for key, value := range secretObj.Data {
-		logger.WithValues("Secret.Key.Name", key, "Secret.Key.Length", len(value)).
+
+	data, exists, err := unstructured.NestedMap(u.Object, []string{"data"}...)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -155,20 +174,28 @@ func (r *Retriever) readSecret(name string, items []string) error {
...
-	configMapObj := corev1.ConfigMap{}
-	err := r.client.Get(r.ctx, types.NamespacedName{Namespace: r.plan.Ns, Name: name}, &configMapObj)
+
+	gvr := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "configmaps"}
+	u, err := r.client.Resource(gvr).Namespace(r.plan.Ns).Get(name, metav1.GetOptions{})
 	if err != nil {
 		return err
 	}
+
+	data, exists, err := unstructured.NestedMap(u.Object, []string{"data"}...)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -127,21 +141,44 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return RequeueError(err)
 	}
 
+	// storing objects used in Retriever
+	for _, obj := range retrievedObjects {
```
golangcibot: S1011: should replace loop with `objectsToAnnotate = append(objectsToAnnotate, retrievedObjects...)` (from `gosimple`)

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -127,21 +141,44 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 		_ = r.setStatus(ctx, instance, bindingFail)
 		logger.Error(err, "On binding application.")
 		return RequeueOnNotFound(err, requeueAfter)
-	} else if err = r.setApplicationObjects(ctx, instance, updatedObjectNames); err != nil {
+	} else if err = r.setApplicationObjects(ctx, instance, updatedObjects); err != nil {
 		logger.Error(err, "On updating application objects status field.")
 		return RequeueError(err)
 	}
 
+	for _, obj := range updatedObjects {
```
golangcibot: S1011: should replace loop with `objectsToAnnotate = append(objectsToAnnotate, updatedObjects...)` (from `gosimple`)

pkg/controller/servicebindingrequest/mapper.go
```diff
@@ -0,0 +1,24 @@
+package servicebindingrequest
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type Mapper struct{}
```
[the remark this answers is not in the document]
otaviof: Sure, please consider latest changes.

pkg/controller/servicebindingrequest/controller.go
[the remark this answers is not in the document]
otaviof: Please consider latest implementation, considering the discussions we had about the workflow design. We follow a stream of events, changes in Kubernetes, and transform those to become SBRs on the fly, based in annotations or the actual object. So there's a lot less discovery of types, and just straight forward event-driven design.

Issue #154, opened 2019-08-26
Proposal: Add Post-Retrieve Hooks


akashshinde:
# Add Post-Retrieve Hooks

Prepared by: Akash Shinde(akshinde@redhat.com), Avni Sharma(avsharma@redhat.com)

## Abstract

This proposal aims to add support for custom hooks which would get executed before storing bindable variables in Secret.

## Motivation
There are use cases where we would need to update/process bindable variable and then store back to secrets.
One such a usecase we are going through is adding support of custom env variables.

### Architecture
![](https://i.imgur.com/xZmVAO3.png)

### Implementation

Current Retrieve() call
```
func (r *Retriever) Retrieve() error {
    
    ....
    
    Run []hooks
    
    return r.saveDataOnSecret()
}
```

Hook will accept the Retriever instance.
```
struct Hook interface {
    // This would run before storing secrets.
    Run(*Retriever) error
}
```

Then we can implement this hook to add support custom env variables.

##### CustomEnvHook

```
struct CustomEnvHook {
    Hook
}

func (c *CustomEnvHook) Run(r *Retriever) error {
 
    // TODO: Read SBR and figure out all custom env variables to bind
    // Add logic to prepare [key-value] of custom variables. Update r.Data
    // Then r.Data gets persisted.

}
```

## Pros
1. Keeps `Retrieve` modulerized-easy to use. 
2. Easy to test and maintain functional components.
3. Create re-usable, isolated components to avoid redundant logic.

## Cons
1. Implemented in a synchronized manner which might experience some delay.

akashshinde:
@sbose78 basically you would get `retriever` instance to each and every hook which can perform any operation such as 
1. Altering environment variables before storing it to secret.
2. Since you get access to `retriever` you will have kube-connected rest client as well, so basically you can make any k8s resource.

akashshinde:
`retriever` is being shared by all the hooks, so here we need to make sure one `Run()` doesn't disturb the `retriever`. It just should do it's work and write data without tampering the existing one
and then `retriever` gets passed down to next `Hook` and so on.

otaviof:
Interesting design with post-retrieve hooks, good work folks! I wonder if instead of trying to employ post-retrieve hooks straight away, if we could tackle this situation by introducing a new component that Retrieve could interact with, and furthermore, a component that would be passed along in the reconcile workflow. 

Therefore, if we talk about the basic elements of this new component, I would suggest:

``` go
type EnvVar {
[...]

PR #155, opened 2019-08-26
Define a Collector interface to be consumed by Retriever

sbose78:
For https://github.com/redhat-developer/service-binding-operator/issues/154

### Why
`Retriever` today does a lot of things. 
- figures out which configmap keys and secret keys to read
- figures out how to construct environment environment variables and volume paths
- creates the secret(s).

I would like to separate out (1) figuring out which cms and secrets ,  (2) creation of resources.

### Issues
While it is an excellent design to start with, the moment we have different ways to collect the "what to bind", it doesn't scale.
1. reading from a CSV and use env var prefix
2. reading from a CR json paths and use custom env variables.
3. In future, reading from CRD OpenAPI Validation spec

### What
Introducing the `Collector` Interface which makes implementations implement a `Collect()` , it is hard to scale the design when there are multiple ways to find out what is to be bound:

In this PR, I've demonstrated how to implement the existing collector with the new interface.
- CSVCollector - handles the logic to parse the CSV and get env vars out of it.
- CRCollector - handles the logic to parse random CR jsonpaths and get env vars

```
type CSVCollector struct {

}

// Collect returns all bindable metadata by finding the same from CSVs
func (c *CSVCollector) Collect() (*BindableMetadata, error) {

    // business logic to collect all bindable metadata

}
```

All collectors get called in Retrieve(..)

```
// Retrieve loop and read data pointed by the references in plan instance.
func (r *Retriever) Retrieve() error {

	for _, c := range r.collectors {
		c.Collect()
		// c.Collect() returns BindableMetadata, use it to populate data structures in Retriever.
	}

	return r.saveDataOnSecret()
}
```

### How to add a new `Collector`

1. Create an implementation of the `Collector` interface, ie, implement a new structure which has a `Collect()` method.
2. Modify Reconcile(..) to register a new collector

```
retriever.RegisterCollector(collectors.NewCSVCollector(ctx, r.client, plan, instance.Spec.EnvVarPrefix))
```

And that's it. No code changes needed in retriever.go!

### DIFF
--- a/pkg/controller/servicebindingrequest/collectors/collector.go
+++ b/pkg/controller/servicebindingrequest/collectors/collector.go
@@ -1,0 +1,14 @@
+package collectors
+
+type Collector interface {
+	Collect() (BindableMetadata, error)
+}
+
+type BindableMetadata struct {
+	BindableSecrets
+	BindableVolumePaths
+}
+
+type BindableSecrets map[string][]byte
+
+type BindableVolumePaths []string
--- a/pkg/controller/servicebindingrequest/collectors/cr.go
+++ b/pkg/controller/servicebindingrequest/collectors/cr.go
@@ -1,0 +1,1 @@
+package collectors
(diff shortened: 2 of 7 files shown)


sbose78:

PR #156, opened 2019-08-27
Fix a broken link to an image file

ldimaggi:


sbose78:
@ldimaggi There's a lint failure.

PR #157, opened 2019-08-28
Upgrade controller runtime version

baijum:


baijum:
> @baijum did we need to vendor anything?

I ran `go mod vendor` command. But no changes appear in the vendor directory.
Maybe it was already updated?

PR #158, opened 2019-08-28
Extend automated E2E tests to permute order of Backing service, Application and Service Binding Request

pmacik:
This PR:
* Covers: https://jira.coreos.com/browse/APPSVC-126
* [x] Extracts DB, App and SBR creation into individual functions representing the steps to be permutated
* [x] Implements the permutation capability - the test should verify that the binding is done correctly 
 as soon as the Binding Service, Application and Service Binding Request are in place no matter in which order the 3 pieces were created.
* [x] Executes each permutation a clean (empty) namespace (either in a new one created before each permutation is executed, or the current test namespace that has been cleaned from all resources before each permutation is executed.)


test/e2e/servicebindingrequest_test.go
```diff
@@ -183,3 +191,45 @@ func serviceBindingRequestTest(t *testing.T, ctx *framework.TestCtx, f *framewor
 	_ = f.Client.Delete(todoCtx, &sbrSecret)
 	_ = f.Client.Delete(todoCtx, &d)
 }
+
+func CreateDB(todoCtx context.Context, t *testing.T, ctx *framework.TestCtx, f *framework.Framework, ns string, resourceRef string, secretName string) pgv1alpha1.Database {
```
otaviof: Would be nice to leave documentation comments on those methods as well, so we can assess what they are doing and use that description on composing test cases.

otaviof:
> We could have multiple files for specific scenarios like these. What do you think @pmacik ?

I also like the idea, folks. Would be nice to split the functions needed for the test-case scenarios from the scenarios themselves. Would also be a good opportunity to have a good initial documentation block with what the test-scenario aims to do.

pmacik:
> We could have multiple files for specific scenarios like these. What do you think @pmacik ?

Yes, that would make the tests much more readable and maintainable. Thanks for that suggestion, I will see to it.

test/e2e/servicebindingrequest_test.go
pmacik: Good point @otaviof, thanks for that - fixed.

test/e2e/servicebindingrequest_test.go
```diff
@@ -59,25 +59,14 @@ func TestAddSchemesToFramework(t *testing.T) {
 	require.Nil(t, framework.AddToFrameworkScheme(pgsqlapis.AddToScheme, &dbList))
 
 	t.Run("end-to-end", func(t *testing.T) {
+		// scenario-1
 		t.Run("scenario-db-app-sbr", func(t *testing.T) {
 			ServiceBindingRequest(t, []Step{DBStep, AppStep, SBRStep})
 		})
-		t.Run("scenario-app-db-sbr", func(t *testing.T) {
```
pmacik: Let's implement just the 2 existing scenarios here and add the other permutations when this is merged.

PR #159, opened 2019-08-29
Feat: Add custom env support

akashshinde:
**To test** use following SBR

```
kind: ServiceBindingRequest
metadata:
  name: binding-request
spec:
  applicationSelector:
    matchLabels:
      connects-to: postgres
      environment: demo
    group: apps.openshift.io
    version: v1
    resource: deploymentconfigs
  backingServiceSelector:
    group: postgresql.baiju.dev
    version: v1alpha1
    kind: Database
    resourceRef: db-demo
  mountPathPrefix: “”
  envVar:
    - name: JDBC_CONNECTION_STRING
      value: '{{ .spec.imageName }}@{{ .spec.image }}_{{ index .status.dbConfigMap "db.username" }}'
```

**To validate** 
1. Check if intermediate secret has `JDBC_CONNECTION_STRING` key.
2. Ssh into pod and check `JDBC_CONNECTION_STRING` env variable.


pkg/controller/servicebindingrequest/custom_env_path.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func NewCustomEnvPath(envMap []v1alpha1.EnvMap, cache map[string]interface{}) *CustomEnvPath {
+	return &CustomEnvPath{
+		EnvMap: envMap,
+		Cache:  cache,
+	}
+}
+
+func (c *CustomEnvPath) Parse() (map[string][]byte,error) {
```
otaviof: Since we tend to use `map[string]interface{}` in a number of places, should we give it a specific `type`?

pkg/controller/servicebindingrequest/custom_env_path_test.go
```diff
@@ -0,0 +1,43 @@
...
+			Name:  "JDBC_CONNECTION_STRING",
+			Value: `{{ .spec.dbName }}:{{ .status.creds.user }}@{{ .status.creds.pass }}`,
+		},
+		v1alpha1.EnvMap{
+			Name:  "ANOTHER_STRING",
+			Value: `{{ .spec.dbName }}_{{ .status.creds.user }}`,
+		},
+	}
+	customEnvPath := NewCustomEnvPath(envMap, cache)
+	values, err := customEnvPath.Parse()
```
otaviof: You can just replace this by `assert.NoErro(t, err)`

pkg/controller/servicebindingrequest/custom_env_path_test.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
```
otaviof: You can just replace this by `assert.NoError(t, err)`

ron1:
This feature seems to couple binding secret generation with target application binding secret injection. Consider treating binding secret generation and binding secret injection as separate concerns so that volume mounted secrets can also benefit from the binding secret generation portions of this feature.

ron1:
Regarding secret injection as environment variables, I would think the mainstream use case to target leverages envFrom secretRef. Here, the secret has already been generated in a form that is ready for consumption by the application.

pkg/controller/servicebindingrequest/retriever_test.go
```diff
@@ -193,11 +195,71 @@ func TestRetrieverWithConfigMap(t *testing.T) {
...
+		CR: cr,
+	}
+
+	fakeDynClient := f.FakeDynClient()
+
+	retriever = NewRetriever(context.TODO(), fakeDynClient, plan, "SERVICE_BINDING")
+	require.NotNil(t, retriever)
+
+	t.Run("Should detect custom env values", func(t *testing.T) {
+		retriever.Retrieve()
```
golangcibot: Error return value of `retriever.Retrieve` is not checked (from `errcheck`)

pkg/controller/servicebindingrequest/custom_env_path.go
sbose78: Yeah, good idea.

akashshinde: Done https://github.com/redhat-developer/service-binding-operator/pull/159/commits/39aa0717f996e734b697c629687656a36e74bacc

sbose78:
Tested, works !

image

pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -134,9 +155,29 @@ func (r *Retriever) extractConfigMapItemName(xDescriptor string) string {
...
+}
+
 // readSecret based in secret name and list of items, read a secret from the same namespace informed
 // in plan instance.
-func (r *Retriever) readSecret(name string, items []string) error {
+func (r *Retriever) readSecret(
+	name string,
+	items []string,
+	fromPath string,
+	path string) error {
```
otaviof: @akashshinde, @Avni-Sharma, I see in many places formatting errors, like in `error` here. Could you double check if `go format` is working as expected, as in, every time a file is saved?

akashshinde: @otaviof I don't see any error with formatting and `make lint` passed too

sbose78:
Since `envVar` conflicts with `envVar` in the podspec, could we rename it to `customEnvVar`?

deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml
```diff
@@ -82,6 +94,7 @@ spec:
           required:
           - backingServiceSelector
           - applicationSelector
+          - envVar
```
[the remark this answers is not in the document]
sbose78: Right.

pkg/controller/servicebindingrequest/planner.go
```diff
@@ -30,6 +30,7 @@ type Plan struct {
 	Name           string                      // plan name, same than ServiceBindingRequest
 	CRDDescription *olmv1alpha1.CRDDescription // custom resource definition description
 	CR             *unstructured.Unstructured  // custom resource object
+	SBR v1alpha1.ServiceBindingRequest	// service binding request
```
sbose78: Let's not use (possibly vague :) ) abbreviations in the struct, call it `ServiceBindingRequest`.

pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -27,6 +27,7 @@ type Retriever struct {
 	plan          *Plan                        // plan instance
 	volumeKeys    []string                     // list of keys found
 	bindingPrefix string                       // prefix for variable names
+	Cache         map[string]interface{}	// store visited paths
```
sbose78: Can this be `cache` instead of `Cache`?

sbose78:
@otaviof @akashshinde @Avni-Sharma What are the major open concerns here? :) If nothing, let's get this merged by Friday.

pkg/controller/servicebindingrequest/planner.go
otaviof: @sbose78 we have `sbr` as a abbreviation for it in many places, btw :-) So, I would not find strange to find SBR along the code.

otaviof:
> @otaviof @akashshinde @Avni-Sharma What are the major open concerns here? :) If nothing, let's get this merged by Friday.

I think it's worth tackling `retriever.go` and splitting up the concerns that are piling up in that component. Would make a lot more sense to have "Retriever" only fetching the data, and other component to organize the environment variables and store in a secret, before "Binder".

Most of the code for this change is already in place, and now we would need to apply further organization to have a more coherent set of components.

pkg/controller/servicebindingrequest/planner.go
sbose78: > we have sbr as a abbreviation

Agreed :) I'm fine with having the usage in function scoped variables. The `Plan` struct is a very well exposed structure - and wish to have cleaner naming.

sbose78:
> I think it's worth tackling retriever.go and splitting up the concerns 

👍

pkg/controller/servicebindingrequest/retriever.go
akashshinde: Done https://github.com/redhat-developer/service-binding-operator/pull/159/commits/601169cda0eefd1de21c503f001e38d1f104d626

pkg/apis/apps/v1alpha1/zz_generated.openapi.go
```diff
@@ -184,12 +211,24 @@ func schema_pkg_apis_apps_v1alpha1_ServiceBindingRequestSpec(ref common.Referenc
...
+									SchemaProps: spec.SchemaProps{
+										Ref: ref("./pkg/apis/apps/v1alpha1.EnvMap"),
+									},
+								},
+							},
+						},
+					},
 				},
-				Required: []string{"backingServiceSelector", "applicationSelector"},
+				Required: []string{"backingServiceSelector", "applicationSelector", "envVar"},
```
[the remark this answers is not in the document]
akashshinde: Done https://github.com/redhat-developer/service-binding-operator/pull/159/commits/78bd8c8d11bb79fe8d60eff40f0bc7db836cfc30

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -30,6 +30,16 @@ type ServiceBindingRequestSpec struct {
 	// TriggerRebinding forcefully triggers a binding operation if the event
 	// for the same was missed for some reason
 	TriggerRebinding *bool `json:"triggerRebinding,omitempty"`
+
```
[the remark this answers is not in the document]
akashshinde: Done https://github.com/redhat-developer/service-binding-operator/pull/159/commits/78bd8c8d11bb79fe8d60eff40f0bc7db836cfc30

deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml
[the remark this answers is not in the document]
akashshinde: Done https://github.com/redhat-developer/service-binding-operator/pull/159/commits/78bd8c8d11bb79fe8d60eff40f0bc7db836cfc30

deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml
```diff
@@ -78,6 +94,7 @@ spec:
           required:
           - backingServiceSelector
           - applicationSelector
+          - customEnvVar
```
ron1: Is there a reason why customEnvVar is declared as required? In cases where a volume mount is used exclusively, there would be no need to define custom environment variables, right?

deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml
```diff
@@ -69,6 +69,22 @@ spec:
               - kind
               - resourceRef
               type: object
+            customEnvVar:
```
ron1: Would much rather see the ```SecretTransforms``` structure from ServiceBinding/AppBinding (https://github.com/kmodules/custom-resources/blob/114abf10dfe25d845bf8a96c8e5820ee2427aab4/apis/appcatalog/v1alpha1/appbinding_types.go#L47) used here rather than this new ```CustomEnvVar``` structure whose name seems to restrict its use for environment variables only and not volume mounts.

deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml
ron1: Is ```customEnvVar``` intended to be required? In cases where volume mounts are used exclusively, would there be any need for ```customEnvVar```?

sbose78: Nope, this should not be `required`.

deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml
sbose78: Right, I'll let this PR move in for now - and then have the custom volume mounts addressed separately. We'll make a breaking API change if needed. Your point is valid ( CustomEnvVar is too specific ..) . https://github.com/redhat-developer/service-binding-operator/issues/172

@ron1 , In our case, `SecretTransforms` is not entirely valid because I'm trying to see if the binding can be done merely by the user expressing what she needs by looking at the backing service CR , even if there's nothing in the OLM descriptor. ( we have a long term goal to make this friendly for non-OLM users as well, not sure if we'll accomplish that though ) :)

ron1: OLM Annotations + Secret Transforms is a well-understood tech combo that could work today and solve most known use cases. However, it does require buy-in from the OLM backing service operator community and exposes JSON Path which is not entirely user-friendly.

OTOH, I find the current use of both OLM Annotations AND the ```customEnvVar``` struct confusing. In particular, it is not clear that ```customEnvVar``` is both an alternative to OLM Annotations for locating interesting keys to be included in the binding secret and a mechanism to transform binding secret values. It is also not clear whether the transform language references the names of OLM annotation fields or paths to backing service CR fields.

deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml
akashshinde: Done https://github.com/redhat-developer/service-binding-operator/pull/159/commits/723e0b74a21e75a8208b65cbb544c12cbfcac88a

sbose78:
Taking this in. We can address any refactoring in a different PR. 
/lgtm

sbose78:
( and I will get to all the comments shortly :) )

deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml
sbose78: > In particular, it is not clear that customEnvVar is both an alternative to OLM

It intends to be, but it isn't yet :)

PR #164, opened 2019-09-03
Read objects incl. DC/D before annotating

sbose78:
Annotating/updating the DC looks unreliable because right after the binding is done, the `generation`
 changes frequently for a few seconds and hence any update on the object could potentially be an attempt to update a stale  object.

```
2019-09-03T11:03:43.142+0530    ERROR   servicebindingrequest   unable to set/update annotations in object      {"SBR.Namespace": "demo-akash", "SBR.Name": "binding-request", "Resource.GVK": "apps.openshift.io/v1, Kind=DeploymentConfig", "Resource.Namespace": "demo-akash", "Resource.Na
me": "nodejs-rest-http-crud", "error": "Operation cannot be fulfilled on deploymentconfigs.apps.openshift.io \"nodejs-rest-http-crud\": the object has been modified; please apply your changes to the latest version and try again"}
```

In this PR, I have proposed that we `get` the object right before `update`ing the annotation. This only reduces the probability of having a stale object - doesn't eliminate it.

@akashshinde @otaviof  Giving it a second thought, if the update fails the first time and a few reconciles are triggered because of the error in updating the DC, can we let it be that way - as long as it manages to finally find a stable non-frequently-changing DC to update the annotation into?

### DIFF
--- a/pkg/controller/servicebindingrequest/annotations.go
+++ b/pkg/controller/servicebindingrequest/annotations.go
@@ -84,15 +84,8 @@ func SetSBRAnnotations(
 	namespacedName types.NamespacedName,
 	objs []*unstructured.Unstructured,
 ) error {
-	for _, obj := range objs {
-		annotations := obj.GetAnnotations()
-		if annotations == nil {
-			annotations = make(map[string]string)
-		}
 
-		annotations[sbrNamespaceAnnotation] = namespacedName.Namespace
-		annotations[sbrNameAnnotation] = namespacedName.Name
-		obj.SetAnnotations(annotations)
+	for _, obj := range objs {
 
 		gvk := obj.GroupVersionKind()
 		gvr, _ := meta.UnsafeGuessKindToResource(gvk)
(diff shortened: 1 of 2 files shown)


isutton:
> Annotating/updating the DC looks unreliable because right after the binding is done, the `generation`
> changes frequently for a few seconds and hence any update on the object could potentially be an attempt to update a stale object.

Assuming it changes because other actors are touching it, then those operations would generate N events, but only 1 work item will be en-queued. The described behavior happens because you have too many free workers, where each update will trigger an evaluation.

It is worth to mention conflict always happen, but with less frequency in busy systems since those intermediate updates happen while the work item has already been en-queued.

sbose78:
Thanks for your thoughts, @isutton 

> then those operations would generate N events, but only 1 work item will be en-queued

In this case, events are very relevant since the troublesome part is that the DC keeps changing every second for a few seconds till the DC is stable again. Updating that makes it hard since chances are that we could run into a stale DC.

sbose78:
@otaviof 's PR https://github.com/redhat-developer/service-binding-operator/pull/166 is going to handle these issues.

sbose78:

PR #165, opened 2019-09-03
Don't annotate DC/D

sbose78:
As per https://github.com/redhat-developer/service-binding-operator/pull/152#discussion_r318690627 , we decided to not watch DCs/Ds , in that case, how about we don't try annotating Ds/DCs ?

If we do wish to annotate an object which is _in motion_ :) , then a few requeue would occur.  We just need to verify that the requeue would eventually end after a successful annotation on a stable DC/D.


sbose78:

PR #166, opened 2019-09-03
Object Updates as Last Reconciliation Step

otaviof:
Refactoring reconciliation process to only update objects as last step. Status of `ServiceBindingRequest` is kept as a local variable, and updated as the reconciliation process goes, and attempts to update object status will happen in case of error, or at the end of reconciliation.

This PR addresses errors on updating objects that have been modified, and considerably reduces the amount of API calls during reconciliation.

Furthermore, fixing panic triggered on `planner.go:66` due a `nil` variable returned, covering issue scenario with unit-test. The test is asserting the reconciliation processes will return error and cause re-queue in case of expected resources are not found.

### DIFF
--- a/pkg/controller/servicebindingrequest/annotations.go
+++ b/pkg/controller/servicebindingrequest/annotations.go
@@ -1,8 +1,6 @@
 package servicebindingrequest
 
 import (
-	"context"
-
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -79,7 +77,6 @@ func IsSBRNamespacedNameEmpty(namespacedName types.NamespacedName) bool {
 // SetSBRAnnotations update existing annotations to include operator's. The annotations added are
 // referring to a existing SBR namespaced name.
 func SetSBRAnnotations(
-	ctx context.Context,
 	client dynamic.Interface,
 	namespacedName types.NamespacedName,
 	objs []*unstructured.Unstructured,
--- a/pkg/controller/servicebindingrequest/binder.go
+++ b/pkg/controller/servicebindingrequest/binder.go
@@ -254,34 +254,34 @@ func (b *Binder) update(objs *unstructured.UnstructuredList) ([]*unstructured.Un
 		logger := b.logger.WithValues("Obj.Name", name, "Obj.Kind", obj.GetKind())
 		logger.Info("Inspecting object...")
 
-		modifiedObj, err := b.updateSpecContainers(logger, &obj)
+		updatedObj, err := b.updateSpecContainers(logger, &obj)
 		if err != nil {
 			return nil, err
 		}
 
 		if len(b.volumeKeys) > 0 {
-			modifiedObj, err = b.updateSpecVolumes(logger, &obj)
+			updatedObj, err = b.updateSpecVolumes(logger, &obj)
 			if err != nil {
 				return nil, err
 			}
 		}
 
 		logger.Info("Updating object...")
-		if err := b.client.Update(b.ctx, modifiedObj); err != nil {
+		if err := b.client.Update(b.ctx, updatedObj); err != nil {
 			return nil, err
 		}
 
 		logger.Info("Reading back updated object...")
 		// reading object back again, to comply with possible modifications
 		namespacedName := types.NamespacedName{
-			Namespace: modifiedObj.GetNamespace(),
-			Name:      modifiedObj.GetName(),
+			Namespace: updatedObj.GetNamespace(),
+			Name:      updatedObj.GetName(),
 		}
-		if err = b.client.Get(b.ctx, namespacedName, modifiedObj); err != nil {
+		if err = b.client.Get(b.ctx, namespacedName, updatedObj); err != nil {
 			return nil, err
 		}
 
-		updatedObjs = append(updatedObjs, modifiedObj)
+		updatedObjs = append(updatedObjs, updatedObj)
 	}
 
 	return updatedObjs, nil
--- a/pkg/controller/servicebindingrequest/binder.go
+++ b/pkg/controller/servicebindingrequest/binder.go
@@ -3,16 +3,17 @@ package servicebindingrequest
 import (
 	"context"
 	"fmt"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"time"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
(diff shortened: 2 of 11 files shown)


pkg/controller/servicebindingrequest/annotations.go
```diff
@@ -102,11 +96,28 @@ func SetSBRAnnotations(
...
 			"SBR.Name", namespacedName.Name,
 			"Resource.GVK", gvk,
-			"Resource.Namespace", obj.GetNamespace(),
-			"Resource.Name", obj.GetName(),
+			"Resource.Namespace", ns,
+			"Resource.Name", name,
 		)
+
+		logger.Info("Reading resource annotations before updating object...")
+		resourceClient := client.Resource(gvr).Namespace(ns)
```
sbose78: nice!

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -33,57 +34,89 @@ const (
...
-func (r *Reconciler) setStatus(
-	ctx context.Context,
-	instance *v1alpha1.ServiceBindingRequest,
-	status string,
-) error {
-	instance.Status.BindingStatus = status
-	return r.client.Status().Update(ctx, instance)
+// setSecretName update the CR status field to "in progress", and setting secret name.
+func (r *Reconciler) setSecretName(namespacedName types.NamespacedName, name string) error {
+	sbr, err := r.getServiceBindingRequest(namespacedName)
```
sbose78: I have a concern here. If the SBR has changed after the reconcile was triggered, in that case, it is fair to have the `update` fail, and let the next re-queue take care if it. This is specific to the SBR since our controller manages that and we know that nothing on the controller side would be messing with the object.

sbose78: Looks like we should have the latest object on updating the status because the returned value is copied over into the `instance` reference by design, right?
https://github.com/redhat-developer/service-binding-operator/blob/master//pkg/controller/servicebindingrequest/reconciler.go#L49:22

What do you think?

otaviof: @sbose78, I think we should be updating the `status`in SBR only once, and only at the end of the workflow, that would give a more clear approach in terms of code, and avoid this type of implementation all together. :-)

otaviof: > Looks like we should have the latest object on updating the status because the returned value is copied over into the `instance` reference by design, right?
> https://github.com/redhat-developer/service-binding-operator/blob/master//pkg/controller/servicebindingrequest/reconciler.go#L49:22

I don't see a difference in approach here. I moved the part that fetches the actual instance of an SBR to a method, and reused that method in different places.
Before, the value was copied over to `instance` using the same approach, since the `Reconciler` will only received a namespaced-named of an SBR, nothing more than that.

sbose78: >  I think we should be updating the statusin SBR only once, and only at the end of the workflow

+1 I'm good with that!

sbose78:
Otavio, if you could confirm that you've seen attempted stale updates on SBR, I'm good with this change.

otaviof:
> Otavio, if you could confirm that you've seen attempted stale updates on SBR, I'm good with this change.

Please consider latest interactions on this PR, it's fairly different than first attempt, but in line with our more recent conversations.

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -265,11 +266,21 @@ func (b *Binder) update(objs *unstructured.UnstructuredList) ([]*unstructured.Un
 			}
 		}
 
-		logger.Info("Updating object in Kube...")
+		logger.Info("Updating object...")
 		if err := b.client.Update(b.ctx, updatedObj); err != nil {
 			return nil, err
 		}
 
+		logger.Info("Reading back updated object...")
```
sbose78: In line 270, `updatedObj` should have the updated object, right?

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -34,43 +35,88 @@ const (
...
 
 // setSecretName update the CR status field to "in progress", and setting secret name.
-func (r *Reconciler) setSecretName(
-	ctx context.Context,
-	instance *v1alpha1.ServiceBindingRequest,
-	name string,
-) error {
-	instance.Status.BindingStatus = bindingInProgress
-	instance.Status.Secret = name
-	return r.client.Status().Update(ctx, instance)
```
sbose78: good!

pkg/controller/servicebindingrequest/binder.go
otaviof: @sbose78 using the dynamic client the `update` method returns the updated object back, but in this case, it does not, therefore, reading back the changes from cache, a couple lines later.

sbose78: OK, so this is an over-cautious read, fair enough.

otaviof: Actually, that's a necessary action to avoid this error during annotations update:

```
2019-09-06T12:30:23.051+0200	ERROR	servicebindingrequest	unable to set/update annotations in object	{\"SBR.Namespace\": \"test-namespace-b4c393b8-8ce5-480b-b877-753c4f714b87\", \"SBR.Name\": \"e2e-service-binding-request\", \"Resource.GVK\": \"apps/v1, Kind=Deployment\", \"Resource.Namespace\": \"test-namespace-b4c393b8-8ce5-480b-b877-753c4f714b87\", \"Resource.Name\": \"e2e-application\", \"error\": \"Operation cannot be fulfilled on deployments.apps \\"e2e-application\\": the object has been modified; please apply your changes to the latest version and try again\"}
github.com/go-logr/zapr.(*zapLogger).Error
	/Users/otaviof/go/src/github.com/redhat-developer/service-binding-operator/vendor/github.com/go-logr/zapr/zapr.go:128
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest.SetSBRAnnotations
	/Users/otaviof/go/src/github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/annotations.go:108
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest.(*Reconciler).Reconcile
	/Users/otaviof/go/src/github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/reconciler.go:207
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/Users/otaviof/go/src/github.com/redhat-developer/service-binding-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:215
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1
	/Users/otaviof/go/src/github.com/redhat-developer/service-binding-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:158
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
	/Users/otaviof/go/src/github.com/redhat-developer/service-binding-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133
k8s.io/apimachinery/pkg/util/wait.JitterUntil
	/Users/otaviof/go/src/github.com/redhat-developer/service-binding-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134
k8s.io/apimachinery/pkg/util/wait.Until
	/Users/otaviof/go/src/github.com/redhat-developer/service-binding-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88
```

otaviof: Using the dynamic client the API is handier to get back the modified object, but in this case we also had to use a sub-sequential `get`. However, this `get` should come from local cache.

PR #134, opened 2019-08-20
Add OPTIONAL triggerRebinding field in the Spec

sbose78:
- If the user updated the CR to have a `triggerRebinding` to true, a reconcile is triggered where the value is set to`false`. 
- The predicate takes care of rejecting the event if the event was generated because `triggerRebinding` was set from `false` to `true`.

In no scenario, does the controller never set it to `true`.

```
apiVersion: apps.openshift.io/v1alpha1
kind: ServiceBindingRequest
metadata:
  name: binding-request
  namespace: sbose
spec:
  applicationSelector:
    matchLabels:
      connects-to: postgres
      environment: demo
    group: apps.openshift.io
    version: v1
    resource: deploymentconfigs
  triggerRebinding: true
  backingServiceSelector:
    group: postgresql.baiju.dev
    kind: Database
    version: v1alpha1
    resourceRef: demo-database
```

This would come handy when custom resources not owned by the operator are watched.

PR #162, opened 2019-08-31
Adding Watches when a CSV is Created/Updated

otaviof:
Refactoring controller to accommodate watch over `ClusterServiceVersion` (`CSV`) objects in order to inspect and create controller's watch over described `GroupVersionKind` references.

Fixes: #142

### How to Simulate

To simulate this PR, you would have to create a new `CSV` object in the same namespace the operator is working at.

First, spin up operator locally, wait for bootstrap to be completed:

```
$ make local
```

In another shell clone [postgresql-operator](https://github.com/operator-backing-service-samples/postgresql-operator.git), and temporarily edit `manifests/stable/postgresql-operator.v0.0.6.clusterserviceversion.yaml` to use the same namespace than operator.

```
$ sed 's/namespace: placeholder/namespace: YOUR_NAMESPACE/g' manifests/stable/postgresql-operator.v0.0.6.clusterserviceversion.yaml > /var/tmp/example-csv.yaml
$ kubectl apply -f /var/tmp/example-csv.yaml --validate=false
```

On applying `postgresql-operator.v0.0.6`, you should see the operator print the following log line:

```
{"level":"info","ts":1567266681.516482,"logger":"olm","msg":"Extracting GVK from CRDDescription","CRDDescription.Name":"databases.postgresql.baiju.dev"}
2019-08-31T17:51:21.516+0200	INFO	servicebindingrequest.CSVToWatcherMapper	Adding watch for GVK	{"Obj.NamespacedName": "otaviof/postgresql-operator.v0.0.6", "GVK": "postgresql.baiju.dev/v1alpha1, Kind=Database"}
```

PR #171, opened 2019-09-05
Reverting Trigger Rebinding

otaviof:
Removing logic to update `ServiceBindingRequest` objects during reconciliation. The `spec` of a given resource should always belong to the user, not subject for operator driven changes.

Forceful trigger-rebinding changes were added on PR #134 and reverted here.

### Current Alternative

The operator respects `metadata.generation`, so if generation is incremented, the operator will reconcile that object. Please notice the predicate action employed in `controller.go`:

```go
pred := predicate.Funcs{
	UpdateFunc: func(e event.UpdateEvent) bool {
		// ignore updates to CR status in which case metadata.Generation does not change
		return e.MetaOld.GetGeneration() != e.MetaNew.GetGeneration()
	},
}
```

You can see where the operator takes the decision.


otaviof:
> Curious, why do you think this is an operator driven change? This **is** for the user to trigger an update by setting `triggerBinding : true`

It might have been written for the user, but the operator needs to `update` the object to avoid a infinite loop when this flag was set.
 
The primary reasoning behind this change is to use default tooling in Kubernetes instead of cooking our own.

sbose78:
> but the operator needs to update the object to avoid a infinite loop when this flag was set.

Agreed.

> use default tooling in Kubernetes 

What could a user use to trigger rebinding?

isutton:
Wouldn’t bumping `metadata.generation` suffice?

Another option would be using an annotation if being explicit is a requirement.

sbose78:
:) The API server doesn't let the user bump the generation. I tried - it's just sets it back to the old one.

otaviof:
> :) The API server doesn't let the user bump the generation. I tried - it's just sets it back to the old one.

Indeed, I have just tested this scenario, and updating generation of a CRD is not allowed for the time being. For instance: https://github.com/kubernetes/kubernetes/issues/58778

otaviof:
Therefore, I'm going to pursue the annotations path, sounds like second best option in this story.

sbose78:
> Therefore, I'm going to pursue the annotations path, sounds like second best option in this story.

Sure. Could you describe a bit how that would look ?

otaviof:
> What could a user use to trigger rebinding?

That would come from the user editing a given resource that is under watch by this operator. The use case of "forcing rebinding" also implies the operator is not watching all the resources that it should, and since #162 is fixing this, we should take that route IMHO.

otaviof:
> Sure. Could you describe a bit how that would look ?

We are already managing annotations to identify which rebinding instance a given resource under watch belongs to. The same principle could be applied to add one more flag to trigger a forceful rebinding, and then, since the operator would update annotations at the end of the process anyways, that would be removed as well.

sbose78:
> The same principle could be applied to add one more flag to trigger a forceful rebinding

If I were to add an annotation to a ServiceBindingRequest CR, I don't think that would trigger an event.

otaviof:
> > The same principle could be applied to add one more flag to trigger a forceful rebinding
> 
> If I were to add an annotation to a ServiceBindingRequest CR, I don't think that would trigger an event.

Just tested it, and it does create an event. So, we could manage annotations for this use-case.

However, the best approach here would not need to force any reconciliation, but to watch resources that would trigger this instead. :-)

sbose78:
> Just tested it, and it does create an event.

You mean, adding an annotation triggered a reconcile?

otaviof:
> > Just tested it, and it does create an event.
> 
> You mean, adding an annotation triggered a reconcile?

Yep, it does trigger a reconcile. However, intentionally I added a `return true` on the predicate, since I knew the `metadata.generation` would still be the same, despite of different annotations.

sbose78:
> Yep, it does trigger a reconcile. 

great.

> However, intentionally I added a return true on the predicate

right, else the event wouldn't trickle in. 

OK, I'm convinced there's the annotations path in future if we need to! 🍾

PR #174, opened 2019-09-18
Add java app sample

ldimaggi:
This sample will also demonstrate custom env variables.


pmacik:
This depends on https://github.com/redhat-developer/service-binding-operator/pull/159

otaviof:
Could we use something shorter? The directory name is `bind_imported_java_app_to_incluster_operator_managed_PostgreSQL_db`, where we could be using something like `java_postgresql`.

sbose78:
> Could we use something shorter? The directory name is bind_imported_java_app_to_incluster_operator_managed_PostgreSQL_db, where we could be using something like java_postgresql

@ldimaggi ,
Or maybe just remove the "bind_imported_" sub string since it is obvious and repetitive.

pmacik:
> Could we use something shorter? The directory name is `bind_imported_java_app_to_incluster_operator_managed_PostgreSQL_db`, where we could be using something like `java_postgresql`.

Yes, but let's do it in another PR ;)

PR #177, opened 2019-09-19
Comb examples

pmacik:
This PR:
* Renames examples directories to shorter names.
* Removes optional `mountPathPrefix` from service binding requests, where not necessary.


sbose78:
> Removes optional mountPathPrefix from service binding requests, where not necessary.

Hope you've validated? :)

PR #179, opened 2019-09-20
Skip adding annotations to the Deployment

sbose78:
I've noticed infinite reconciles & panics because addition of the annotation on the DC/D keeps failing. This PR safely removes addition of the annotation. 

Since we do not make use of the annotation ( we don't watch DCs/Ds anyway ), not adding the annotation has no adverse impact.

### DIFF
--- a/pkg/controller/servicebindingrequest/reconciler.go
+++ b/pkg/controller/servicebindingrequest/reconciler.go
@@ -226,15 +226,6 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	// storing objects used in Binder
 	objectsToAnnotate = append(objectsToAnnotate, updatedObjects...)
 
-	//
-	// Annotating objects related to binding
-	//
-
-	if err = SetSBRAnnotations(r.dynClient, request.NamespacedName, objectsToAnnotate); err != nil {
-		logger.Error(err, "On setting annotations in related objects.")
-		return r.onError(err, sbr, &sbrStatus, updatedObjects)
-	}
-
 	// updating status of request instance
 	if err = r.updateStatusServiceBindingRequest(sbr, &sbrStatus); err != nil {
 		logger.Error(err, "On updating status of ServiceBindingRequest.")
--- a/pkg/controller/servicebindingrequest/reconciler.go
+++ b/pkg/controller/servicebindingrequest/reconciler.go
@@ -224,6 +224,15 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	// saving on status the list of objects that have been touched
 	r.setApplicationObjects(&sbrStatus, updatedObjects)
 
+	//
+	// Annotating objects related to binding
+	//
+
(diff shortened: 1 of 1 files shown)


otaviof:
@sbose78 I think a better course of action is to fix the panics when trying to update `Deployment` or `DeploymentConfigs`. Since by removing the update of annotations we fundamentally change how this operator works, and that should be based in user configuration instead.

sbose78:
> better course of action is to fix the panics 

Definitely.

> we fundamentally change how this operator works


Umm.. sorry, didn't get you. What exactly gets impacted?

otaviof:
> Umm.. sorry, didn't get you. What exactly gets impacted?

I mean, by removing the annotations from `DC`/`D` resources we lose the ability to trigger reconciliation when they change. The annotation in the object is the clue that makes the operator be able to recognize from which `ServiceBindingRequest` object it belongs to, and not being able to re-reconcile in this scenario is a fundamental change in how the operator works.

So my suggestion was based on, if we are going to change how the operator works, we would need to add this behavior as a configuration.

sbose78:
> by removing the annotations from DC/D resources we lose the ability to trigger reconciliation when they change

Here https://github.com/redhat-developer/service-binding-operator/pull/152#discussion_r318690627
we had dropped watching DCs/Ds.

Given we don't watch / 'manage' DCs/Ds, does not adding annotations to Ds/DCs have any impact?

otaviof:
> Here [#152 (comment)](https://github.com/redhat-developer/service-binding-operator/pull/152#discussion_r318690627)
> we had dropped watching DCs/Ds.
> 
> Given we don't watch / 'manage' DCs/Ds, does not adding annotations to Ds/DCs have any impact?

Good point. However, in #152 we've dropped static watch on those. However, if we can specify `Deployment`/`DeploymentConfigs` using `OLM`/`CSV`, then `ServiceBindingOperator` will still watch over those GVKs. Please consider [`watch.go`](https://github.com/redhat-developer/service-binding-operator/blob/24863e3f648fae10443eee01612dfff2f4c6e4da/pkg/controller/servicebindingrequest/watch.go#L15-L42).

sbose78:
Adding a watch should be fairly simple given the good structuring of the code base. In fact, if we do watch DCs/Ds, it must go into the static list https://github.com/redhat-developer/service-binding-operator/blob/cebeaf828075e9a3abfd5b4f6e1d5913ec8bf9b4/pkg/controller/servicebindingrequest/sbrcontroller.go#L67 here. 

My point is: today we don't watch it :) And I don't think we plan to because we don't wish to manage/respond-to changes to that. In that case, annotations on Ds/DCs don't matter.

otaviof:
@sbose78 okay, I'm willing to see it working without annotating the `DC`/`D` resources. I have a few concerns about it, but we can tackle them in a new PR, if needed at all.

baijum:
@sbose78 @otaviof Any further thought on this PR? Can we merge this PR?

sbose78:
I need an lgtm , this can be merged :)

Issue #112, opened 2019-08-12
Feedback: Handle Backing Services whose CRs link to Secrets/ConfigMaps using CR name-based conventions


ron1:
Backing services such as the ElasticSearch elastic-cloud operator (https://github.com/operator-framework/community-operators/tree/master/upstream-community-operators/elastic-cloud) have a CR that links to Secrets/ConfigMaps using the CR name as a common prefix. As described here: https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-accessing-elastic-services.html#k8s-request-elasticsearch-endpoint, in order to bind to the service endpoint for the Elasticsearch CR, the following secrets must be accessed: 

> $NAME-es-http-certs-public - to retrieve the CA Certificate with key "tls.crt"
> $NAME-es-elastic-user - to retrieve the password of the elastic user with key "elastic"

These secrets are identified via the common CR name prefix "$NAME" followed by well-known suffixes for specific secrets. This is a common pattern used by other operators as well. In the case of the strimzi-kafka-operator KafkaUser CR (https://github.com/operator-framework/community-operators/tree/master/upstream-community-operators/strimzi-kafka-operator) (https://strimzi.io/docs/master/#proc-creating-kafka-user-tls-using-uo), the same name is used for the KafkaUser credentials Secret as the KafkaUser CR itself.

The service-binding-operator should support backing services whose CRs link to Secrets/ConfigMaps using CR name-based conventions. Backing service operator maintainers should not be required to modify their CRDs to support the service-binding-operator. Rather, they should only need to add annotations to the ClusterServiceVersion for their backing service operator to be accessible by the service-binding-operator.

sbose78:
> Rather, they should only need to add annotations to the ClusterServiceVersion for their backing service operator to be accessible by the service-binding-operator.

Agreed, that's my stand as well :)


> The service-binding-operator should support backing services whose CRs link to Secrets/ConfigMaps using CR name-based conventions. 

@ron1  , Are you suggesting that if there's a configmap/secret that is _owned_ by the CR and uses the same naming convention, we consume that as 'interesting to bind' as well - irrespective of whether annotations were added for such objects in the csv?

ron1:
How would interesting configmaps/secrets be identified for annotation in the CVS? No path exists from the CR since the link is via naming convention.

sbose78:
I really like the idea - let me think over it and come up with an end to end picture of this.

ron1:
[...]

PR #180, opened 2019-09-20
Initial draft of operators' best practices guide

ldimaggi:
This doc is WIP until doc review/feedback and dependent stories are completed:
* https://jira.coreos.com/browse/APPSVC-157

### DIFF
--- a/docs/OperatorBestPractices.md
+++ b/docs/OperatorBestPractices.md
@@ -1,0 +1,121 @@
+# Operator Best Practices
+
+## Introduction
+
+The goals of the Service Binding Operator is to make it easier for
+applications developers to bind applications with needed backing
+services, without having to perform manual configuration of secrets,
+configmaps, etc. and to assist operator providers in promoting and
+expanding the adoption of their operators.
+
+When a ServiceBindingRequest is created the Service Binding Operator
+collects binding information and shares it with application. The
+Binding Service Operator's controller injects the binding information
+into the application's "DeploymentConfig", "Deployment" or “Replicaset"
+as environment variables via an intermediate Secret called "binding-request".
+The binding also works with Knative services as it works with any "deployment"
+controller which has the podspec defined in the its jsonpath as
+"spec.template.spec.containers".
+
+This document provides "best practices" guidelines for the development of
+Operators that manage backing services to be bound together with applications
+by the Service Binding Operator.
+
+## Making an Operator Managed Backing Service Bindable
+
+In order to make a service bindable, the operator provider needs to express
+the information needed by applications to bind with the services provided by
+the operator. In other words, the operator provider must express the
+information that is “interesting” to applications.
+
+There are three methods for making Operator Managed Backing Service Bindable:
+
+* [Operator Providing Metadata in CRD Annotations](#operator-providing-metadata-in-crd-annotations)
+* [Operator Providing Metadata in OLM](#operator-providing-metadata-in-olm)
+* [Operator Not Providing Metadata](#operator-not-providing-metadata) Coming Soon!
+
+### Operator Providing Metadata in CRD Annotations
+
+This feature enables operator providers who do not use OLM (Operator Lifecycle
+Manager) to provide metadata outside of an OLM descriptor. In this method,
+the binding information is provided as annotations in the CRD of the operator
+that manages the backing service. The Service Binding Operator extracts the
+annotations to bind the application together with the backing service.
+
+For example, this is a *bind-able* operator's annotations in its CRD for a
+PostgreSQL database backing operator.
+``` yaml
+---
+[...]
+kind: CustomResourceDefinition
+apiVersion: apiextensions.k8s.io/v1beta1
+metadata:
+  name: databases.postgresql.baiju.dev
+  annotations:
+    servicebindingoperator.redhat.io/status.dbConfigMap.password: 'binding:env:object:secret'
+    servicebindingoperator.redhat.io/status.dbConfigMap.username: 'binding:env:object:configmap'
+    servicebindingoperator.redhat.io/status.dbName: 'binding:env:attribute'
+    servicebindingoperator.redhat.io/spec.Token.private: 'binding:volumemount:secret'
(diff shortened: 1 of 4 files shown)


baijum:
@ldimaggi For docs, I would suggest using 80 characters per line. Ref: https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#general-file-guidelines

sbose78:
@baijum 

As soon as you've validated your approach code-wise ( in an incomplete PR/branch ) - I would be eagerly waiting for a 🍏 green signal in the next couple of days :)

Please share an update here so that we can do the following dependent tasks:

- update the Best Practices guide.
- update our sample operators to have annotations instead of descriptors. (we may keep both)
- update our blog post draft to not talk about OLM descriptors anymore.

sbose78:
@ldimaggi https://jira.coreos.com/browse/APPSVC-115 shouldn't be a dependency.

ldimaggi:
> @ldimaggi For docs, I would suggest using 80 characters per line. Ref: https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#general-file-guidelines

Thanks Baiju!

docs/OperatorBestPractices.md
```diff
@@ -0,0 +1,103 @@
...
+sub-resources defined in the backing service CR by populating the binding
+secret with information from Routes, Services, ConfigMaps, and Secrets owned
+by the backing service CR. The binding is initated by the introduction of
+this API option in the backing service CR:
+``` yaml
+detectBindingResources : true
+```
+When this API option is set to true, the Service Binding Operator 
+automatically detects Routes, Services, ConfigMaps, and Secrets owned by
+the backing service CR.
```
ron1: How are potential naming collisions across these various backing service resources handled? How does the customEnvVar syntax distinguish between fields in each of the various backing service resources? Supporting three different binding methods as described in this document seems very confusing to me. I agree there is a need to support Operators that provide no metadata. Although it is a bit more complex and verbose, I think the proposal presented here: https://github.com/redhat-developer/service-binding-operator/issues/112#issuecomment-529271134 is more comprehensive and coherent than what is currently being described in this best practices guide.

sbose78: > Supporting three different binding methods as described in this document seems confusing and disjointed.

- CSV metadata will go away ( make it a bit OLM agnostic )
- CRD annotations will replace CSV metadata.
- CustomEnvVar is not fully baked now.

So, yes, there would be two ways to do things and I agree it is confusing. 

It needs more work and definition. I think https://github.com/redhat-developer/service-binding-operator/issues/112#issuecomment-529271134 is a alternative way to do `customEnvVar`. Since we are still alpha APIs, we will (un)happily :) break the API to remove `customEnvVar` after evaluating the alternatives - I can assure you that!

docs/OperatorBestPractices.md
```diff
@@ -0,0 +1,103 @@
...
+  description: Database connection IP address
+    displayName: DB IP address
+    path: dbConnectionIP
+    x-descriptors:
+      - binding:env:attribute
+```
+
+The operator is available [here.](https://github.com/operator-backing-service-samples/postgresql-operator)
+
+### Operator Providing Metadata not in OLM
```
sbose78: This section already looks good! Mark this as something like "Coming soon!" and I am happy to take this in.

docs/OperatorBestPractices.md
```diff
@@ -0,0 +1,103 @@
...
+    servicebindingoperator.redhat.io/status.dbConfigMap.username: 'env:configmap'
+    servicebindingoperator.redhat.io/status.dbName: env
+```
+
+### Operator Not Providing Metadata
+
+This feature enables operators that manage backing services but which don't
+have any metadata in their CSV to use the Service Binding Operator to bind
+together the service and applications. The Service Binding Operater binds all
+sub-resources defined in the backing service CR by populating the binding
```
sbose78: This is how a resource and sub-resource relationship is set. Let's mention that:
https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents

docs/OperatorBestPractices.md
```diff
@@ -0,0 +1,103 @@
...
+apiVersion: apiextensions.k8s.io/v1beta1
+metadata:
+  name: databases.postgresql.baiju.dev
+  annotations:
+    servicebindingoperator.redhat.io/status.dbConfigMap.password: 'volume:configmap'
+    servicebindingoperator.redhat.io/status.dbConfigMap.username: 'env:configmap'
+    servicebindingoperator.redhat.io/status.dbName: env
+```
+
+### Operator Not Providing Metadata
```
sbose78: Since this is work-in-progress by @akashshinde,  Indicate that this is a "Coming soon!"

ldimaggi: Perfect - Done!

docs/OperatorBestPractices.md
ldimaggi: Nice reference - done!

docs/OperatorBestPractices.md
ldimaggi: Done!

ldimaggi:
> You might want to also show off the reference operators as part of the Best Practices. Can we have a step by step flow described.
> 
> "Here are the 3 things you need to do.."

Add references to the reference operators and the usage scenario examples - please review - thx!

docs/OperatorBestPractices.md
ron1: OK. The annotations seem to include "env" or "volume" in their declaration. Don't you think the client developer rather than the CRD-developer should decide how these values are consumed? Inclusion of "env" or "volume" in the CRD couples extraction/selection and injection which seems undesirable.

ldimaggi:
@sbose78 Do we want to remove, or mark as deprecated, the OLM descriptor based binding approach? Please have a look at the updated doc - thx!

PR #181, opened 2019-09-23
Add knative example.

pmacik:
This PR:
* Adds an example of binding together a [knative application](https://github.com/sbose78/using-spring-data-jpa-quarkus/pull/2) and a [PostgreSQL DB](https://github.com/operator-backing-service-samples/postgresql-operator).
* Introduces a PoC for common scripts usable by any example (such as installing the SBO to OS).


examples/knative_postgresql_customvar/README.md
```diff
@@ -0,0 +1,252 @@
...
+---
+apiVersion: apps.openshift.io/v1alpha1
+kind: ServiceBindingRequest
+metadata:
+  name: binding-request
+  namespace: service-binding-demo
+spec:
+  applicationSelector:
+    matchLabels:
+      connects-to: postgres
```
sbose78: Maybe let's use a resourceRef in this sample?

otaviof:
@pmacik, @ldimaggi, Reading the `Makefile` changes I would recommend having a `hack` script instead. I see that a number of commands are needed and maintaining such lengthy list in `Makefile` is quite out of hand, so I suggest we start to make use of the `hack` scripts and move things over there as they start to make less sense in the `Makefile`.

sbose78:
@pmacik  let's link to https://docs.openshift.com/container-platform/4.1/serverless/installing-openshift-serverless.html as well.

examples/knative_postgresql_customvar/service-binding-request.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: ServiceBindingRequest
+metadata:
+  name: binding-request
+  namespace: service-binding-demo
+spec:
+  applicationSelector:
+    matchLabels:
+      connects-to: postgres
+      environment: demo
+    group: serving.knative.dev
```
sbose78: perfect! :)

examples/knative_postgresql_customvar/Makefile
```diff
@@ -0,0 +1,180 @@
...
+## Create the Backing Service Database
+create-backing-db-instance:
+	@. ../../hack/examples-commons.sh && \
+	install_postgresql_db_instance
+
+#PHONY: set-labels-on-knative-app
+## Set the binding labels for the app
+#set-labels-on-knative-app:
+#	@oc project service-binding-demo
+#	@oc patch dc knative-app -p '{"metadata": {"labels": {"connects-to": "postgres", "environment": "demo"}}}'
```
ldimaggi: Would we also want to create a hack script for this target? The script could take the full DeploymentConfig string as input. Or - maybe would get more clever and find the DC based on the "knative-app" substring in the DC name string?

pmacik:
> Only one question - see separate comment. The hack scripts look great! we can use this a model for all current/future examples!

@ldimaggi Yes, that is the point of this PoC ;)

examples/knative_postgresql_customvar/README.md
pmacik: @sbose78 Yes, good idea ;)

examples/knative_postgresql_customvar/Makefile
pmacik: This target is specific for the example's application and could be (in the worst case) different for each example. So there's no need for the "hack script function" for this and the target should remain in the example's `Makefile`.

In the case of this (knative) example the `ServiceBindingRequest`'s `applicationSelector` uses the `resourceRef` field rather then `matchLabels` so set-labels-on-app target is unnecessary

examples/knative_postgresql_customvar/README.md
pmacik: @sbose78  This is tricky, assuming I want to use `resourceRef` instead of the `matchLabels`.

See: https://github.com/redhat-developer/service-binding-operator/issues/188

pmacik:
Affected by https://github.com/redhat-developer/service-binding-operator/issues/187 and https://github.com/redhat-developer/service-binding-operator/issues/188

pmacik:
> @pmacik let's link to https://docs.openshift.com/container-platform/4.1/serverless/installing-openshift-serverless.html as well.

@sbose78 Thanks for the tip - done (https://github.com/redhat-developer/service-binding-operator/pull/181/commits/16c41695923468ecd620a48c37be0cf955151718#diff-6a53c227ba422af429ac90627029a02fR56)

sbose78:
> Affected by #187 and #188

You could unblock yourself by using the label selector and let @DhritiShikhar take care of updating the docs once she opens a PR to fix it :)

examples/knative_postgresql_customvar/README.md
```diff
@@ -0,0 +1,248 @@
...
+applications. We'll use [postgresql-operator](https://github.com/operator-backing-service-samples/postgresql-operator) to
+demonstrate a sample use case.
+
+We can install all by running the following make target:
+
+```shell
+make install-all
+```
+
+or we can install one by one via the following make targets.
```
ldimaggi: Suggestion - change the text to read:

or we can install one by one via the following six individual make targets.

examples/knative_postgresql_customvar/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+This makes the `KnativeServing` custom resource available, that the application developer will use later to deploy the application.
+
+##### Install Serverless UI
+
+```shell
+make install-knative-serving
+```
+
+This enables `Serverless` view in the UI.
```
ldimaggi: Suggestion:

Change the text to read:

This enables `Serverless` view in the UI. Note that installing the Serverless Operator will require up to ten minutes for the command to run.

examples/knative_postgresql_customvar/README.md
```diff
@@ -0,0 +1,248 @@
...
+attributes. The Backing Service Operator may represent a database or other services required by
+applications. We'll use [postgresql-operator](https://github.com/operator-backing-service-samples/postgresql-operator) to
+demonstrate a sample use case.
+
+We can install all by running the following make target:
+
+```shell
+make install-all
+```
+
```
ldimaggi: Suggestion: Add this text here:

Note that "install-all" command will take up to fifteen minutes to run.

ldimaggi:
Seeing some unexpected behavior:

Screenshot from 2019-10-02 17-51-53

Is this a side effect of #187 and #188?

```
NAME                                           READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/db-demo-postgresql             1/1     1            1           2m13s
deployment.apps/knative-app-2r4wz-deployment   1/1     1            1           2m2s
deployment.apps/knative-app-2x58j-deployment   0/1     1            0           102s
deployment.apps/knative-app-55q5j-deployment   0/1     1            0           111s
deployment.apps/knative-app-9ff54-deployment   0/1     1            0           19m
deployment.apps/knative-app-9nnbm-deployment   0/1     1            0           86s
deployment.apps/knative-app-9wkhk-deployment   0/1     1            0           109s
deployment.apps/knative-app-dk2q9-deployment   0/1     1            0           97s
deployment.apps/knative-app-h2np5-deployment   0/1     1            0           24s
deployment.apps/knative-app-l49v7-deployment   0/1     1            0           113s
deployment.apps/knative-app-pz2x2-deployment   0/1     1            0           119s
deployment.apps/knative-app-rmp25-deployment   0/1     1            0           115s
deployment.apps/knative-app-tcdpj-deployment   1/1     1            1           2m1s
deployment.apps/knative-app-v44hr-deployment   0/1     1            0           65s
deployment.apps/knative-app-vfbr5-deployment   0/1     1            0           107s
deployment.apps/knative-app-vvcqx-deployment   0/1     1            0           105s
deployment.apps/knative-app-zh5wh-deployment   0/1     1            0           117s

```

examples/knative_postgresql_customvar/README.md
pmacik: Thx, updated.

examples/knative_postgresql_customvar/README.md
pmacik: Thanks, good point - updated.

examples/knative_postgresql_customvar/README.md
pmacik: Thanks, good point - updated.

pmacik:
> Is this a side effect of #187 and #188?

No - that is unrelated, I actually switched the `applicationSelector` back to use  `matchLabels` until those issues are fixed. So currently the example does not suffer from it.

README.md
```diff
@@ -92,13 +92,10 @@ We'll add more examples in the future. The following section in this README file
...
+[Binding an Imported app with an In-cluster Operator Managed PostgreSQL Database](examples/nodejs_postgresql/README.md)
 
-[Binding an Imported app to an Off-cluster Operator Managed AWS RDS Database](examples/nodejs_awsrds_varprefix/README.md)
+[Binding an Imported app with an Off-cluster Operator Managed AWS RDS Database](examples/nodejs_awsrds_varprefix/README.md)
 
-[Binding an Imported Java Spring Boot app to an In-cluster Operator Managed PostgreSQL Database](examples/java_postgresql_customvar/README.md)
+[Binding an Imported Java Spring Boot app with an In-cluster Operator Managed PostgreSQL Database](examples/java_postgresql_customvar/README.md)
 
-
-## Run Tests
```
sbose78: @pratikjagrut Removed this after merging your PR because I realized that this needs to go into a CONTRIBUTION.md an we lack one :( CC @ldimaggi .

PR #183, opened 2019-09-26
Read metadata from CRD annotations

baijum:
Jira: https://jira.coreos.com/browse/APPSVC-157

For being more upstream friendly, folks have asked for non-OLM ways of letting operator authors define what is interesting for binding. Given that we did not find a good way to use the OpenAPI validation spec for letting operator authors express what is interesting for binding, we could let operator authors express the same information as annotations in the backing service CRD.

This is how the CRD annotations could look like:

```
apiVersion: apiextensions.k8s.io/v1beta1
metadata:
  name: databases.postgresql.baiju.dev
  annotations:
    servicebindingoperator.redhat.io/spec.dbName: 'binding:env:attribute'
    servicebindingoperator.redhat.io/status.dbCredentials-user: 'binding:env:object:secret'
    servicebindingoperator.redhat.io/status.dbCredentials-password: 'binding:env:object:secret'
    servicebindingoperator.redhat.io/status.dbConnectionIP: 'binding:env:attribute'
    servicebindingoperator.redhat.io/status.dbConnectionPort: 'binding:env:attribute'
    servicebindingoperator.redhat.io/status.dbName: 'binding:env:attribute'
    servicebindingoperator.redhat.io/status.dbConfigMap-db.host: 'binding:env:object:configmap'
    servicebindingoperator.redhat.io/status.dbConfigMap-db.port: 'binding:env:object:configmap'
    servicebindingoperator.redhat.io/status.dbConfigMap-db.name: 'binding:env:object:configmap'
    servicebindingoperator.redhat.io/status.dbConfigMap-db.user: 'binding:env:object:configmap'
    servicebindingoperator.redhat.io/status.dbConfigMap-db.password: 'binding:env:object:configmap'
spec:
  group: postgresql.baiju.dev
  version: v1alpha1
  names:
    plural: databases
    singular: database
    shortNames:
      - db
    kind: Database
    listKind: DatabaseList
  scope: Namespaced
  validation:
    openAPIV3Schema:
      properties:
        apiVersion:
          description: >-
            APIVersion defines the versioned schema of this representation of an
            object. Servers should convert recognized schemas to the latest
            internal value, and may reject unrecognized values.
          type: string
        kind:
          description: >-
            Kind is a string value representing the REST resource this object
            represents. Servers may infer this from the endpoint the client
            submits requests to.
          type: string
        metadata:
          type: object
        spec:
          type: object
        status:
          type: object
  subresources:
    status: {}
  versions:
    - name: v1alpha1
      served: true
      storage: true
  conversion:
    strategy: None
status:
  conditions:
    - type: NamesAccepted
      status: 'True'
      lastTransitionTime: '2019-09-16T09:37:26Z'
      reason: NoConflicts
      message: no conflicts found
    - type: Established
      status: 'True'
      lastTransitionTime: null
      reason: InitialNamesAccepted
      message: the initial names have been accepted
  acceptedNames:
    plural: databases
    singular: database
    shortNames:
      - db
    kind: Database
    listKind: DatabaseList
  storedVersions:
    - v1alpha1
```


pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -26,6 +26,7 @@ type Retriever struct {
 	volumeKeys    []string                     // list of keys found
 	bindingPrefix string                       // prefix for variable names
 	cache         map[string]interface{}       // store visited paths
+	Annotation    bool                         // If true use CRD annotation otherwise fallback to CSV
```
DhritiShikhar: why is `Annotation` an exported field?

baijum: Removed. e67ab21318272503bc6a3d356fd9a66282d19eea

pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -141,6 +141,67 @@ func (r *Retriever) read(place, path string, xDescriptors []string) error {
 	return nil
 }
 
+func (r *Retriever) readAnnotation(key, value string) error {
```
otaviof: I see that `readAnnotation` is turning out to be a duplication of `read` method. How could we make this generic enough so we don't need to copy-and-paste logic?

otaviof:
@sbose78, @baijum instead of having those changes in `retriever.go`, have we consider generating on-the-fly artificial OLM/CSV objects using annotations and passing that long for `retriever`?

sbose78:
> generating on-the-fly artificial OLM/CSV objects

We are avoiding OLM as much as possible, we are still watching them as of now, though.

otaviof:
> > generating on-the-fly artificial OLM/CSV objects
> 
> We are avoiding OLM as much as possible, we are still watching them as of now, though.

I meant as an internal construction, we could very well "fake OLM" objects with annotations, and let the rest of the implementation as they are. Right now we are creating a duplication inside `Retriever`, instead of reusing primitives we've already built.

test/mocks/mocks.go
```diff
@@ -492,3 +565,32 @@ func ConvertToUnstructured(cr interface{}) (*unstructured.Unstructured, error) {
 	data, err := runtime.DefaultUnstructuredConverter.ToUnstructured(cr)
 	return &ustrv1.Unstructured{Object: data}, err
 }
+
+// DatabaseCRMockWithAnnotation based on PostgreSQL operator, returning a instantiated object.
+func DatabaseCRMockWithAnnotation(ns, name string) *pgv1alpha1.Database {
```
isutton: Wouldn't be an argument to the existing `mocks.DatabaseCRMock` function a better option instead?

test/mocks/mocks.go
```diff
@@ -492,3 +565,32 @@ func ConvertToUnstructured(cr interface{}) (*unstructured.Unstructured, error) {
...
+			DBName:    "test-db",
+		},
+		Status: pgv1alpha1.DatabaseStatus{
+			DBCredentials: "db-credentials",
+		},
+	}
+}
+
+// UnstructuredDatabaseCRMockWithAnnotation returns a unstructured version of DatabaseCRMockWithAnnotation.
+func UnstructuredDatabaseCRMockWithAnnotation(ns, name string) (*unstructured.Unstructured, error) {
```
isutton: The same remarks of `DatabaseCRMockWithAnnotation` apply here.

test/mocks/mocks.go
```diff
@@ -492,3 +565,32 @@ func ConvertToUnstructured(cr interface{}) (*unstructured.Unstructured, error) {
 	data, err := runtime.DefaultUnstructuredConverter.ToUnstructured(cr)
 	return &ustrv1.Unstructured{Object: data}, err
 }
+
+// DatabaseCRMockWithAnnotation based on PostgreSQL operator, returning a instantiated object.
+func DatabaseCRMockWithAnnotation(ns, name string) *pgv1alpha1.Database {
+	return &pgv1alpha1.Database{
```
isutton: I fail to find the annotation in the defined value; shouldn't it be somewhere in here?

pkg/controller/servicebindingrequest/retriever.go
isutton: From the usage of both, I don't think they should be unified: `read` reads OLM descriptor related data, and `readAnnotation` returns one of the CR's annotations.

Either way, _read_ is not a descriptive for those methods; I'd use _get_ instead, for example _getOLMDescriptors_ and _getAnnotation_.

test/mocks/mocks.go
baijum: Well, this CR mock is not really required. I will remove it.
The annotation should be there in CRD mock.

test/mocks/mocks.go
baijum: Well, this CR mock is not really required. I will remove it.
The annotation should be there in CRD mock.

pkg/controller/servicebindingrequest/planner.go
```diff
@@ -44,15 +45,37 @@ func (p *Planner) searchCR(kind string) (*unstructured.Unstructured, error) {
...
+func (p *Planner) searchCRD() (*unstructured.Unstructured, error) {
+	bss := p.sbr.Spec.BackingServiceSelector
+	gvk := schema.GroupVersionKind{Group: "apiextensions.k8s.io", Version: "v1beta1", Kind: "CustomResourceDefinition"}
+	gvr, _ := meta.UnsafeGuessKindToResource(gvk)
+	opts := metav1.GetOptions{}
+
+	logger := p.logger.WithValues("CR.GVK", gvk.String(), "CR.GVR", gvr.String())
+	logger.Info("Searching for CRD instance...")
+
+	crd, err := p.client.Resource(gvr).Namespace(p.sbr.GetNamespace()).Get(bss.Kind, opts)
```
DhritiShikhar: There is a difference in namespace. You created your mock in "planner" namespace -> https://github.com/redhat-developer/service-binding-operator/pull/183/files#diff-8187ea18445343ad6726c5be04e8b8d5R57

But `p.sbr.GetNamespace()` here gives me a different namespace ("testing").

```
crd, err := p.client.Resource(gvr).Namespace(p.sbr.GetNamespace()).Get(bss.Kind, opts)
```

pkg/controller/servicebindingrequest/planner.go
```diff
@@ -44,15 +45,37 @@ func (p *Planner) searchCR(kind string) (*unstructured.Unstructured, error) {
...
-		logger.Error(err, "during reading CR")
+		logger.Info("during reading CR")
 		return nil, err
 	}
 
 	logger.WithValues("CR.Name", cr.GetName()).Info("Found target CR!")
 	return cr, nil
 }
 
+// searchCRD based on a CustomResourceDefinitionDescription and name, search for the object.
```
DhritiShikhar: This comment needs to be updated. It's the same as the comment for `searchCR()` method

pkg/controller/servicebindingrequest/planner.go
baijum: There is another test which pass the value as "testing". Can you please check that?
`pkg/controller/servicebindingrequest/retriever_test.go` (Line 260)

pkg/controller/servicebindingrequest/planner.go
```diff
@@ -47,19 +49,52 @@ func (p *Planner) searchCR(kind string) (*unstructured.Unstructured, error) {
...
+
+	logger.WithValues("CR.Name", crd.GetName()).Info("Found target CR!")
+	return crd, nil
+}
+
 // Plan by retrieving the necessary resources related to binding a service backend.
 func (p *Planner) Plan() (*Plan, error) {
 	bss := p.sbr.Spec.BackingServiceSelector
 	gvk := schema.GroupVersionKind{Group: bss.Group, Version: bss.Version, Kind: bss.Kind}
+	//logger := p.logger.WithValues("CR.GVK", gvk.String())
```
DhritiShikhar: you might want to remove this line.

sbose78:
If we are still watching CSVs, we should stop doing so and only do a watch on `crds` - however, we need to careful since CRDs are cluster scoped, and we would have change our `permissions` in our operator csv to `clusterPermissions` 

CC @otaviof @isutton

sbose78:
> If we are still watching CSVs, we should stop doing so and only do a watch on crds - however, we need to careful since CRDs are cluster scoped, and we would have change our permissions in our operator csv to clusterPermissions

We can do it in a separate PR, @baijum .

baijum:
This PR has been tested based on the example README here:
https://github.com/redhat-developer/service-binding-operator/tree/master/examples/nodejs_postgresql
The only difference is that I ran the operator using `operator-sdk up local` command.

pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -171,7 +171,7 @@ func (r *Retriever) readAnnotation(key, value string) error {
 	if len(fullpathslice) > 1 {
 		value = value + ":" + fullpathslice[1]
 	}
-	logger.Debug("Value", value)
+	logger.Debug("Value", "%#v", value)
```
otaviof: @baijum you will need to use something like this:

```go
log.WithValues("SectionMap", sectionMap).Debug("<some-interesting-log-message here>")
```
```

Comment (baijum) [reply pkg/controller/servicebindingrequest/retriever.go]: Fixed 01058a52b00a36a82504b26f8094d5d745e8d475

**Comment by isutton** [comment]
@baijum Given we have the manual test procedure encoded in the README file, do you think it is possible to implement E2E tests that exercise the feature before this PR is merged? 

Having it in place allows us to 1) validate whether the whole E2E work-flow makes sense, 2) verify whether the feature successfully passes the E2E tests and 3) more confidently make and propose changes since regressions could be easily spotted if everything else is correct.

**Comment by sbose78** [comment]
> do you think it is possible to implement E2E tests that exercise the feature before this PR is merged?

Let's do that, @baijum .

Given that this PR is quite a significant enhancement/change, if our confidence using a manual validation isn't high, let's take out a couple of days to get e2e tests in - and merge the PR only after we are satisfied with that.

File: test/e2e/annotation_test.go [hunk 10]
```diff
@@ -0,0 +1,83 @@
...
+		t.Fatalf("failed to add custom resource scheme to framework: %v", err)
+	}
+
+	dbList := pgv1alpha1.DatabaseList{}
+	require.Nil(t, framework.AddToFrameworkScheme(pgsqlapis.AddToScheme, &dbList))
+
+	dpList := appsv1.DeploymentList{}
+	require.Nil(t, framework.AddToFrameworkScheme(appsv1.AddToScheme, &dpList))
+
+	ctx := framework.NewTestCtx(t)
```
otaviof: @baijum by using `mocks.NewFake` it should execute necessary boilerplate to run unit-tests.

test/e2e/annotation_test.go
```diff
@@ -0,0 +1,83 @@
...
+	defer ctx.Cleanup()
+
+	err = ctx.InitializeClusterResources(&framework.CleanupOptions{TestContext: ctx, Timeout: cleanupTimeout, RetryInterval: retryInterval})
+	if err != nil {
+		t.Fatalf("failed to initialize cluster resources: %v", err)
+	}
+
+	// get namespace
+	namespace, err := ctx.GetNamespace()
```
otaviof: @baijum it should only be needed in e2e tests, not during unit-tests.

test/e2e/annotation_test.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// get global framework variables
+	f := framework.Global
+
+	db := mocks.DatabaseCRMock(namespace, "psql")
+
+	err = f.Client.Create(context.TODO(), db, &framework.CleanupOptions{TestContext: ctx, Timeout: time.Second * 5, RetryInterval: time.Second * 1})
```
otaviof: @baijum also during unit-tests you would not need to call `Create`, since there's no real API server running in this phase.

baijum:
> > do you think it is possible to implement E2E tests that exercise the feature before this PR is merged?
> 
> Let's do that, @baijum .
> 
> Given that this PR is quite a significant enhancement/change, if our confidence using a manual validation isn't high, let's take out a couple of days to get e2e tests in - and merge the PR only after we are satisfied with that.

I have written the test and it's passing. Please let me if this is sufficient.

test/e2e/annotation_test.go
```diff
@@ -0,0 +1,96 @@
...
+		t.Fatal(err)
+	}
+
+	sbr := mocks.ServiceBindingRequestMock(namespace, name, resourceRef, appResourceRef, matchLabels, false)
+	err = f.Client.Create(context.TODO(), sbr, &framework.CleanupOptions{TestContext: ctx, Timeout: time.Second * 5, RetryInterval: time.Second * 1})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(1 * time.Minute)
```
isutton: I would use poll here instead of sleeping:

```suggestion
	err = wait.Poll(2*time.Second, 30*time.Second, func() (done bool, err error) {
		if err = f.Client.Get(context.TODO(), namespacedName, sbr2); err != nil {
			if errors.IsNotFound(err) {
				return false, nil
			}
			return true, err
		}
		return true, nil
	})
	if err != nil {
		t.Error(err)
	}
```

isutton:
Some questions:

- Is there anything that requires us to test unbinding in the scope of this PR?
- Is the only side-effect after a SBR to be processed for the secret to exist? If there are other side-effects, are those covered in other test cases somewhere else?

sbose78:
> Is there anything that requires us to test unbinding in the scope of this PR?

Good point, but not in the scope of this PR.

> Is the only side-effect after a SBR to be processed for the secret to exist? If there are other side-effects, are those covered in other test cases somewhere else?

The secret has to exist and the podspec has to be updated. ( That's the minimal contract )

baijum:
> > Is the only side-effect after a SBR to be processed for the secret to exist? If there are other side-effects, are those covered in other test cases somewhere else?
> 
> The secret has to exist and the podspec has to be updated. ( That's the minimal contract )

I have added assertion for seret ref. here:
95994d2f63cd4c54f3359b6d2b7ea0ad2461d1dd

test/e2e/annotation_test.go
pmacik: @otaviof This is not a unit test, but e2e, right?

test/e2e/annotation_test.go
pmacik: @otaviof This is not a unit test, but e2e where the real cluster's API server is involved, right?

test/e2e/annotation_test.go
baijum: Done. 1d1f7612f14e26a81d9b1802f41396192245384c

pmacik:
Hi @baijum, I've made some suggestion to your e2e test code: https://github.com/baijum/service-binding-operator/pull/1

baijum:
> Hi @baijum, I've made some suggestion to your e2e test code: [baijum#1](https://github.com/baijum/service-binding-operator/pull/1)

Thanks @pmacik !

Can you please update your PR. I have pushed one more commit.

PR #189, opened 2019-10-03
Update README to add test target

pratikjagrut:
This PR adds the `test` target in README. It is an important target and should have a place in README.


akram:
I think /LooksGoodToReadMe

PR #193, opened 2019-10-04
Adds a version-update target for bumping up a version

Avni-Sharma:
Adds a new file `update-version.sh` for upgrading to a new version under hack/.
Run `./update-version 1.2.3`
Suitable versioning would be done at appropriate places - package.yaml, csv.yaml, Makefile and Dockerfile.registry.


Makefile
```diff
@@ -353,3 +353,30 @@ else
 		-r $(REPO_OWNER)/$(REPO_NAME) \
 		-Z > codecov-upload.log
 endif
+
+## -- Target for bumping up from a version 'n' to a version 'n+1'
+
+.PHONY :version-update
+## Upgrading operator version
+version-update:
```
otaviof: @Avni-Sharma what if we move those lines to a `hack` script?

otaviof:
Good initiative!

Makefile
isutton: This should be set as guideline: if it is more than a check, then the logic should be moved to its own script.

sbose78:
Heya,
I'd like to use this to release the next version, are we good to take this in?

pratikjagrut:
Can we add a makefile target to run this script?
Like: 
```
NEW-VERSION ?= "0.0.20"
.PHONY: update-version
update-version:
	@- chmod +x ./hack/update-version.sh
	@- ./hack/update-version.sh $(NEW-VERSION)
```

Then we can directly run the script like this:
```
make update-version NEW-VERSION=0.0.21
```
WIll be easier and consistent with Makefile approach in managing project.

But might need to make few minor changes in the `paths used in shell script` as the process dir will be the root dir of the project, not the hack dir.

WDYT? @Avni-Sharma @sbose78

Avni-Sharma:
Yeah, that sounds good. I shall add it

Avni-Sharma:
> Can we add a makefile target to run this script?
> Like:
> 
> ```
> NEW-VERSION ?= "0.0.20"
> .PHONY: update-version
> update-version:
> 	@- chmod +x ./hack/update-version.sh
> 	@- ./hack/update-version.sh $(NEW-VERSION)
> ```
> 
> Then we can directly run the script like this:
> 
> ```
> make update-version NEW-VERSION=0.0.21
> ```
> 
> WIll be easier and consistent with Makefile approach in managing project.
> 
> But might need to make few minor changes in the `paths used in shell script` as the process dir will be the root dir of the project, not the hack dir.
> 
> WDYT? @Avni-Sharma @sbose78 @otaviof

Sounds good. Made appropriate changes

hack/update-version.sh
```diff
@@ -0,0 +1,25 @@
...
+MANIFESTS_DIR="manifests"
+OPERATOR_VERSION_UPDATE=$1
+
+if [ "${OPERATOR_VERSION_UPDATE}" != "${OPERATOR_VERSION}" ] ; then
+    NEW_VERSION="$(echo ${OPERATOR_VERSION_UPDATE} | tr '.' '\n' | tail -1)"
+    OLD_VERSION="$(echo ${OPERATOR_VERSION} | tr '.' '\n' | tail -1)"
+    EXPECTED_VERSION="$(echo ${OLD_VERSION} + 1 | bc )"
+    if [ "${NEW_VERSION}" ==  "${EXPECTED_VERSION}" ] ; then
+        sed -i -e 's/0.0.'${OLD_VERSION}'/0.0.'${NEW_VERSION}'/g' Makefile
+        sed -i -e 's/0.0.'${OLD_VERSION}'/0.0.'${NEW_VERSION}'/g' hack/update-version.sh
```
otaviof: Having the script modifying itself is not a good approach. Could we have a routine to retrieve the current version, and later, the script can bump this version and update on the relevant files.

hack/update-version.sh
```diff
@@ -0,0 +1,25 @@
...
+OPERATOR_VERSION="0.0.23"
+MANIFESTS_DIR="manifests"
+OPERATOR_VERSION_UPDATE=$1
+
+if [ "${OPERATOR_VERSION_UPDATE}" != "${OPERATOR_VERSION}" ] ; then
+    NEW_VERSION="$(echo ${OPERATOR_VERSION_UPDATE} | tr '.' '\n' | tail -1)"
+    OLD_VERSION="$(echo ${OPERATOR_VERSION} | tr '.' '\n' | tail -1)"
+    EXPECTED_VERSION="$(echo ${OLD_VERSION} + 1 | bc )"
+    if [ "${NEW_VERSION}" ==  "${EXPECTED_VERSION}" ] ; then
+        sed -i -e 's/0.0.'${OLD_VERSION}'/0.0.'${NEW_VERSION}'/g' Makefile
```
otaviof: You can export this `sed` action to a `function funcion_name() {}`, and call out this function as you need.

hack/update-version.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+OPERATOR_VERSION="0.0.23"
+MANIFESTS_DIR="manifests"
+OPERATOR_VERSION_UPDATE=$1
+
+if [ "${OPERATOR_VERSION_UPDATE}" != "${OPERATOR_VERSION}" ] ; then
```
otaviof: What about inverting the logic, by checking the required conditions first, and later execute the version update steps? For instance:

```bash
if [[ "${A}" ==  "${B}" ]] ; then
    echo error
    exit 1
fi
```

Makefile
```diff
@@ -347,8 +347,8 @@ clean:
 ## -- Targets for uploading code coverage reports to Codecov.io--
 
 .PHONY: upload-codecov-report
-# Uploads the test coverage reports to codecov.io.
-# DO NOT USE LOCALLY: must only be called by OpenShift CI when processing new PR and when a PR is merged!
+## Uploads the test coverage reports to codecov.io.
+## DO NOT USE LOCALLY: must only be called by OpenShift CI when processing new PR and when a PR is merged!
```
otaviof: Why it can't run locally?

hack/update-version.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
otaviof: Could you add a header comment block explaining why this script exists, describe what's the problem it's trying to solve?

Makefile
```diff
@@ -368,3 +368,9 @@ else
 		-r $(REPO_OWNER)/$(REPO_NAME) \
 		-Z > codecov-upload.log
 endif
+
+NEW-VERSION ?= "0.0.23"
+.PHONY: update-version
+update-version:
+	chmod +x ./hack/update-version.sh
+	./hack/update-version.sh $(NEW-VERSION)
```
pratikjagrut: IMHO: Instead of new variable `NEW-VERSION` we might be able to use an existing variable of version `OPERATOR_VERSION`. It will reduce your work of updating one more variable. Can you once try this way?
```suggestion
	./hack/update-version.sh $(OPERATOR_VERSION)
```
And if the operator version is supposed to update by `+1` i.e `n->n+1` then we can directly embed this logic in code so we won't need to pass any command-line args. WDYT?

Makefile
```diff
@@ -368,3 +368,9 @@ else
 		-r $(REPO_OWNER)/$(REPO_NAME) \
 		-Z > codecov-upload.log
 endif
+
```
pratikjagrut: This comment will work as separator in `make help` output. 
```suggestion
## -- Target to update the Operator version
```

Makefile
```diff
@@ -368,3 +368,9 @@ else
...
+NEW-VERSION ?= "0.0.23"
+.PHONY: update-version
+update-version:
```
pratikjagrut: This comment is necessary as it describes the target in `make help` output.
```suggestion
## This target calls script ./hack/update-version.sh and takes new version as arg
update-version:
```

Makefile
Avni-Sharma: It was mentioned already in master. Just added an extra # so that it shows up on `make help`

Makefile
Avni-Sharma: Removing Makefile target because scripts under hack/ can be run directly. and it is more convenient to do the same by directly passing arguments rather than passing via env variables through Makefile.

pratikjagrut:
Hi @Avni-Sharma,
Did you remove the Makefile target to run this shell script?

Avni-Sharma:
> Hi @Avni-Sharma,
> Did you remove the Makefile target to run this shell script?

Yes. I feel hack/scripts can be run independently

baijum:
@Avni-Sharma If the Makefile changes are no more relevant for this PR, please make those changes in a separate PR.

Avni-Sharma:
> @Avni-Sharma If the Makefile changes are no more relevant for this PR, please make those changes in a separate PR.

not related, but when I was running make command those description were not showing up so added an extra #

DhritiShikhar:
@Avni-Sharma Even if the operator version is downgraded, the message I see is that operator version is upgraded:

```
➜  hack git:(version-update) ✗ ./update-version.sh 0.0.0 

 ✔ Operator version upgraded from 1.2.3 to 0.0.0 
```

pratikjagrut:
Hi @Avni-Sharma,
I tried it with different inputs such as `numbers(1.2.3)`, `characters(hi.hello)` and `alpha-numeric inputs(123#@vre)` and it can also revert the operator version(this might come in handy, but not sure). For all this inputs script successfully updates the operator version. So shouldn't be there some sanitization or verification to check the correct operator version format? WDYT?

output

Avni-Sharma:
> Hi @Avni-Sharma,
> I tried it with different inputs such as `numbers(1.2.3)`, `characters(hi.hello)` and `alpha-numeric-special-character inputs(123#@vre)` and it can also downgrade the operator version(this might come in handy, but not sure). For all this inputs script successfully updates the operator version. So shouldn't be there some sanitization or verification to check the correct operator version format? WDYT?
> 
> output

I don't think so. It is a script to be used by someone from the team and not by the end-users. Therefore we moved it from the Makefile to hack/  Ultimately after updating the version, using this script, a PR needs to be raised with all the modifications done by the script. That won't be approved if that is not the version we require.

PR #195, opened 2019-10-07
Unify and split logging into levels to reduce log flooding in production

pmacik:
The operator logs everything to console with the `info` level of verbosity - incl. information that should be logged on the `debug` level. That not only makes the log too verbose and unreadable but generally has a significant impact on performance and footprint.

That's why the logging should be split into various levels (such as Error, Warning, Info, Debug, Trace... ( https://dave.cheney.net/2015/11/05/lets-talk-about-logging )) and only use levels Info and below in the "production".

This PR:
* Resolves https://jira.coreos.com/browse/APPSVC-197 story


sbose78:
Good one!

Makefile
```diff
@@ -30,6 +31,15 @@ ifeq ($(VERBOSE),2)
...
 	X_FLAG = -x
+	ZAP_LEVEL_FLAG = --zap-level 1
+endif
+ifeq ($(VERBOSE),3)
+	Q_FLAG =
+	QUIET_FLAG =
+	S_FLAG =
+	V_FLAG = -v
+	X_FLAG = -x
+	ZAP_LEVEL_FLAG = --zap-level 2
```
otaviof: should we also add `--zap-encoder=console`? It would be helpful to read log messages in a regular console format, instead of JSON strings.

pkg/logging/logging.go
```diff
@@ -0,0 +1,98 @@
+package logging
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+/*
```
otaviof: Should we use `//` instead of `/**/`? I see the `//` as a more common practice to define comment blocks in Golang.

pkg/logging/logging.go
```diff
@@ -0,0 +1,98 @@
...
+
+Based on the go-logr package.
+
+The Log type enables various verbosity levels to be logged based on to the --zap-level argument to the operator.
+
+To create an instance of the logger:
+
+    log := logging.Logger("main")
+
+When --zap-level is not provided, the logging defaults to INFO level where log.Warning() and log.Info() are logged,
```
otaviof: Could you make sure lines have the same width than specified in `.editorconfig`?

pkg/logging/logging.go
```diff
@@ -0,0 +1,98 @@
...
+// See Info for documentation on how key/value pairs work.
+func (l *Log) WithValues(keysAndValues ...interface{}) *Log {
+	lgr := ((*l.logger).WithValues(keysAndValues...))
+	log := &Log{
+		logger: &lgr,
+	}
+	return log
+}
+
+// WithName adds a new element to the logger's name.
```
otaviof: Here as well, the line should be formatted to have the same width than in `.editorconfig` file.

pmacik: Done.

pkg/logging/logging.go
pmacik: Right... done

pkg/controller/servicebindingrequest/planner.go
```diff
@@ -21,16 +24,16 @@ type Planner struct {
...
 }
 
 // Plan outcome, after executing planner.
 type Plan struct {
-	Ns             string                      // namespace name
-	Name           string                      // plan name, same than ServiceBindingRequest
-	CRDDescription *olmv1alpha1.CRDDescription // custom resource definition description
-	CR             *unstructured.Unstructured  // custom resource object
-	SBR v1alpha1.ServiceBindingRequest	// service binding request
+	Ns             string                         // namespace name
```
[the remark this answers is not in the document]
pmacik: My pleasure :)

cmd/manager/main.go
```diff
@@ -63,20 +64,20 @@ func main() {
 	pflag.CommandLine.AddFlagSet(zap.FlagSet())
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.Parse()
-	logf.SetLogger(zap.Logger())
+	logging.SetLogger(zap.Logger())
```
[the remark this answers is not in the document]
pmacik: Done.

Makefile
```diff
@@ -30,6 +31,15 @@ ifeq ($(VERBOSE),2)
 	S_FLAG =
 	V_FLAG = -v
 	X_FLAG = -x
+	ZAP_LEVEL_FLAG = --zap-level 1
+endif
+ifeq ($(VERBOSE),3)
```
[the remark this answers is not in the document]
pmacik: @otaviof What do you mean? I haven't changed the tab-size here...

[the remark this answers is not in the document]
otaviof: nvm, it looks good!

PR #203, opened 2019-10-14
[WIP Increase retryInterval to enable e2e tests to pass

ldimaggi:
A few seconds more are needed to enable deployments to be successful and enable the e2e tests to pass.

### DIFF
--- a/test/e2e/servicebindingrequest_test.go
+++ b/test/e2e/servicebindingrequest_test.go
@@ -26,7 +26,7 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


pratikjagrut:
Still not working for me. Seeing the same error. And I'm not sure how increasing the retry interval will help? Increasing the retry interval will decrease the number of retries as the timeout is the same.

ldimaggi:
Closing this - the change did not solve the issue - still investigating.

PR #220, opened 2019-10-23
Layout manifests for ART/downstream & upstream

sbose78:
ART needs manifests to be present in a specific path:
Example:
https://github.com/openshift/cluster-logging-operator/tree/release-4.3/manifests

Without that, you would get an error
https://github.com/openshift/ocp-build-data/commit/10fecd094c0c7d706b87f93b302e7329d16de261

Therefore, the new directory structure proposed in this PR:
- manifests-upstream --> stuff that goes into https://github.com/operator-framework/community-operators/pull/797/files , should be removed later on, so that we don't have to maintain in two places.
- manifests --> downstream manifests


manifests/4.3.0/service-binding-operator.v4.3.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,135 @@
...
+                    env:
+                      - name: WATCH_NAMESPACE
+                        value: ""
+                      - name: POD_NAME
+                        valueFrom:
+                          fieldRef:
+                            fieldPath: metadata.name
+                      - name: OPERATOR_NAME
+                        value: service-binding-operator
+                    image: quay.io/redhat-developer/app-binding-operator:v0.0.20
```
sbose78: this gets replaced by ART - so doesn't matter what we have here.

manifests/4.3.0/service-binding-operator.v4.3.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,135 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    alm-examples: '[{"apiVersion":"apps.openshift.io/v1alpha1","kind":"ServiceBindingRequest","metadata":{"name":"example-servicebindingrequest"},"spec":{"applicationSelector":{"matchLabels":{"connects-to":"postgres","environment":"production"},"resourceKind":"Deployment"},"backingServiceSelector":{"resourceName":"database.example.org","resourceVersion":"v1alpha1"}}}]' # yamllint disable-line rule:line-length
+    capabilities: Basic Install
+    containerImage: quay.io/redhat-developer/app-binding-operator:v0.0.20
```
sbose78: gets replaced by ART

PR #229, opened 2019-10-30
Update README.md with etcd example link

akashshinde:


Avni-Sharma:
Can we use something like this in our project https://github.com/remarkjs/remark-validate-links

PR #237, opened 2019-11-12
Make the RHEL Dockerfile the primary Dockerfile

sbose78:
- Removed `Dockerfile`
- Renamed `Dockerfile.rhel` to `Dockerfile`
- This branch has triggered a build https://quay.io/repository/redhat-developer/app-binding-operator/build/d1916965-2fdc-4c16-ac45-11dec33f7577 which will generate `quay.io/redhat-developer/app-binding-operator:rhel8-dockerfile`

PR #239, opened 2019-11-12
Removed Old Dockerfile

sbose78:
The only Dockerfile that will be around is the Dockerfile.rhel
- It was tested with ART and some builds might be configured with it. Hence since some places may still refer to that, I will avoid renaming it to `Dockerfile`
- @Avni-Sharma has validated the image build with Dockerfile.rhel in #237 
- Reconfigured the trigger on Quay to use `Dockerfile.rhel` https://quay.io/repository/redhat-developer/app-binding-operator?tab=builds


baijum:
Let's close #237

PR #240, opened 2019-11-12
Remove sdk generated build dir

sbose78:
I don't think we use it anywhere :)


baijum:
This is required for `operator-sdk up local`:
```
$ operator-sdk up local
INFO[0000] Running the operator locally.                
INFO[0000] Using namespace default.                     
FATA[0000] must run command in project root dir: project structure requires build/Dockerfile 
```

sbose78:
Ah OK, then need to add a tiny Readme in that directory.

Avni-Sharma:
https://github.com/redhat-developer/service-binding-operator/pull/277

PR #243, opened 2019-11-14
Add knative service unit tests

pratikjagrut:
This PR adds knative services unit test against binder.Bind(). 
To run all unit tests `make test-unit`
To run only knative test 
`cd pkg/controller/servicebindingrequest`
`go test -run TestKnativeServicesContractWithBinder -v`

It also vendors dep needed for knative service mock.


go.mod
```diff
@@ -1,36 +1,43 @@
...
+	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e // indirect
+	google.golang.org/api v0.10.0 // indirect
+	google.golang.org/appengine v1.6.2 // indirect
+	google.golang.org/genproto v0.0.0-20190911173649-1774047e7e51 // indirect
 	k8s.io/api v0.0.0-20190612125737-db0771252981
 	k8s.io/apiextensions-apiserver v0.0.0-20190228180357-d002e88f6236
 	k8s.io/apimachinery v0.0.0-20190612205821-1799e75a0719
 	k8s.io/client-go v12.0.0+incompatible
-	k8s.io/code-generator v0.0.0-20181203235156-f8cba74510f3
+	k8s.io/code-generator v0.0.0-20191109100332-a9a0d9c0b3aa
```
otaviof: @pratikjagrut `code-generator` is used by `operator-sdk`, would be good to validate `operator-sdk generate` command continue to work.

hack/yamls/knativeServiceCrd.yaml
```diff
@@ -0,0 +1,64 @@
+# Copyright 2019 The Knative Authors
```
otaviof: That's a file that we need for testing, however it's located at `hack` directory. I suggest we always keep test related files under `test` directory.
Furthermore, I suggest a file that contains an third-party CRD also carries the CRD version on the name, or an easy way to identify the version.

pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -28,7 +28,8 @@ func TestBinderNew(t *testing.T) {
 	}
 
 	f := mocks.NewFake(t, ns)
-	sbr := f.AddMockedServiceBindingRequest(name, "ref", "", matchLabels)
+	gvr := "apps/v1/deployments" // Group/Version/Resource for sbr
```
otaviof: @pratikjagrut instead of using a hardcoded string, you can programmatically build it, using the following [example](https://github.com/redhat-developer/service-binding-operator/blob/b9616709b8ec3081ecbfbe8ce70ba2076a049b7a/pkg/controller/servicebindingrequest/olm.go#L36):

```go
gvr := olmv1alpha1.SchemeGroupVersion.WithResource(csvResource)
```

pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -133,7 +133,8 @@ func TestBinderApplicationName(t *testing.T) {
 	name := "service-binding-request"
 
 	f := mocks.NewFake(t, ns)
-	sbr := f.AddMockedServiceBindingRequest(name, "backingServiceResourceRef", "applicationResourceRef", nil)
+	gvr := "apps/v1/deployments" // Group/Version/Resource for sbr
```
otaviof: Same than previous comment, and all other occurrences.

test/e2e/knativeservice_test.go
```diff
@@ -0,0 +1,120 @@
+package e2e
```
otaviof: @pratikjagrut Considering the tests executed against Knative resources, we should move those as unit-tests. Since there's no backing operator for Knative services, we would not see the benefits of writing an end-to-end test for this purpose, and in unit-tests level we are completely able to validate contracts, in the same line than @DhritiShikhar's PR #247.

go.mod
pratikjagrut: it was automatically updated. So I didn't look into it much

test/e2e/knativeservice_test.go
isutton: Perhaps a little bit more clarity is required here.

The reason this test makes sense to be considered a unit test is that it in its current form doesn't exercise much more than those, but with the burden of having to provision a cluster. 

Using the existing libraries made available by the operator-sdk to marshal/unmarshal manifests are the same used in the API server, so exercising marshalling and unmarshalling is enough to guarantee compatibility between different systems; in this particular case would be modifying Knative manifest to enable the binding in such a way it is properly encoded (the value encoded fits a given type) and semantically valid (against OpenAPI validation rules declared in the CRD). 

Producing properly encoded and valid changes in the API server is what we are pursuing at each reconciliation, and for this we need to rely on the assumption the other system will work. I find end-to-end tests really valuable to confirm that all integrated systems behave as expected; please note this is fundamentally different from unit tests, which is to prove the contract is correct.

test/mocks/mocks.go
```diff
@@ -380,7 +382,9 @@ func ServiceBindingRequestMock(
 	applicationResourceRef string,
 	matchLabels map[string]string,
 	bindUnannotated bool,
+	groupVersionResource string,
```
isutton: Please use [schema.GroupVersionResource](https://godoc.org/k8s.io/apimachinery/pkg/runtime/schema#GroupVersionResource) here.

test/e2e/knativeservice_test.go
pratikjagrut: @isutton so this test can be directly moved under unit test as it is?

test/mocks/mocks.go
pratikjagrut: Oh, thanks for this I didn't know!

test/e2e/knativeservice_test.go
isutton: @pratikjagrut I'm not sure about the "as is" part. If you have access from `Reconcile()` to the data you expect to exist then executing `Reconcile()` directly with a client that already has all the manifests you expect to exist in the cluster is as effective as if you're testing against a cluster (assuming the contract is being properly tested, it is).

Given that, configuring a client with supporting manifests (all the manifests that are expected to be consulted during reconciliation) to be tested and exercising the business logic by triggering the reconciliation process is an approach I sympathize with.

pratikjagrut: I pushed the unit test code. Resolving this conversation.

sbose78:
@pratikjagrut resolve conflict please.

pratikjagrut:
> @pratikjagrut resolve conflict please.

@sbose78 Resolved.

go.mod
pratikjagrut: I've updated it accordingly operator-sdk's dep, and ran unit and e2e tests its working fine.

pratikjagrut:
@otaviof and @isutton can you please review new changes?

test/mocks/mocks.go
```diff
@@ -380,6 +382,7 @@ func ServiceBindingRequestMock(
 	applicationResourceRef string,
 	matchLabels map[string]string,
 	bindUnannotated bool,
+	groupVersionResource schema.GroupVersionResource,
```
otaviof: Typically this is called `gvr` in the rest of the project, I suggest the same here.

test/mocks/fake.go
```diff
@@ -34,9 +35,10 @@ func (f *Fake) AddMockedServiceBindingRequest(
 	backingServiceResourceRef string,
 	applicationResourceRef string,
 	matchLabels map[string]string,
+	groupVersionResource schema.GroupVersionResource,
```
otaviof: Here as well, it should be named `gvr` simply.

test/mocks/fake.go
```diff
@@ -59,9 +62,10 @@ func (f *Fake) AddMockedUnstructuredServiceBindingRequest(
...
```
otaviof: And the same for the other occurrences :-)

pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -167,3 +170,43 @@ func TestBinderApplicationName(t *testing.T) {
 	})
 
 }
+
+func TestKnativeServicesContractWithBinder(t *testing.T) {
+	ns := "binder"
```
otaviof: Could you make sure formatting is set to 4 spaces and following `.editorconfig`?

Makefile
```diff
@@ -406,4 +406,3 @@ endif
 consistent-crds-manifests-upstream:
 	$(Q)cd ./manifests-upstream/${OPERATOR_VERSION}/ && ln -srf ../../deploy/crds/apps_v1alpha1_servicebindingrequest_crd.yaml \
 	servicebindingrequests.apps.openshift.io.crd.yaml
-
```
otaviof: Please leave last empty line in place.

DhritiShikhar:
@pratikjagrut Can you change the title of the PR. It still says `e2e` -> `Add knative service e2e test`

sbose78:
@pmacik @isutton In light of latest incidents of failing scenarios, should we have an e2e test for this? Or do you think a unit test is good enough?

isutton:
> @pmacik @isutton In light of latest incidents of failing scenarios, should we have an e2e test for this? Or do you think a unit test is good enough?

Our e2e tests, if you look closely, they only test whether the API server is working. for anything other than Deployments. By adding third party CRDs without installing their infrastructure, e2e is as effective as testing the Reconciler alone with the proper scenario. So I think unit testing for this is enough, assuming the contract works.

pkg/controller/servicebindingrequest/binder_test.go
pratikjagrut: In my editor, it is 4 spaces only. 
spaces

pratikjagrut: At GitHub also it is 4 spaces, I think in reviewing its showing 2
space

Makefile
pratikjagrut: okay!

pkg/controller/servicebindingrequest/binder_test.go
pratikjagrut: Try to view it as a file.

test/mocks/fake.go
pratikjagrut: Well, all this gvr thing is resolved by somebody's PR.

Makefile
pratikjagrut: Done

PR #256, opened 2019-11-21
Unbinding: Handling Resource Deletion

otaviof:
### Motivation

This PR is addressing the deletion of a `ServiceBindingRequest`, it reproduces the steps taken during binding, and undo those actions leveraging Kubernetes Resources Finalizers.

### Changes

Operator components were reorganized to only trigger changes in resources on specific method calls, and refactoring in `Retriever` component took place to split the context of intermediary secret with the component employed to download data.

### Testing

In order to test unbinding, end-to-end tests have been extended to cover deletion of objects, and after fact verification if objects used on binding are back to original. Those changes are in the default testing workflow, therefore are happening in all described Scenarios.


isutton:
I've been following this change closely and even started working on improve upon it, so I'll try to approve it if I have the right permissions :)

/approve

PR #257, opened 2019-11-21
Remove upstream manifests

sbose78:
Since it's already managed in 
https://github.com/operator-framework/community-operators/tree/master/community-operators/service-binding-operator


Avni-Sharma:
openshift-ci/Dockerfile.registry.build copies manifests-upstream to manifests/service-binding-operator in the image. Would that now copy from manifests?

Avni-Sharma:
hack/update-version.sh would be used by manifests now?

sbose78:
Curious, do we use olm manifests in our e2e tests at all today?

sbose78:
Avni, upstream-manifests is effectively moving into the new repo which we created, and would be populated by the automation you are building.

We shouldn't be maintaining any copy of it here anymore. 

Thoughts?

Avni-Sharma:
> Curious, do we use olm manifests in our e2e tests at all today?

As of today, it is commented
https://github.com/openshift/release/blob/30cf2168824fc6ff77dcd60385192955beff3ad2/ci-operator/config/redhat-developer/service-binding-operator/redhat-developer-service-binding-operator-master.yaml#L62

PR #263, opened 2019-11-28
e2e etcd operator test

akashshinde:
### Motivation

This changes add e2e test case by connecting etcd database to application using `detectBindingResources` flag set to true.
### Changes

Added new e2e scenario.
### Testing

Use `make test-e2e` to test.


test/mocks/etcd_operator_mocks.go
```diff
@@ -0,0 +1,61 @@
+package mocks
```
otaviof: @akashshinde good idea to spread mocks in more files! 👍

test/mocks/mocks.go
```diff
@@ -21,6 +21,8 @@ import (
 
 // resource details employed in mocks
 const (
+	// Fixme(Akash): This values are tightly coupled with postgresql operator.
+	// Need to make it more dynamic.
```
otaviof: I think we have those values as constants in `mocks.go`, if not, would be good to organize in the same approach than in `common.go` (under controller).

test/e2e/servicebindingrequest_test.go
```diff
@@ -511,3 +561,53 @@ func serviceBindingRequestTest(
 	_, err = assertDeploymentEnvFrom(todoCtx, f, deploymentNamespacedName, sbrName)
 	require.Error(t, err, "expect deployment not to be carrying envFrom directive")
 }
+
+// CreateServiceBindingRequest implements end-to-end step for creating a Service Binding Request to bind the Backing Service and the Application
+func CreateServiceBindingRequest(
+	todoCtx context.Context,
```
otaviof: When passing the context down the common variable name would be `ctx`, instead of `todoCtx`. Here it goes as an minor suggestion.

test/e2e/servicebindingrequest_test.go
```diff
@@ -127,6 +138,26 @@ func ServiceBindingRequest(t *testing.T, steps []Step) {
...
+	d := &appsv1.Deployment{}
+	if err := f.Client.Get(ctx, namespacedName, d); err != nil {
+		return nil, err
+	}
+
+	containers := d.Spec.Template.Spec.Containers
+	if len(containers) != 1 {
+		return nil, fmt.Errorf("can't find a container in deployment-spec")
+	}
+	//if len(containers[0].EnvFrom) != 1 {
```
otaviof: Should you remove commented lines?

test/e2e/etcd_operator_unannotated_test.go
```diff
@@ -0,0 +1,78 @@
...
+	ctx := framework.NewTestCtx(t)
+	defer ctx.Cleanup()
+
+	ns, f := bootstrapNamespace(t, ctx)
+
+	// executing testing steps on operator
+	serviceBindingRequestTestSetup(t, ctx, f, ns, steps)
+}
+
+func serviceBindingRequestTestSetup(t *testing.T, ctx *framework.TestCtx, f *framework.Framework, ns string, steps []Step) {
```
otaviof: Please leave a documentation comment explaining what this step is doing, and what's expected from it. Also, the line seems to be way over 101 columns limit (`.editorconfig` settings).

test/e2e/etcd_operator_unannotated_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"testing"
+	"time"
+)
+
+func ServiceBindingRequestSetup(t *testing.T, steps []Step) {
```
otaviof: Is the same method we have in the other end-to-end test file? Could we share the same code among those?

test/e2e/servicebindingrequest_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/coreos/etcd-operator/pkg/apis/etcd/v1beta2"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
otaviof: In the other files we've been using `metav1` instead of just `v1`. Please consider.

test/e2e/servicebindingrequest_test.go
```diff
@@ -439,12 +470,14 @@ func serviceBindingRequestTest(
...
 ) {
 	// making sure resource names employed during test are unique
+	name := "e2e-service-binding-request"
 	randomSuffix := rand.Int()
 	csvName := fmt.Sprintf("cluster-service-version-%d", randomSuffix)
 	sbrName := fmt.Sprintf("e2e-service-binding-request-%d", randomSuffix)
 	resourceRef := fmt.Sprintf("e2e-db-testing-%d", randomSuffix)
 	secretName := fmt.Sprintf("e2e-db-credentials-%d", randomSuffix)
 	appName := fmt.Sprintf("e2e-application-%d", randomSuffix)
+	serviceBindingRequestNamespacedName := types.NamespacedName{Namespace: ns, Name: name}
```
otaviof: Consider naming using `sbr`, that's a well known abbreviation. :-)

test/e2e/servicebindingrequest_test.go
```diff
@@ -511,3 +561,53 @@ func serviceBindingRequestTest(
 	_, err = assertDeploymentEnvFrom(todoCtx, f, deploymentNamespacedName, sbrName)
 	require.Error(t, err, "expect deployment not to be carrying envFrom directive")
 }
+
+// CreateServiceBindingRequest implements end-to-end step for creating a Service Binding Request to bind the Backing Service and the Application
+func CreateServiceBindingRequest(
```
otaviof: In the other end-to-end file we also have an `CreateSBR` method, could them share the same?

test/mocks/mocks.go
```diff
@@ -373,8 +375,9 @@ func ServiceBindingRequestMock(
 	applicationResourceRef string,
 	matchLabels map[string]string,
 	bindUnannotated bool,
+	backendGvk *metav1.GroupVersionKind,
```
isutton: Since all the information is coming from the caller anyway, why not leave the responsibility to fix the `BackingServiceSelector` field to the caller itself? For example, a SBR mock could be augmented through composition instead:

```go
// create a new mock
sbr := ServiceBindingRequestMock(...)
// configure the SBR's backing service selector
setBackingServiceSelectorMock(backingServiceResourceRef, backendGVK)
```

Dealing with the differences in the caller site is fine in this case, and then only on required places this function would be called to modify the SBR available, instead of introducing a new parameter and setting `nil` in almost every call site, which makes me like this parameter approach less.

akashshinde: @isutton makes sense

test/e2e/servicebindingrequest_test.go
```diff
@@ -127,6 +154,27 @@ func ServiceBindingRequest(t *testing.T, steps []Step) {
 	serviceBindingRequestTest(t, ctx, f, ns, steps)
 }
 
+func assertAppDeployed(
```
golangcibot: `assertAppDeployed` is unused (from `deadcode`)

test/e2e/servicebindingrequest_test.go
```diff
@@ -99,6 +114,7 @@ func TestAddSchemesToFramework(t *testing.T) {
 
 // cleanupOptions using global variables to create the object.
 func cleanupOptions(ctx *framework.TestCtx) *framework.CleanupOptions {
+	return nil
```
isutton: I assume this is a left over?

test/e2e/servicebindingrequest_test.go
```diff
@@ -506,23 +539,20 @@ func serviceBindingRequestTest(
...
-					Group:   "etcd.database.coreos.com",
-					Version: "v1beta2",
-					Kind:    "EtcdCluster",
-				},
-				true,
-			)
+				func(sbr *v1alpha1.ServiceBindingRequest) {
+				setSBRBackendGVK(sbr, resourceRef, v1beta2.SchemeGroupVersion.WithKind(v1beta2.EtcdClusterResourceKind))
+				setSBRBindUnannotated(sbr, true)
+			})
```
isutton: Indentation here looks funny. Did you `gofmt` it?

test/e2e/servicebindingrequest_test.go
```diff
@@ -152,9 +169,9 @@ func assertAppDeployed(
 	if len(containers) != 1 {
 		return nil, fmt.Errorf("can't find a container in deployment-spec")
 	}
-	//if len(containers[0].EnvFrom) != 1 {
-	//	return nil, fmt.Errorf("can't find envFrom in first container")
-	//}
+	if len(containers[0].EnvFrom) > 0 {
+		return nil, fmt.Errorf("Containers  are less than 0")
```
isutton: Is this message correct?  I believe it doesn't match with the conditional.

test/e2e/servicebindingrequest_test.go
akashshinde: We don't have this function anymore

test/e2e/servicebindingrequest_test.go
akashshinde: Outdated

test/e2e/etcd_operator_unannotated_test.go
akashshinde: Outdated

test/e2e/etcd_operator_unannotated_test.go
akashshinde: Outdated

test/e2e/servicebindingrequest_test.go
akashshinde: Outdated

test/e2e/servicebindingrequest_test.go
akashshinde: Now we are using common `CreateSBR` method. :)

test/mocks/mocks.go
akashshinde: @isutton Introduced dedicated function to modify SBR

test/e2e/servicebindingrequest_test.go
akashshinde: Outdated

test/e2e/servicebindingrequest_test.go
akashshinde: did `gofmt`, indentation is this way because this is an function definition in the paramter

test/e2e/servicebindingrequest_test.go
akashshinde: Fixed

test/e2e/servicebindingrequest_test.go
akashshinde: Outdated

test/e2e/servicebindingrequest_test.go
akashshinde: Fixed

PR #265, opened 2019-12-03
Fix double base64 encoding for customEnvVar (#264)

### LINKED ISSUE #264: Double base64 encoding for customEnvVar retrieved from a secret
I am using the customEnvVar to retrieve some values from a secret that I want to inject into a DeploymentConfig with a specific name for the environment variables, as those are the names used in existing node.js code. My service binding request looks like:

```
kind: ServiceBindingRequest
metadata:
  name: binding-request
  namespace: srb
spec:
  applicationSelector:
    matchLabels:
      connects-to: translator
      environment: demo
    group: apps.openshift.io
    version: v1
    resource: deploymentconfigs
  backingServiceSelector:
    group: ibmcloud.ibm.com
    version: v1alpha1
    kind: Binding
    resourceRef: mytranslator-binding
  customEnvVar:
    - name: LANGUAGE_TRANSLATOR_URL
      value: '{{ index .status.secretName "url" }}'
    - name: LANGUAGE_TRANSLATOR_IAM_APIKEY
      value: '{{ index .status.secretName "apikey" }}'
```

Expected Result: 
`LANGUAGE_TRANSLATOR_URL` and `LANGUAGE_TRANSLATOR_IAM_APIKEY` are present in the `binding-request` secret and the value is base64 encoded, so that I can get the actual value with base64 decode

Actual Result:
`LANGUAGE_TRANSLATOR_URL` and `LANGUAGE_TRANSLATOR_IAM_APIKEY` are present in the `binding-request` secret  but they are base64 encoded twice, so doing base64 decode provides a base64 encoded string which needs to be base64 decoded to get the actual value. This does not work with the existing node.js code.

pdettori:
### Motivation

Fix #264 

### Changes

Changes are in retriever.go to check if variables for customEnvVar are already base 64-encoded and decoding them if they are, since all variables are base 64 encoded when written to the binding request secret.

### Testing

Did not add new tests, just checked that double base 64 encoding does not happen for customEnvVar retrieved from a secret and no changes are made when variables are retrieved directly from status.

### DIFF
--- a/pkg/controller/servicebindingrequest/retriever.go
+++ b/pkg/controller/servicebindingrequest/retriever.go
@@ -199,7 +199,7 @@ func (r *Retriever) readSecret(
 		log.Debug("Inspecting secret key...")
 		r.markVisitedPaths(path, k, fromPath)
 		// update cache after reading configmap/secret in cache
-		r.cache[fromPath].(map[string]interface{})[path].(map[string]interface{})[k] = string(value)
+		r.cache[fromPath].(map[string]interface{})[path].(map[string]interface{})[k] = string(data)
 		// making sure key name has a secret reference
 		r.store(fmt.Sprintf("configMap_%s", k), data)
 		r.store(fmt.Sprintf("secret_%s", k), data)
--- a/pkg/controller/servicebindingrequest/retriever_test.go
+++ b/pkg/controller/servicebindingrequest/retriever_test.go
@@ -243,7 +243,7 @@ func TestCustomEnvParser(t *testing.T) {
 		if err != nil {
 			t.Error(err)
 		}
-		require.Equal(t, "dXNlcg==_cGFzc3dvcmQ=", values["ANOTHER_STRING"], "Custom env values are not matching")
-		require.Equal(t, "postgres@cGFzc3dvcmQ=", values["JDBC_CONNECTION_STRING"], "Custom env values are not matching")
+		require.Equal(t, "user_password", values["ANOTHER_STRING"], "Custom env values are not matching")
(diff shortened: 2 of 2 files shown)


baijum:
A test would be nice.

sbose78:
Thank you very much @pdettori for the fix!

isutton:
I can see values read from Secrets are already being decoded (see below), so I'm not sure what is the error. Can you be more specific in which scenarios you experience the double encoding issue?

https://github.com/redhat-developer/service-binding-operator/blob/a96fa4ea37595f4b4e072d62a928fb36362b408f/pkg/controller/servicebindingrequest/retriever.go#L192-L194

pdettori:
> Can you be more specific in which scenarios you experience the double encoding issue?

@isutton the scenario and the SRB I am using  is described in #264. It happens with customEnvVar referencing values from a secret.

isutton:
I understand that, but I don't think the source of the double encoding has been fixed, only the collateral damage it produced. From what I read in master, all secrets are read by `readSecret`, which decodes the expected base64 string.

Actually, while reading the code I found the issue (see below): it is storing `value` instead of the decoded content stored in `data`.

https://github.com/redhat-developer/service-binding-operator/blob/a96fa4ea37595f4b4e072d62a928fb36362b408f/pkg/controller/servicebindingrequest/retriever.go#L202

Can you check if it works on your case?

pdettori:
@isutton you are correct, after seeing your comment I noticed that the issue is in that line. Will test and update the PR to fix that line if that is OK w/ you.

pdettori:
@isutton it works, updated the PR.

PR #266, opened 2019-12-03
Update SDK version

baijum:
### Motivation

The current version of Operator SDK and dependencies are old.

### Changes

Updated SDK version and dependencies.

### Testing


baijum:
> @baijum we also need to change `operator-sdk` version in `.travis.yml` and `Makefile`. Please consider those. Ideally, this version would come from a single place, that would be a good improvement considering all components should rely in the same exact version.

I have changed `.travis.yml`. I don't see any SDK version in `Makefile`.

PR #273, opened 2019-12-09
Use corev1.EnvVar instead of CustomEnvMap

baijum:
### Motivation

Fixes https://jira.coreos.com/browse/APPSVC-325

### Changes

The change should be backward compatible

### Testing

Existing tests should run


deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
```diff
@@ -0,0 +1,210 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
sbose78: Why are we adding this file?

pkg/apis/apps/v1alpha1/zz_generated.openapi.go
```diff
@@ -216,13 +190,13 @@ func schema_pkg_apis_apps_v1alpha1_ServiceBindingRequestSpec(ref common.Referenc
...
 							Description: "BackingServiceSelector is used to identify the backing service operator.",
-							Ref:         ref("./pkg/apis/apps/v1alpha1.BackingServiceSelector"),
+							Ref:         ref("github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1.BackingServiceSelector"),
 						},
 					},
 					"applicationSelector": {
 						SchemaProps: spec.SchemaProps{
 							Description: "ApplicationSelector is used to identify the application connecting to the backing service operator.",
-							Ref:         ref("./pkg/apis/apps/v1alpha1.ApplicationSelector"),
+							Ref:         ref("github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1.ApplicationSelector"),
```
isutton: I noticed those lines change quite frequently from "github.com/redhat-developer/..." to "./pkg/..." and back. Perhaps this is related with GO111MODULE configuration?

isutton:
> `make build` on an clean tree results in the following:
> 
> ```
> GOARCH=amd64 GOOS=linux go build  -o ./out/operator cmd/manager/main.go
> verifying github.com/operator-framework/operator-sdk@v0.12.0: checksum mismatch
>         downloaded: h1:aD4qPbSAbZgRj1jFdFLq/dBI4P4aKX8d4rJowyQtTYM=
>         go.sum:     h1:9eAD1L8e6pPCpFCAacBUVf2eloDkRuVm29GTCOktLqQ=
> 
> SECURITY ERROR
> This download does NOT match an earlier download recorded in go.sum.
> The bits may have been replaced on the origin server, or an attacker may
> have intercepted the download attempt.
> 
> For more information, see 'go help module-auth'.
> make: *** [Makefile:283: out/operator] Error 1
> ```
> 
> Removing the offending line and executing `make build` update `go.sum` with the correct checksum, so either something is missing from your side or we need to verify the reason we have different checksums.

Please disregard this comment, I found an issue in my environment that didn't make it clear at the time I wrote this, although the inner comment about GO111MODULE and the path being used in the generated file still applies though.

deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
baijum: That was autogenerated. I have removed it.8149e90d0df51eb0cbc93ed3aa58177915826f7e

pkg/apis/apps/v1alpha1/zz_generated.openapi.go
baijum: Which version of operator-sdk is used in your system?
This is my version:
```
$ operator-sdk version
operator-sdk version: "v0.12.0", commit: "2445fcda834ca4b7cf0d6c38fba6317fb219b469", go version: "go1.13.3 linux/amd64"
```

isutton: I've noticed this pattern of changes between several commits, always from or to pkg (see this file's history).

baijum: Anyway, I think both should work.

PR #274, opened 2019-12-09
Test to verify binding a single backing service with two individual applications

DhritiShikhar:
Story -> https://jira.coreos.com/browse/APPSVC-246


pmacik:
@DhritiShikhar Your PR actually adds a unit test, which is completely fine and cool. However, the [story](https://jira.coreos.com/browse/APPSVC-246) is about adding an end-2-end test.

DhritiShikhar:
@pmacik I have updated the Jira story

pmacik:
> @pmacik I have updated the Jira story

@DhritiShikhar Wait, my point was not about the story being wrong, but to let you know that this PR (while it adds a unit test) do not implements the story... The original story was correct - to add an e2e test)

pmacik:
@DhritiShikhar And, (hopefully) to avoid further confusion - This PR is completely valid - it just adds a slightly different thing than the original story.

ldimaggi:
This comment is to late as the PR has been merged - but - an interesting negative test would be to track the error (if any returned) if a user inadvertently bound the same backing service more than once.

Issue #272, opened 2019-12-06
"Error calling index" when using customEnvVar


pdettori:
I am running the service binding operator from master (b0c9641574269f58b6126a406bace6a4b2c07afd on Dec 5) and using `make local` to run. When I try to run the example in `examples/nodejs_postgresql` with the following service binding request:

```
apiVersion: apps.openshift.io/v1alpha1
kind: ServiceBindingRequest
metadata:
  name: binding-request
  namespace: service-binding-demo
spec:
  applicationSelector:
    matchLabels:
      connects-to: postgres
      environment: demo
    group: apps.openshift.io
    version: v1
    resource: deploymentconfigs
  backingServiceSelector:
    group: postgresql.baiju.dev
    version: v1alpha1
    kind: Database
    resourceRef: db-demo
  customEnvVar:
    - name: DB_USER
      value: '{{ index .status.dbConfigMap "db.username" }}'  
```

I get the following error:

```
{"level":"error","ts":1575664614.742795,"logger":"reconciler.bind","msg":"On saving secret data..","Request.Namespace":"service-binding-demo","Request.Name":"binding-request","ServiceBindingRequest.Name":"binding-request","error":"template: set:1:3: executing \"set\" at <index .status.dbConfigMap \"db.username\">: error calling index: index of untyped nil","stacktrace": ...
```

pdettori:
the problem seems to happen in `custom_env_parser.go`, in 

```Golang
// Parse interpolates and caches the templates in EnvMap.
func (c *CustomEnvParser) Parse() (map[string]interface{}, error) {
	data := make(map[string]interface{})
	for _, v := range c.EnvMap {
		tmpl, err := template.New("set").Parse(v.Value)
		if err != nil {
			return data, err
		}

		// evaluating template and storing value in a buffer
		buf := new(bytes.Buffer)
		err = tmpl.Execute(buf, c.Cache)
		if err != nil {
			return data, err
		}

		// saving buffer in cache
		data[v.Name] = buf.String()
	}
	return data, nil
}
```

pdettori:
Seems like the content of the cache is not what expected, printing c.Cache as json provides:
```json
{
  "DATABASE_CONFIGMAP_DB_HOST": "MTcyLjMwLjY3LjE1NA==",
  "DATABASE_CONFIGMAP_DB_NAME": "ZGItZGVtbw==",
[...]

PR #281, opened 2019-12-11
Reuse retrieve.Cache for customEnvParser in secret.go

Avni-Sharma:
Fix for https://github.com/redhat-developer/service-binding-operator/issues/272

### DIFF
--- a/pkg/controller/servicebindingrequest/reconciler.go
+++ b/pkg/controller/servicebindingrequest/reconciler.go
@@ -242,6 +242,7 @@ func (r *Reconciler) bind(
 	binder *Binder,
 	secret *Secret,
 	retrievedData map[string][]byte,
+	retrievedCache map[string]interface{},
 	sbr *v1alpha1.ServiceBindingRequest,
 	sbrStatus *v1alpha1.ServiceBindingRequestStatus,
 	objectsToAnnotate []*unstructured.Unstructured,
@@ -249,7 +250,7 @@ func (r *Reconciler) bind(
 	logger = logger.WithName("bind")
 
 	logger.Info("Saving data on intermediary secret...")
-	secretObj, err := secret.Commit(retrievedData)
+	secretObj, err := secret.Commit(retrievedData, retrievedCache)
 	if err != nil {
 		logger.Error(err, "On saving secret data..")
 		return r.onError(err, sbr, sbrStatus, nil)
@@ -352,7 +353,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	logger.Debug("Retrieving data to create intermediate secret.")
 	retriever := NewRetriever(r.dynClient, plan, sbr.Spec.EnvVarPrefix)
-	retrievedData, err := retriever.Retrieve()
+	retrievedData, retrievedCache, err := retriever.Retrieve()
 	if err != nil {
 		logger.Error(err, "On retrieving binding data.")
 		return r.onError(err, sbr, sbrStatus, nil)
@@ -374,5 +375,5 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 
 	logger.Info("Starting the bind of application(s) with backing service...")
-	return r.bind(logger, binder, secret, retrievedData, sbr, sbrStatus, objectsToAnnotate)
+	return r.bind(logger, binder, secret, retrievedData, retrievedCache, sbr, sbrStatus, objectsToAnnotate)
 }
--- a/pkg/controller/servicebindingrequest/retriever.go
+++ b/pkg/controller/servicebindingrequest/retriever.go
@@ -267,19 +267,19 @@ func (r *Retriever) store(key string, value []byte) {
 // Retrieve loop and read data pointed by the references in plan instance. Also runs through
 // "bindable resources", gathering extra data. It can return error on retrieving and reading
 // resources.
-func (r *Retriever) Retrieve() (map[string][]byte, error) {
+func (r *Retriever) Retrieve() (map[string][]byte, map[string]interface{}, error) {
 	var err error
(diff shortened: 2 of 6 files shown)


pkg/controller/servicebindingrequest/retriever_test.go
```diff
@@ -207,7 +204,7 @@ func TestCustomEnvParser(t *testing.T) {
 	crName := "db-testing"
 
 	f := mocks.NewFake(t, ns)
-	f.AddMockedUnstructuredCSV("csv")
+	// f.AddMockedUnstructuredCSV("csv")
```
isutton: Remove comment.

sbose78:
Avni, could you test this feature end to end and provide the detailed steps here on the PR ?

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -352,7 +353,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	logger.Debug("Retrieving data to create intermediate secret.")
 	retriever := NewRetriever(r.dynClient, plan, sbr.Spec.EnvVarPrefix)
-	retrievedData, err := retriever.Retrieve()
+	retrievedData, retrievedCache, err := retriever.Retrieve()
```
otaviof: In the current design of the operator, `Retriever` should only fetch data specified by the `Plan` (from `Planner`), therefore, `Retreive` should only return one single set of data, not two. The idea is the later components won't have to figure out sources of data anymore, and act in a single view of the binding context data.

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -374,5 +375,5 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 
 	logger.Info("Starting the bind of application(s) with backing service...")
-	return r.bind(logger, binder, secret, retrievedData, sbr, sbrStatus, objectsToAnnotate)
+	return r.bind(logger, binder, secret, retrievedData, retrievedCache, sbr, sbrStatus, objectsToAnnotate)
```
otaviof: Same idea here, make sure `retreievedData` contains it all.

pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -267,19 +267,19 @@ func (r *Retriever) store(key string, value []byte) {
...
 // "bindable resources", gathering extra data. It can return error on retrieving and reading
 // resources.
-func (r *Retriever) Retrieve() (map[string][]byte, error) {
+func (r *Retriever) Retrieve() (map[string][]byte, map[string]interface{}, error) {
 	var err error
 	r.logger.Info("Looking for spec-descriptors in 'spec'...")
 	for _, specDescriptor := range r.plan.CRDDescription.SpecDescriptors {
 		if err = r.read("spec", specDescriptor.Path, specDescriptor.XDescriptors); err != nil {
-			return nil, err
+			return nil, nil, err
```
otaviof: Same here, it should return a single unified context data.

pkg/controller/servicebindingrequest/secret.go
```diff
@@ -20,12 +20,8 @@ type Secret struct {
 
 // customEnvParser parse informed data in order to interpolate with values provided by custom
 // environment component.
-func (s *Secret) customEnvParser(data map[string][]byte) (map[string][]byte, error) {
+func (s *Secret) customEnvParser(data map[string][]byte, cache map[string]interface{}) (map[string][]byte, error) {
```
isutton: This method could be moved as is to `Retriever`and used by `Retrieve()`.

pkg/controller/servicebindingrequest/reconciler.go
Avni-Sharma: > Same idea here, make sure `retreievedData` contains it all.

But retrievedData maintains a key,value pair list whereas for cache we need to retain the whole structure (hierarchy) as mentioned here https://github.com/redhat-developer/service-binding-operator/issues/272#issuecomment-562735404

Avni-Sharma: Having just the `secretData` would not maintain that structure

sbose78:
Looks like the `customEnvVar` feature broke during our unbinding feature work https://github.com/redhat-developer/service-binding-operator/pull/256/files#diff-2c6d61012a9ded0b460e51c66cd47057R25

Since, fixing this the way suggested above is taking time, I would suggest taking in this PR to un-break the feature.

Here's what I would like to do :
1. Track the suggestion as a story.
2. Take this PR in.
3. @Avni-Sharma & @isutton work closely to understand what needs to be done. Take your time and improve the code.

At the same time, our tests need to improve so that we catch these issues.

What do you think?

Avni-Sharma:
Tested locally
Cache hierarchy maintained. Output:
```
{"ignored key": {"spec":{"dbName":"db-demo","image":"docker.io/postgres","imageName":"postgres"},"status":{"db.host":{"db-demo":{}},"db.name":{"db-demo":{}},"db.password":{"db-demo":{}},"db.port":{"db-demo":{}},"db.user":{"db-demo":{}},"dbConfigMap":{"db.host":"172.30.89.99","db.name":"db-demo","db.password":"password","db.port":"5432","db.username":"postgres"},"dbConnectionIP":"172.30.89.99","dbConnectionPort":5432,"dbCredentials":"db-demo-postgresql","dbName":"db-demo","password":{},"user":{}}}}
```

Referred example https://github.com/redhat-developer/service-binding-operator/tree/master/examples/java_postgresql_customvar
Tested using devconsole UI

Avni-Sharma:
```
status:
  applicationObjects:
  - service-binding-demo/java-app
  bindingStatus: Success
  secret: java-app-d-db-demo-postgresql-d
```

PR #282, opened 2019-12-18
Ensure MountPathPrefix is optional

baijum:
Refer https://issues.redhat.com/browse/APPSVC-320


isutton:
Isn't it required to run `make generate-k8s` to update the generated code?

sbose78:
> Isn't it required to run make generate-k8s to update the generated code?

Yes!

sbose78:
@baijum , could we address the feedback and get this in ?

baijum:
I ran `make generate-k8s` and `make generate-openapi`, but it's not making any difference. If anyone can run it locally with the change in this PR, that would be helpful.

baijum:
The struct tag `omitempty` was already making the field as optional. So, this change is redundant. But for consistency, we can merge this PR.

PR #285, opened 2019-12-24
Avoid using / root paths in Dockerfile

sbose78:
### Motivation

Avoid writing to `/` in the Dockerfile

### Changes

- Avoid using GOPATH-ish paths
- Avoid writing to any path which is typically writable to, only by root.
- Collapse the multiple layers for copying source code into the image.

### Testing

https://quay.io/repository/redhat-developer/app-binding-operator/build/5ed422ff-eb4a-404a-bc28-94adffedba47

Build validated above.

### DIFF
--- a/Dockerfile.rhel
+++ b/Dockerfile.rhel
@@ -1,3 +1,5 @@
+ARG BUILD_DIR=/tmp/service-binding-operator
+
 # On CI, use
 # FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
 
@@ -9,7 +11,7 @@ ENV GIT_COMMITTER_EMAIL devtools@redhat.com
 LABEL com.redhat.delivery.appregistry=true
 
 # GOPATH-ish dir structure not needed anymore
-WORKDIR /tmp/service-binding-operator
+WORKDIR $BUILD_DIR
 
 # TODO: Use a RUN command to copy over specific files/dirs in a different PR
 # Collapsing it into a single COPY to reduce layers, as of now.
@@ -30,7 +32,7 @@ LABEL maintainer "Devtools <devtools@redhat.com>"
 LABEL author "Shoubhik Bose <shbose@redhat.com>"
 ENV LANG=en_US.utf8
 
-COPY --from=builder /tmp/service-binding-operator/out/operator /usr/local/bin/service-binding-operator
+COPY --from=builder $BUILD_DIR/out/operator /usr/local/bin/service-binding-operator
 
 USER 10001
 
--- a/Dockerfile.rhel
+++ b/Dockerfile.rhel
@@ -4,6 +4,7 @@ ARG BUILD_DIR=/tmp/service-binding-operator
 # FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
 
 FROM openshift/origin-release:golang-1.13 AS builder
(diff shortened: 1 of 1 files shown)


Dockerfile.rhel
```diff
@@ -8,31 +8,20 @@ ENV GIT_COMMITTER_NAME devtools
...
-COPY pkg pkg
-COPY test test
-COPY vendor vendor
-COPY go.mod .
-COPY go.sum .
-COPY LICENSE .
-COPY Makefile .
-COPY tools.go .
+# GOPATH-ish dir structure not needed anymore
+WORKDIR /tmp/service-binding-operator
```
pedjak: We can avoid repeating the path by declaring an argument before first `FROM` statement and then use it in all stages:

```
ARG BUILD_DIR=/tmp/service-binding-operator
.
.
FROM  openshift/origin-release:golang-1.13 AS builder
.
.
WORKDIR $BUILD_DIR
.
.
COPY --from=builder $BUILD_DIR/out/operator /usr/local/bin/service-binding-operator

Comment (sbose78) [reply Dockerfile.rhel]: Done, thanks!

Comment (sbose78) [reply Dockerfile.rhel]: Predrag, doesn't work 🤔  https://quay.io/repository/redhat-developer/app-binding-operator/build/01c0a07e-9645-4fd2-88ae-244933963252

Comment (pedjak) [reply Dockerfile.rhel]: I am not authorized to view the logs, but I can confirm the issue locally, and was able to fix it. You need to redeclare the arg in each stage, as described in the [doc](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact). Check this diff:

```diff
diff --git a/Dockerfile.rhel b/Dockerfile.rhel
index 33e2bf9..8f458e2 100644
--- a/Dockerfile.rhel
+++ b/Dockerfile.rhel
@@ -5,6 +5,8 @@ ARG BUILD_DIR=/tmp/service-binding-operator
 
 FROM openshift/origin-release:golang-1.13 AS builder
 
+ARG BUILD_DIR
+
 ENV LANG=en_US.utf8
 ENV GIT_COMMITTER_NAME devtools
 ENV GIT_COMMITTER_EMAIL devtools@redhat.com
@@ -27,6 +29,8 @@ RUN make build
 
 FROM registry.access.redhat.com/ubi7/ubi-minimal
 
+ARG BUILD_DIR
+
 LABEL com.redhat.delivery.appregistry=true
 LABEL maintainer "Devtools <devtools@redhat.com>"
 LABEL author "Shoubhik Bose <shbose@redhat.com>"
```

sbose78: Let me try - my local docker is defunct

Dockerfile.rhel
```diff
@@ -1,47 +1,40 @@
...
-COPY pkg pkg
-COPY test test
-COPY vendor vendor
-COPY go.mod .
-COPY go.sum .
-COPY LICENSE .
-COPY Makefile .
-COPY tools.go .
+# GOPATH-ish dir structure not needed anymore
+WORKDIR $BUILD_DIR
```
baijum: Since the path is fixed, how about setting the value as `/tmp/service-binding-operator`

sbose78: does this help?
`ARG BUILD_DIR=/tmp/service-binding-operator`

PR #286, opened 2020-01-03
SBO design direction with multiple services support

isutton:
### Motivation

Some changes are required to implement arbitrary paths to embed secret in Application resources and support multiple backing services. 
### Disclaimer

This PR serves as reference to SBO design direction with multiple services support; it is very likely the work should be re-done from `master` and existing components be used in several PRs paving the way to reach the desired design.

### Changes

This PR proposes a design direction using and modifying existing components, and also add some high level components and types.

* The `Reconcile` function was re-factored, creating an extra component and a builder function; the builder function wraps the data gathering process, encoded by `Planner` and `Retriever`, and the `ServiceBinder` component which is responsible for the orchestration of objects in Kubernetes, using `Binder` and `Secret`.

* `Retriever` received the "detect bindable resources" functionality; previous bindable resources implementation was too late in the process. I believe that we can simplify it further as well, but can't pinpoint exactly which direction as of yet.

* `RelatedResources` field was added to `Plan` to support multiple backing services, deprecating both `CRDDescription` and `CR`. `RelatedResource` is a CRDDescription/CR tuple.

### Testing

End-to-end tests have not been modified in this branch, so it is expected that everything that was working prior this work should be working.

Tests have been added for `ServiceBinder`, and tests covering existing components have been modified accordingly.


pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -52,10 +56,8 @@ type ServiceBindingRequestStatus struct {
 // BackingServiceSelector defines the selector based on resource name, version, and resource kind
 // +k8s:openapi-gen=true
 type BackingServiceSelector struct {
-	Group       string `json:"group"`
-	Version     string `json:"version"`
-	Kind        string `json:"kind"`
-	ResourceRef string `json:"resourceRef"`
+	schema.GroupVersionKind `json:",inline" yaml:",inline"`
```
sbose78: @isutton Could I assign https://issues.redhat.com/browse/APPSVC-324 to you 
CC @baijum ?

baijum: Can we do this change in a separate PR merge it first?

sbose78: Sure if Igor is ok with it

sbose78:
What is the plan with this PR ?

isutton:
This has landed in master.

PR #291, opened 2020-01-14
Use Kubernetes label selector

baijum:
Fixes https://issues.redhat.com/browse/APPSVC-322

### Motivation

Make API familiar to users through adopting Kubernetes native APIs

### Changes

Use `LabelSelector` from `k8s.io/apimachinery/pkg/apis/meta/v1` package

### Testing

The existing tests would be fine. The mock object has been modified.


sbose78:
FYI: It is fine to have breaking changes with respect to labels since UI isn't consuming it?

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -169,9 +169,9 @@ func checkSBR(sbr *v1alpha1.ServiceBindingRequest, log *log.Log) error {
 		log.Debug("Spec.ApplicationSelector.ResourceRef not found")
 
 		// Check if MatchLabels is present
```
DhritiShikhar: The comment should be updated

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -63,11 +63,11 @@ type BackingServiceSelector struct {
 // +k8s:openapi-gen=true
 type ApplicationSelector struct {
 	// +optional
-	MatchLabels map[string]string `json:"matchLabels,omitempty"`
-	Group       string            `json:"group,omitempty"`
-	Version     string            `json:"version"`
-	Resource    string            `json:"resource"`
-	ResourceRef string            `json:"resourceRef"`
+	LabelSelector *metav1.LabelSelector `json:",inline"`
```
DhritiShikhar: Why an extra comma in the beginning `,inline`?

baijum: The tag name for the field is irrelevant here. For example, if the tag is like this: `json:"labelSelector,inline"` the word `labelSelector` will not appear anywhere as the values are embedded.

There is an explanation about `omitempty` option which is somewhat similar to `inline` option.
https://golang.org/pkg/encoding/json/#Marshal

I can't find any explanation about `inline` anywhere. But to understand the behavior you can run this example: https://gist.github.com/baijum/e761bc8fb28ba3855e5d1cd118e9192b

Let me know if any other clarification required.

baijum:
> FYI: It is fine to have breaking changes with respect to labels since UI isn't consuming it?

Since the existing YAML structure is supported, I think we can consider it as backward compatible.
I am embedding the label selector. Let me know if this is fine.

akashshinde:
@sbose78 @baijum UI isn't really consuming `matchLabels` anywhere yet, so I assume it's safe to update this field.

baijum:
I have changed the API to use a nested structure.

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -63,11 +63,11 @@ type BackingServiceSelector struct {
 // +k8s:openapi-gen=true
 type ApplicationSelector struct {
 	// +optional
-	MatchLabels map[string]string `json:"matchLabels,omitempty"`
-	Group       string            `json:"group,omitempty"`
-	Version     string            `json:"version"`
-	Resource    string            `json:"resource"`
-	ResourceRef string            `json:"resourceRef"`
+	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
+	Group         string                `json:"group,omitempty"`
```
isutton: Perhaps GroupVersionResource can be embedded here as well?

baijum: We can do that in a separate PR.

PR #293, opened 2020-01-14
pre-req refactor to support multiple backing services

isutton:
### Motivation

This PR is a preparation for binding multiple backing services with the specified application.

### Changes

It introduces the ServiceBinder component, which is responsible for orchestrating the low-level components handling operational tasks such as updating Kubernetes resources related to the Service Binding Request.

ServiceBinder is still incomplete in this PR, but the changes made in the current tests should prove whether the existing functionality is still covered.

Since it is incomplete, its tests are currently skipped and will be enabled when all dependencies are in place.

### DIFF
--- a/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
+++ b/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
@@ -29,6 +29,9 @@ type ServiceBindingRequestSpec struct {
 	// BackingServiceSelector is used to identify the backing service operator.
 	BackingServiceSelector BackingServiceSelector `json:"backingServiceSelector"`
 
+	// BackingServiceSelectors is used to identify multiple backing services.
+	BackingServiceSelectors []BackingServiceSelector `json:"backingServiceSelectors"`
+
 	// ApplicationSelector is used to identify the application connecting to the
 	// backing service operator.
 	ApplicationSelector ApplicationSelector `json:"applicationSelector"`
--- a/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
+++ b/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
@@ -3,7 +3,6 @@ package v1alpha1
 import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 // Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
--- a/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
+++ b/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
@@ -3,6 +3,7 @@ package v1alpha1
 import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 // Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
--- a/pkg/controller/servicebindingrequest/binding.go
+++ b/pkg/controller/servicebindingrequest/binding.go
@@ -1,0 +1,319 @@
+package servicebindingrequest
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"gotest.tools/assert/cmp"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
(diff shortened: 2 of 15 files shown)


pkg/controller/servicebindingrequest/binding.go
```diff
@@ -0,0 +1,319 @@
...
+	// gather retriever's read data
+	// TODO: do not return error
+	retrievedData, err := retriever.Get()
+	if err != nil {
+		return nil, err
+	}
+
+	// gather related secret, again only appending it if there's a value.
+	secret := NewSecret(options.DynClient, plan)
+	secretObj, found, err := secret.Get()
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/controller/servicebindingrequest/binding.go
```diff
@@ -0,0 +1,319 @@
...
+var InvalidOptionsErr = errors.New("invalid options")
+
+// BuildServiceBinder creates a new binding manager according to options.
+func BuildServiceBinder(options *ServiceBinderOptions) (*ServiceBinder, error) {
+	if !options.Valid() {
+		return nil, InvalidOptionsErr
+	}
+
+	// objs groups all extra objects related to the informed SBR
+	objs := make([]*unstructured.Unstructured, 0, 0)
```
golangcibot: S1019: should use make([]*unstructured.Unstructured, 0) instead (from `gosimple`)

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -29,6 +29,9 @@ type ServiceBindingRequestSpec struct {
 	// BackingServiceSelector is used to identify the backing service operator.
 	BackingServiceSelector BackingServiceSelector `json:"backingServiceSelector"`
```
baijum: Is this kept for backward compatibility?

pkg/controller/servicebindingrequest/binding.go
```diff
@@ -0,0 +1,322 @@
...
+	namespacedName := types.NamespacedName{Namespace: b.SBR.GetNamespace(), Name: b.SBR.GetName()}
+	if err = SetSBRAnnotations(b.DynClient, namespacedName, b.Objects); err != nil {
+		b.Logger.Error(err, "On setting annotations in related objects.")
+		return b.onError(err, b.SBR, sbrStatus, updatedObjects)
+	}
+
+	// updating status of request instance
+	sbr, err := b.updateStatusServiceBindingRequest(b.SBR, sbrStatus)
+	if err != nil {
+		return RequeueOnConflict(err)
```
Avni-Sharma: Shouldn't this be RequeError(err)?

Avni-Sharma:
Calling BuildServiceBinder() would be implemented in a separate PR?

pkg/controller/servicebindingrequest/binding.go
```diff
@@ -0,0 +1,322 @@
...
+		objs = append(objs, secretObj)
+	}
+
+	return &ServiceBinder{
+		Logger:    options.Logger,
+		Binder:    NewBinder(ctx, options.Client, options.DynClient, options.SBR, retriever.volumeKeys),
+		DynClient: options.DynClient,
+		SBR:       options.SBR,
+		Objects:   objs,
+		Data:      retrievedData,
```
Avni-Sharma: Initially, we had a retrievedCache as well, that captures the hierarchical format of retrievedData. Are we capturing that or omitting that?

baijum:
@isutton The new code is not integrated with the reconciler, right? Is the integration going to happen in the next PR?

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
isutton: Yes; this PR is not meant to change functionality as of yet, just introduce some components and adjust some existing APIs they require so no user visible change should be introduced.

pkg/controller/servicebindingrequest/binding.go
isutton: Not all errors should be retried, for example a 404. I find it appropriate to only retry on conflicts, since this is the only indication the SBR still exists and can be updated.

pkg/controller/servicebindingrequest/binding.go
isutton: The hierarchical data format might be useful only when collecting and aggregating the configuration values and/or secrets; `retrievedData` is the aggregated information that will be persisted in a later stage.

isutton:
@Avni-Sharma @baijum Yes, `ServiceBinder` has not been integrated with the reconciler in this PR, but will at a later stage.

As I've discussed with @Avni-Sharma recently, the `ServiceBinder` can provide already a home for the conditions to be created and enter in the same planned SBR request update.

pkg/controller/servicebindingrequest/binding.go
Avni-Sharma: I am referring to this issue https://github.com/redhat-developer/service-binding-operator/issues/272 which was later fixed in https://github.com/redhat-developer/service-binding-operator/pull/281 .

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
sbose78: +1

pkg/controller/servicebindingrequest/binding.go
isutton: So we are keeping that. The behavior introduced in #281 hasn't been changed in this PR, and `ServiceBinder` will use the same `Retriever` implementation being currently used.

PR #295, opened 2020-01-17
Update examples with label selectors

baijum:
### Motivation

The example was not updated with #291 

### Changes

Update the example CR and CRD.

### Testing

No code change.


sbose78:
Aren't these files supposed to be auto-generated when the Golang API is updated?

baijum:
> Aren't these files supposed to be auto-generated when the Golang API is updated?

The CRD was auto-generated with a different file name: `apps.openshift.io_servicebindingrequests_crd.yaml`
I missed to commit it earlier. Now I copied to existing CRD.
Should I rename it to the default file name that is generated?

The CR should be updated manually.

sbose78:
Yeah the default file name would be nice to have.

sbose78:
In a different PR, I think we need to have some kind of (indirect) validation if the CRD inside `deploy/` isn't updated. @Avni-Sharma 's auto-release PR uses the yamls inside `/deploy` to generate releases, hence it's critical to ensure we catch misses using automation.

baijum:
> Yeah the default file name would be nice to have.

@sbose78 I have renamed the file

Avni-Sharma:
The community-operators one has the previous name https://github.com/operator-framework/community-operators/blob/master/community-operators/service-binding-operator/0.0.20/servicebindingrequests.apps.openshift.io.crd.yaml 
In the next release, we should be changing that

PR #307, opened 2020-01-24
Updates deploy cr

DhritiShikhar:
### Motivation

Updates deploy CR

### Changes

Updates deploy CR to include `resourceRef`


Avni-Sharma:
The alm-examples generated under the CSV consider the CR under deploy/cr.yaml which is failing at the community-operators side.

sbose78:
We would still need to fix the issue with resourceRef being mandatory, right?

DhritiShikhar:
@sbose78 I know, but at least we can get the community side PR merged first. 
Last time you mentioned that the Community side PR should not be holding an error for a long time.

sbose78:
Dhriti, right! Helps bump up community-operators PR up their review queue :-)

PR #311, opened 2020-02-03
validation of olm bundles part1

Avni-Sharma:
### Motivation

Refers - https://issues.redhat.com/browse/APPSVC-392

### Changes

[WIP]

### Testing

Pending

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -455,13 +455,15 @@ push-bundle-to-quay:
 .PHONY: dev-release
 ## validating the operator by installing new quay releases
 dev-release:
-	$(Q)kubectl delete operatorsources community-operators -n=openshift-marketplace
-	# Listing of the operator
-	$(Q)kubectl apply -f ./test/operator-hub/operator_source.yaml
+	$(Q)oc apply -f ./test/operator-hub/operator_source.yaml
 	# Subscribing to the operator
-	$(Q)kubectl apply -f ./test/operator-hub/subscription.yaml
-	$(eval VERSION_NUMBER := $(shell k get csvs  -n=default -o jsonpath='{.items[*].spec.version}'))
-	@if [ ${VERSION_NUMBER} -eq ${BUNDLE_VERSION} ]; \
+	$(Q)oc apply -f ./test/operator-hub/subscription.yaml
+	while [ "$(oc get pods --field-selector=status.phase=Running -n openshift-marketplace" == "No resources found." ]
+	do
+		sleep 1s
+	done
+	$(eval VERSION_NUMBER := $(shell oc get csvs  -n=default -o jsonpath='{.items[*].spec.version}'))
+	@if [ ${VERSION_NUMBER} == ${BUNDLE_VERSION} ]; \
 	then \
     	echo -e "OLM Bundle Version validation succeeded \n " \
 		exit 0; \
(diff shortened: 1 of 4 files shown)


test/operator-hub/catalog_source.yaml
```diff
@@ -6,6 +6,6 @@ metadata:
   namespace: openshift-operator-lifecycle-manager
 spec:
   sourceType: grpc
-  image: REPLACE_IMAGE
+  image: quay.io/redhat-developer/app-binding-operator:4c0a725
```
baijum: Why tag is hard-coded here?

baijum: I don't see this file is being used in the Makefile. Do we really need the catalog source?

test/operator-hub/subscription.yaml
```diff
@@ -2,10 +2,11 @@
...
   namespace: openshift-operators
 spec:
-  channel: dev
-  name: service-binding
-  source: my-catalog
-  sourceNamespace: openshift-operator-lifecycle-manager
+  channel: alpha
+  installPlanApproval: Automatic
+  name: service-binding-operator
+  source: example-operators
```
baijum: In `./test/operator-hub/operator_source.yaml`  the name is given as `my-catalog`. Can you please confirm if this is correct? Also, there is a `REPLACE_IMAGE` in that file. When is that value getting replaced?

baijum: Please ignore the above comment. That file doesn't exist. But it is referred from Makefile.

Makefile
```diff
@@ -455,5 +455,19 @@ push-bundle-to-quay:
 .PHONY: dev-release
 ## validating the operator by installing new quay releases
 dev-release:
-	echo  "HELLO BUNDLE VERSION -${BUNDLE_VERSION}"
-	sleep 10m
+	$(Q)kubectl delete operatorsources community-operators -n=openshift-marketplace
+	# Listing of the operator
+	$(Q)kubectl apply -f ./test/operator-hub/operator_source.yaml
```
baijum: This file doesn't exist. Did you forget to commit it?

Makefile
```diff
@@ -455,5 +455,19 @@ push-bundle-to-quay:
 .PHONY: dev-release
 ## validating the operator by installing new quay releases
 dev-release:
-	echo  "HELLO BUNDLE VERSION -${BUNDLE_VERSION}"
-	sleep 10m
+	$(Q)kubectl delete operatorsources community-operators -n=openshift-marketplace
```
sbose78: You sure the CVO operator which manages this wouldn't bring this resource back? 🤕

Makefile
```diff
@@ -455,5 +455,19 @@ push-bundle-to-quay:
...
+	# Listing of the operator
+	$(Q)kubectl apply -f ./test/operator-hub/operator_source.yaml
+	# Subscribing to the operator
+	$(Q)kubectl apply -f ./test/operator-hub/subscription.yaml
+	$(eval VERSION_NUMBER := $(shell k get csvs  -n=default -o jsonpath='{.items[*].spec.version}'))
```
sbose78: You might also want to check if installation has succeeded?

baijum: Replace `k` with `kubectl`

sbose78:
How do I validate this PR?

Makefile
```diff
@@ -455,5 +455,19 @@ push-bundle-to-quay:
...
+	@if [ ${VERSION_NUMBER} -eq ${BUNDLE_VERSION} ]; \
```
baijum: I think you need to use `==` instead of `-eq`.

baijum:
> How do I validate this PR?

The script is designed to run on master branch. It's counting the commits for validation. So, you can squash the commits in this PR and run the Makefile target against any cluster.

Avni-Sharma:
> How do I validate this PR?

https://prow.svc.ci.openshift.org/job-history/origin-ci-test/logs/periodic-ci-redhat-developer-service-binding-operator-master-dev-release

PR #316, opened 2020-02-06
update to a rhel8 base

sbose78:
### Motivation
Downstream is built using rhel8 base, let's do the same upstream too.

### Changes

The base image.

### Testing

Once merged, we should test with the latest OperatorSource


sbose78:
Tested, build has passed in https://quay.io/repository/redhat-developer/app-binding-operator?tab=builds

Dockerfile.rhel
```diff
@@ -32,9 +32,8 @@ RUN make build
 #--------------------------------------------------------------------
 
 # On CI , use
-# FROM registry.svc.ci.openshift.org/ocp/ubi-minimal:7
 
-FROM registry.access.redhat.com/ubi7/ubi-minimal
+FROM registry.access.redhat.com/ubi8/ubi
```
pedjak: why not `registry.access.redhat.com/ubi8/ubi-minimal`

Dockerfile.rhel
```diff
@@ -32,9 +32,8 @@ RUN make build
 #--------------------------------------------------------------------
 
 # On CI , use
```
pedjak: I would remove this line as well.

Dockerfile.rhel
sbose78: done https://github.com/redhat-developer/service-binding-operator/pull/316/commits/4c4b2a8ddc0919ec5b401c8b3593e8d91417a15c

Dockerfile.rhel
sbose78: done https://github.com/redhat-developer/service-binding-operator/pull/316/commits/4c4b2a8ddc0919ec5b401c8b3593e8d91417a15c

PR #321, opened 2020-02-07
Replace BindingStatus field with Kubernetes compatible conditions

Avni-Sharma:
Refer https://issues.redhat.com/browse/APPSVC-321

In case of success:-
```
status:
  applicationObjects:
  - service-binding-demo/nodejs-rest-http-crud-git
  bindingStatus: BindingSuccess
  conditions:
  - lastHeartbeatTime: "2020-01-28T10:30:15Z"
    lastTransitionTime: "2020-01-28T10:30:15Z"
    status: "True"
    type: Ready
  secret: node-todo-git
```
Note- Reason and message has been omitted here, as in case of success it is not necessary.

In case of failure, it can happen like below
```
status:
  conditions:
  - lastHeartbeatTime: "2020-01-28T11:57:18Z"
    lastTransitionTime: "2020-01-28T11:56:58Z"
    message: 'binding service selectors are empty'
    reason: EmptyBackingServiceSelectors
    status: "False"
    type: Ready
```
The above status is `False` when we do not provide the `backingServiceSelectors` in the service binding request. Other failure conditions have been mentioned in the JIRA above


sbose78:
Having this makes things look pretty in the UI

image

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -48,6 +49,8 @@ type ServiceBindingRequestSpec struct {
 type ServiceBindingRequestStatus struct {
 	// BindingStatus is the status of the service binding request.
 	BindingStatus string `json:"bindingStatus,omitempty"`
+	// conditions describes the state of the operator's reconciliation functionality.
```
baijum: s/conditions/Conditions

pkg/controller/servicebindingrequest/binding.go
```diff
@@ -60,26 +68,26 @@ type ServiceBinder struct {
...
+	// SBR is the ServiceBindingRequest associated with binding.
 	SBR *v1alpha1.ServiceBindingRequest
 	// Secret is the Secret associated with the Service Binding Request.
 	Secret *Secret
 }
 
-// updateServiceBindingRequest execute update API call on a SBR request. It can return errors from
-// this action.
-func (b *ServiceBinder) updateServiceBindingRequest(
+func UpdateServiceBindingRequest(
```
baijum: This method needs a documentation comment

baijum: Is this really need to be an exported method?

Avni-Sharma: yes, reconciler.go is using it

baijum: reconciler.go is in the same package, right?

PR #332, opened 2020-02-13
Check for CSV existence

baijum:
### Motivation

Make validation more reliable


sbose78:
May I have some context on when this gets executed?

baijum:
> May I have some context on when this gets executed?

Before comparing the actual dev release version which is there in CSV against the expected version.
And after the ensuring CRD has been creating through subscription.

PR #334, opened 2020-02-14
Refactor knative test: abridge test by removing unnecessary assertion.

pratikjagrut:
### Motivation

Refactor knative test

### Testing
`go test ./pkg/controller/servicebindingrequest -run TestKnativeServicesContractWithBinder -v` 
or
`make test-unit` to run all the unit tests

### DIFF
--- a/pkg/controller/servicebindingrequest/binder_test.go
+++ b/pkg/controller/servicebindingrequest/binder_test.go
@@ -300,19 +300,11 @@ func TestKnativeServicesContractWithBinder(t *testing.T) {
 
 	require.NotNil(t, binder)
 
-	updatedObjects, err := binder.Bind()
-	require.NoError(t, err)
-	require.Len(t, updatedObjects, 1)
+	t.Run("Knative service contract with service binding operator", func(t *testing.T) {
+		list, err := binder.search()
+		assert.Nil(t, err)
+		assert.Equal(t, 1, len(list.Items))
 
-	containersPath := []string{"spec", "template", "spec", "containers"}
-	containers, found, err := unstructured.NestedSlice(updatedObjects[0].Object, containersPath...)
-	require.NoError(t, err)
-	require.True(t, found)
-	require.Len(t, containers, 1)
+	})
 
-	c := corev1.Container{}
-	u := containers[0].(map[string]interface{})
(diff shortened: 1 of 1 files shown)


sbose78:
Curious, why is it unnecessary?

pratikjagrut:
> Curious, why is it unnecessary?

@sbose78 a few days back I asked @isutton about why tests in `binder_test` do not check binder.Bind(). So he said Bind() is tested in reconciler_test, here we just need to check if the resources are being available or not. So to make this knative test similar to all other tests I replaced binding assertion with the searching assertion.

pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -300,19 +300,11 @@ func TestKnativeServicesContractWithBinder(t *testing.T) {
...
-
-	c := corev1.Container{}
-	u := containers[0].(map[string]interface{})
-	err = runtime.DefaultUnstructuredConverter.FromUnstructured(u, &c)
-	require.NoError(t, err)
-	require.Equal(t, name, c.EnvFrom[0].SecretRef.Name)
+	t.Run("Knative service contract with service binding operator", func(t *testing.T) {
+		list, err := binder.search()
+		assert.Nil(t, err)
+		assert.Equal(t, 1, len(list.Items))
```
akashshinde: @pratikjagrut should we consider adding more assertion or is checking item length is enough ?

pratikjagrut: Yeah, kinda it's enough. search will add found knative service in the list, and we are creating only one service at that time, so the length is gonna be 1, and if it is greater, then something is fishy, which is not part of this test, and can be seen as a bug or test failure. SO I feel this assertion is enough and although all other present unit tests are tested on same assertion.

pratikjagrut: The aim here is just to make sure resources are being available.

pradeepto:
@sbose78 @baijum @akashshinde @otaviof Can we get a closure on this?

pkg/controller/servicebindingrequest/binder_test.go
isutton: I'm fine with this change. An issue should be created to include a case where `matchLabels` is `nil` and multiple knative services.

PR #337, opened 2020-02-18
Propagate Makefile failures to the travis.

pmacik:
### Motivation

The travis.ci passes even if any of the make commands fails - tha failures are thrown away.

### Changes

All the `make` commands under the `if`condition in the `.travis.yaml`'s `script` are executed in a way, which returns a non-zero exit code, in case any of those fails.

### Testing


baijum:
@pmacik 
Looks like Travis is failing https://travis-ci.org/redhat-developer/service-binding-operator/jobs/652043284#L345

PR #344, opened 2020-02-20
Roadmap for v1

isutton:
This PR defines the Roadmap for v1.

It contains an example use case from MVP to application decoupling that'll be used to guide the scope.

### DIFF
--- a/docs/Usage.md
+++ b/docs/Usage.md
@@ -1,0 +1,179 @@
+# Service Binding Operator Usage
+
+This document describes how the Service Binding Operator v1.0.0 would help the developer at
+different development stages.
+
+## Bootstrapping
+
+The first example would be a `webshop` application, which requires only a PostgreSQL database.
+The binding between both resources are as follow:
+
+```yaml
+---
+apiVersion: apps.openshift.io/v1
+kind: ServiceBinding
+metadata:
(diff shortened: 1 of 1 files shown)


sbose78:
Thanks a ton for getting this started. Are you using something as a reference for organizing the roadmap? I found https://github.com/tektoncd/pipeline/blob/master/roadmap-2019.md

isutton:
Thanks for the reference! 

I will start a roadmap document based on Tekton's.

sbose78:
API changes look good to me.

Things to do:

- Generic binding - in progress
- custom env var - can we specify which backing service ? ( important )
-  maybe it's just Workload --> Service ?


Workload ---> Service

```
apiVersion: apps.openshift.io/v1
kind: ServiceBinding
metadata:
  name: webshop
spec:
  applications:
  - group: apps
    version: v1
    resource: deployments
    name: nodefrontendapp
  services:
  - group: 
    version: corev1
    kind: Service
    name: GolangApp
  - group: 
    version: corev1
    kind: ConfigMap
    name: BackendServiceCredentials
```

sbose78:
Given this is stale, I'll close this. We'll re-open when we work on this later on.

PR #322, opened 2020-02-07
Support multiple backing services

isutton:
### Motivation

This change enables the user to specify multiple backing services in a ServiceBindRequest.

### Changes

`ServiceBinder` is now used to orchestrate the reconciliation process, and it is multiple services aware. It was introduced in an earlier PR together with internal changes required for multiple services support.

### Testing

Tests haven't been changed, only had their coverage increased as this is a major refactoring.

See `examples/multiple-services/example/README.md` for the demo use case, which offers a step-by-step guide to observe the functionality.

--------
```
apiVersion: apps.openshift.io/v1alpha1
kind: ServiceBindingRequest
metadata:
  name: node-todo-git
spec:
  applicationSelector:
    resourceRef: node-todo-git
    group: apps
    version: v1
    resource: deployments
  backingServiceSelectors:
  - group: postgresql.baiju.dev
    version: v1alpha1
    kind: Database
    resourceRef: db-demo
  - group: etcd.database.coreos.com
    version: v1beta2
    kind: EtcdCluster
    resourceRef: etcd-demo
  detectBindingResources: true
```

Issue #296, opened 2020-01-17
Is it possible to opt out from injecting environment variables?


navidsh:
Edit by Shoubhik 
Jira https://issues.redhat.com/browse/APPSVC-9

---

Is it possible to skip the "injecting environment variable" step after the controller created service binding secrets?

We are looking into integrating [Appsody Operator](https://github.com/appsody/appsody-operator/) with Service Binding Operator, and we want our operator to be in charge of modifying the resources (e.g. `Deployment`, `StatefulSet`) created by the Appsody operator to run users' applications.

sbose78:
Heya!


> Is it possible to skip the "injecting environment variable" step after the controller created service binding secrets?

We have a task on our planning board to only generate a managed secret which we never got to :) 


> Just to throw it out there, if applicationSelector is not defined in ServiceBindingRequest, SBO skips the injection.

You mean, it already works? That would be an intended feature then :)

navidsh:
> We have a task on our planning board to only only generate a managed secret which we never got to :)

Nice! Is there any timeline of when it will be implemented?

> You mean, it already works? That would be an unintended feature then :)

No, I think specifying `applicationSelector` is required now. I meant one way to do it is by not defining `applicationSelector` in the CR :)

sbose78:
I'll move it up the board, @navidsh .

@DhritiShikhar Could you start working on https://issues.redhat.com/browse/APPSVC-9 again :) , do have a chat with @isutton to avoid conflicts since there's a major refactoring happening in https://github.com/redhat-developer/service-binding-operator/pull/293 / https://github.com/redhat-developer/service-binding-operator/pull/286 .

sbose78:
Created a public copy of the feature request/story https://github.com/redhat-developer/service-binding-operator/issues/297

navidsh:
@sbose78 Thank you very much!

arthurdm:
thx guys!  this will allow for a seamless integration.  👍

sbose78:
Update:

We merged https://github.com/redhat-developer/service-binding-operator/pull/293 which had refactoring to support multiple backing CRs ( example: PostgreSQLService CR and a PostreSQLCredential CR ) ). 

As far as I remember, this was one of the use cases in some of the IBM operators which could be used as backing services.

sbose78:
@DhritiShikhar is working on this, and should have a PR this week.

arthurdm:
awesome, thanks guys.

[...]

PR #348, opened 2020-02-23
Rename to "applications" and "services"

sbose78:
( .. Based on @isutton 's ideas )

~### This is a proposal and open for discussion/debate.~

### Motivation

* Simplify the API, " connect `applications` to `services` "
* Ensure 100% backward compatibility, no breaking changes in this PR

### Changes

- Rename `spec.applicationSelector` to `spec.applications`
- Rename `spec.backingServiceSelectors` to `spec.services`
- Support a list of apps : `spec.applications`
- In order to maintain backward compatibility, `spec.applications`, `spec.applicationSelector` , `spec.backingServiceSelector` and `spec.backingServiceSelectors` are all optional. The appropriate parameters have been marked as deprecated. 
- In a future PR, when `spec.applicationSelector`,`spec.backingServiceSelector` and `spec.backingServiceSelectors` are removed, we will mark `spec.applications` and `spec.services` as mandatory. Doing so now will break integrations. 

```
kind: ServiceBindingRequest

metadata:
  name: node-golang-todo-git

spec:

  applications:
  - resourceRef: node-todo-git
    group: apps
    version: v1
    resource: deployments

  services:
  - group: postgresql.baiju.dev
    version: v1alpha1
    kind: Database
    resourceRef: db-demo
  - group: etcd.database.coreos.com
    version: v1beta2
    kind: EtcdCluster
    resourceRef: etcd-demo
```

Here's how the API looked before:

```
kind: ServiceBindingRequest
metadata:
  name: node-todo-git
spec:
  applicationSelector:
    resourceRef: node-todo-git
    group: apps
    version: v1
    resource: deployments
  backingServiceSelectors:
  - group: postgresql.baiju.dev
    version: v1alpha1
    kind: Database
    resourceRef: db-demo
  - group: etcd.database.coreos.com
    version: v1beta2
    kind: EtcdCluster
    resourceRef: etcd-demo
```

Care is being taken to maintain backward compatibility.


otaviof:
Overall, I like the proposal, the `ServiceBinding` will look a lot cleaner this way! Thanks.

> Removed support for label selectors in spec.applications

However, I think we should keep label selectors, since they are a general approach to handle Kubernetes based workloads. Therefore, users could choose on using a fully-qualified resource name or labels, should be up to final users.

sbose78:
@isutton , thoughts on removing label selectors?

isutton:
I've recently discussed about the labels topic with @otaviof and I'm fine supporting both direct references and label selectors for `applications`. The only question remaining is how to properly encode this information in the manifest.

One option might be keeping the `applicationSelector` field, but changing its internal fields to only represent or contain a Kubernetes' selector type, and introduce the `applications` field next to it containing a list of explicit application references.

This implies that watches should, if not already, be created to observe creation and updates of all supported applications' manifests in order to evaluate whether a binding should be applied to it or not.

sbose78:
We could have the following then:


```
kind: ServiceBindingRequest

metadata:
  name: node-golang-todo-git

spec:

  applications:
  - group: apps
    version: v1
    resource: deployments
    labels: 
        environment: production
        app: nginx
```

And

```
kind: ServiceBindingRequest

metadata:
  name: node-golang-todo-git

spec:

  applications:
  - group: apps
    version: v1
    resource: deployments
    resourceRef: foo
```

sbose78:
cc @arthurdm @bbrowning @pdettori  
Would love to hear your feedback/thoughts too.

arthurdm:
hey @sbose78 - I like the new look of the CRD, cleaner and easier to understand!  Assuming that https://github.com/redhat-developer/service-binding-operator/issues/296 would also apply to `applications`, all good!

### Side note (not meant to de-rail this PR, but since it also relates to updating the CRD)
We're looking to adopt (from the service binding spec) the ServiceBindingRequest CRD as the artifact by which applications use to request bindings.  We have identified that we would likely need 2 new keys (per service):  
* one defining a service account to be used by the Service Binding Operator (SBO) to read the Secret exposed by the bindable service (secured with RBAC)
* one defining the name of a Secret that contained complimentary binding data that SBO would combine with the data available from the bindable service.  This is the case where an external subscription entity is used to fetch per-user API keys, or credentials/URI for a provisioned Db, and those are made available in a Secret.  So SBO can combine it with the ones it finds, creating a single Secret (to optionally inject back into the app).

Probably this will make more sense once I write it down in a new PR to the spec, just fyi since you guys are thinking about changing the CRD with this PR. =)

The other thing is if `apiVersion: apps.openshift.io/v1alpha1` could be made more generic so it doesn't necessarily tie into OpenShift?     Some thoughts.

pdettori:
@sbose78 great, this looks definitely cleaner. Perhaps for another day/discussion, but have you thought about using a name for the CRD kind that looks more like a noun than an action ? That is, `ServiceBinding` rather than `ServiceBindingRequest` ? I think that would be more consistent with the way most CRD are named in Kubernetes and with desired state based management. A 'Request' seems more like a one-off thing, which I would not expect to keep maintaining the state of the binding, while I suppose that here the Service Binding Operator would reconcile a change in the backing service with the credentials injected in the app.

bbrowning:
+1 to `applications` and `services` as more concise names for the things being connected

sbose78:
Hi, @arthurdm @pdettori , Thank you!

> apiVersion: apps.openshift.io/v1alpha1

Yes, the group name must be changed. `openshift` must go from there. We'll track this and mark it for an openshift 4.5 timeframe because openshift consumes this API today.

> but have you thought about using a name for the CRD kind that looks more like a noun than an action ? That is, ServiceBinding rather than ServiceBindingRequest ?

Yes! We did consider this and there seems to be significant consensus to start calling it `servicebinding`. I'll come up with a swift plan.

> ( field for a `service` to define a service account to be used by the Service Binding Operator (SBO) to read the Secret exposed by the bindable service (secured with RBAC)

This should be optional to handle a situation where the 'default' service account used by the service binding operator doesn't have the privileges to read a specific secret? I wonder if that situation would arise because the service binding operator's service account can read/write secrets in all namespaces.

> one defining the name of a Secret that contained complimentary binding data that SBO would combine with the data available from the bindable service. 

In this case, are you proposing we skip creating a separate binding secret and instead manage the one provided by the service, with additional information in it?

arthurdm:
Hey @sbose78 - thanks for the feedback.  It's great to see these things are in motion.

> This should be optional to handle a situation where the 'default' service account used by the service binding operator doesn't have the privileges to read a specific secret? 

That's right.

> I wonder if that situation would arise because the service binding operator's service account can read/write secrets in all namespaces.

Interesting.  Is that something that could be configurable?  Just thinking if customers would want to have a finer grained RBAC for this, so that everything doesn't become bindable by anyone?

> In this case, are you proposing we skip creating a separate binding secret and instead manage the one provided by the service, with additional information in it?

Ya - probably populating the given Secret with anything missing / additional would be easiest.  Another approach is to just copy the information from the given Secret into the one SBO creates/manages/injects.

sbose78:
> Interesting. Is that something that could be configurable? Just thinking if customers would want to have a finer grained RBAC for this, so that everything doesn't become bindable by anyone?

Custom service accounts would be hard since this operator needs to be watch all namespaces.

I propose a validation webhook & subject-access-review(SAR)-based design to ensure we validate "Does John have the privileges to modify deployment foo?"

What do you think?

arthurdm:
ya - perhaps there's 2 sides to this:  
(a) does the requester have authority to create a binding to this bindable resource?  
(b) (if application injection is not toggled off) does the requester have authority to modify the deployment with the binding data?

From a spec perspective It would be nice to have a single "thing" that we pass into the ServiceBindingRequest where implementations (in this case, SBO) could check for both (a) and (b).  I was thinking a service-account was a native way to achieve it, but you bring up a good point that the SBO would already have permissions for all of them.  In a validation / SAR design, what would be the artifact we pass around?   (let me know if we should move this into a new issue 😄  )

sbose78:
> let me know if we should move this into a new issue 

yes yes :)

sbose78:
If both labels and resourceRef are present in `applications[i]`, should we compile a cumulative list of
- objects by resourceRef
- objects by label selectors ?

@DhritiShikhar @isutton

pkg/controller/servicebindingrequest/binding_test.go
```diff
@@ -229,6 +230,42 @@ func TestServiceBinder_Bind(t *testing.T) {
...
+				{
+					GroupVersionResource: metav1.GroupVersionResource{
+						Group:    d.GetObjectKind().GroupVersionKind().Group,
+						Version:  d.GetObjectKind().GroupVersionKind().Version,
+						Resource: "deployments",
+					},
+					ResourceRef: d.GetName(),
+				},
+			},
+			BackingServiceSelectors: &[]v1alpha1.BackingServiceSelector{
```
Avni-Sharma: This should be `Services` instead of `BackingServiceSelectors` .

pkg/controller/servicebindingrequest/planner_test.go
```diff
@@ -21,13 +21,73 @@ func TestPlannerNew(t *testing.T) {
...
+	})
+}
+
+func TestPlannerAnnotation(t *testing.T) {
+	ns := "planner"
+	name := "service-binding-request"
+	resourceRef := "db-testing"
+
+	f := mocks.NewFake(t, ns)
+	sbr := f.AddMockedServiceBindingRequestV1_1(name, resourceRef, "", deploymentsGVR)
```
Avni-Sharma: V1_1 ?

deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
```diff
@@ -36,7 +36,7 @@ spec:
           properties:
             applicationSelector:
               description: ApplicationSelector is used to identify the application
-                connecting to the backing service operator.
+                connecting to the backing service operator. Deprecated.
```
pedjak: I would move `Deprecated` message upfront and make more descriptive, like when, and how to migrate, etc. So at the end of the paragraph, many would overlook it.

deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
```diff
@@ -98,9 +98,77 @@ spec:
...
+                required:
+                - group
+                - resource
+                - version
+                type: object
+              type: array
             backingServiceSelector:
               description: BackingServiceSelector is used to identify the backing
-                service operator.
+                service operator. Deprecated.
```
pedjak: see above.

pkg/controller/servicebindingrequest/planner_test.go
sbose78: Good question, the function name wasn't too explanatory, let me add a comment.
This function returns a ServiceBindingRequest with the newer API consisting of "applications" & "services".

isutton:
> If both labels and resourceRef are present in `applications[i]`, should we compile a cumulative list of
- objects by resourceRef
- objects by label selectors ?

Since `resourceRef` isn't a mask, if present the search will result always in slice with either 0 or 1 element. 

If the manifest does indeed exist, then the label selectors will be used to evaluate whether it should be in the resulting application set. This could be used to selectively add or remove an application from those the Service Binding Operator manages without directly modifying the service binding manifest, or prevent the binding to happen if the application doesn't yet have a specific label.

When `resourceRef` is not specified the search can result into a set containing potentially the same number of records existing in the backend for the given GVK, being restricted only by matching each element's labels with the specified application's labels.

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -28,21 +28,40 @@ type ServiceBindingRequestSpec struct {
 	// +optional
 	CustomEnvVar []corev1.EnvVar `json:"customEnvVar,omitempty"`
 
+	// NOTE: Services & Applications will be made non-optional
+	// after deprecated fields are removed.
+
+	// Services are used to identify multiple backing services.
+	// +optional
+	Services *[]BackingServiceSelector `json:"services,omitempty"`
```
isutton: There's no need for the pointer there; see https://medium.com/rungo/the-anatomy-of-slices-in-go-6450e3bb2b94.

pkg/controller/servicebindingrequest/planner.go
```diff
@@ -76,7 +76,18 @@ var EmptyBackingServiceSelectorsErr = errors.New("backing service selectors are
 // Plan by retrieving the necessary resources related to binding a service backend.
 func (p *Planner) Plan() (*Plan, error) {
 	ns := p.sbr.GetNamespace()
-	selectors := append([]v1alpha1.BackingServiceSelector{}, p.sbr.Spec.BackingServiceSelectors...)
+
+	var selectors = []v1alpha1.BackingServiceSelector{}
+	if p.sbr.Spec.Services != nil {
+		selectors = append(selectors, *p.sbr.Spec.Services...)
```
isutton: Removing the pointer in the `Services` definition removes the need for dereference here.

pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -162,6 +162,42 @@ func TestBinderNew(t *testing.T) {
...
+		list, err := binder.search()
+		require.Error(t, err)
+		require.Nil(t, list)
+	})
+
+	sbr = f.AddMockedServiceBindingRequest(name, &ns, "foo", "app", deploymentsGVR, map[string]string{})
+	f.AddMockedUnstructuredDeployment("bar", map[string]string{})
+	binder.sbr = sbr
+
+	require.NotNil(t, binder)
```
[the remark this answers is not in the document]
sbose78: Not sure why I have `nil` check after `binder.sbr = sbr` 🤦‍♂  , removing.

pkg/controller/servicebindingrequest/planner_test.go
sbose78: Added comment

```
// AddMockedServiceBindingRequestV1_1 add mocked object from ServiceBindingRequestMock.
// This function returns a ServiceBindingRequest with the newer API
// consisting of "applications" & "services".
```

sbose78:
> Since resourceRef isn't a mask, if present the search will result always in slice with either 0 or 1 element.

right.

> If the manifest does indeed exist, then the label selectors will be used to evaluate whether it should be in the resulting application set. This could be used to selectively add or remove an application from those the Service Binding Operator manages without directly modifying the service binding manifest, or prevent the binding to happen if the application doesn't yet have a specific label. When resourceRef is not specified the search can result into a set containing potentially the same number of records existing in the backend for the given GVK, being restricted only by matching each element's labels with the specified application's labels.


I agree with this. Should we do it in a different PR because it's a change in business logic? I'd like to keep the business logic unchanged.

pkg/controller/servicebindingrequest/binding_test.go
```diff
@@ -330,7 +367,7 @@ func TestServiceBinder_Bind(t *testing.T) {
 				},
 				ResourceRef: d.GetName(),
 			},
-			BackingServiceSelectors: []v1alpha1.BackingServiceSelector{
+			BackingServiceSelectors: &[]v1alpha1.BackingServiceSelector{
```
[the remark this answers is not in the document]
sbose78: The reason I prefer using pointers is that it helps our code distinguish between the following:
- if the user has specified an empty field 
- if the user has skipped the field

When `*BackingServiceSelectors` is used:
- when nil, we know that the field was skipped in the CR.
- when `[]` we know that an empty list was passed in the CR.

When `BackingServiceSelectors` is used:
- when [], the field might have been skipped in the CR. OR the user might have specified an empty list in the CR <-- this is an ambiguity.

To avoid this ambiguity, I prefer using pointers for optional fields because it lets us write code to handle scenarios as applicable.

Moreover,  https://github.com/redhat-developer/service-binding-operator/pull/369

deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
sbose78: Added a long message, check?

pkg/controller/servicebindingrequest/planner_test.go
```diff
@@ -4,11 +4,10 @@ import (
 	"context"
 	"testing"
 
-	"github.com/stretchr/testify/require"
-	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
-
 	"github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1"
 	"github.com/redhat-developer/service-binding-operator/test/mocks"
+	"github.com/stretchr/testify/require"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
```
golangcibot: SA1019: Package sigs.k8s.io/controller-runtime/pkg/runtime/log is deprecated: use pkg/log  (from `staticcheck`)

isutton:
# Response

> The reason I prefer using pointers is that it helps our code distinguish between the following:
> - if the user has specified an empty field 
> - if the user has skipped the field
> 
> 1. `*BackingServiceSelectors` is used:
>     - when nil, we know that the field was skipped in the CR.
>     - when `[]` we know that an empty list was passed in the CR.
> 
> 2. `BackingServiceSelectors` is used:
>     - when `[]`, the field might have been skipped in the CR. OR the user might have specified an empty list in the CR <-- this is an ambiguity.
>
> To avoid this ambiguity, I prefer using pointers for optional fields because it lets us write code to handle scenarios as applicable.
>
> Moreover, https://github.com/redhat-developer/service-binding-operator/pull/369

I'm not sure this scenario in specific require such semantics, since whenever `BackingServiceSelectors` is empty, it will produce an empty set of resources, which might or not be an issue.

Can you provide a case for `BackingServiceSelectors` that requires this distinction?

sbose78:
> I'm not sure this scenario in specific require such semantics, since whenever BackingServiceSelectors is empty, it will produce an empty set of resources, which might or not be an issue.
Can you provide a case for BackingServiceSelectors that requires this distinction?

I can't think of a need to distinct those scenarios at the moment. I usually do this as a way to allow application of custom logic in future, if needed.

isutton:
> > The reason I prefer using pointers is that it helps our code distinguish between the following:
> >
> > [snip]
> >
> > 2. `BackingServiceSelectors` is used:
> >    
> >    * when `[]`, the field might have been skipped in the CR. OR the user might have specified an empty list in the CR <-- this is an ambiguity.

I agree that some situations might require this approach (mostly when dealing with scalars and default values), but in this operator's particular case I'm unsure whether those precautions are required, since there are no semantic differences between an unpopulated or nil slice; additionally if any clean-up or amendment if necessary should be based on the differences between the previous state and the new desired state.

sbose78:
I had actually moved out the change to a different PR because of a different observation as well https://github.com/redhat-developer/service-binding-operator/pull/369 . This PR no longer introduces that change.

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
Avni-Sharma: This has been added to avoid `API rule violation: list_type_missing,github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1,ServiceBindingRequestSpec, Services ,` I believe

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -28,28 +28,44 @@ type ServiceBindingRequestSpec struct {
...
 	// backing service operator.
+	// Deprecated.
 	// +optional
-	ApplicationSelector ApplicationSelector `json:"applicationSelector"`
-
-	// DetectBindingResources is flag used to bind all non-bindable variables from
-	// different subresources owned by backing operator CR.
-	// +optional
-	DetectBindingResources bool `json:"detectBindingResources"`
+	ApplicationSelector *ApplicationSelector `json:"applicationSelector"`
```
Avni-Sharma: this should be omitempty

sbose78: Good catch, fixed :)

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
otaviof: @sbose78 I think it's worth removing this pointer, btw.

sbose78: @otaviof the change isn't really in this PR, the change is in https://github.com/redhat-developer/service-binding-operator/pull/369 .

ron1:
Up until now, it has been my understanding that a single ServiceBindingRequest CR was intended to bind one or more applications to a **single** backing service. The motivation for `backingServiceSelectors` being plural was to support complex backing services that had needed binding properties spread around in multiple resources.

Is the name change from `backingServiceSelectors` to `services` intended to change the scope of the SBR CR such that now it binds one or more applications to **multiple** backing services rather than a single backing service? If so, then I think it would be good to have a discussion about the pros and cons of such a scope change. To me, this is much bigger than a simple name change.

OTOH, if this name change does not change the scope and a SBR CR still represents binding to a **single** backing service with potentially multiple selectors, then maybe the name change should be from `backingServiceSelectors` to the more simple `serviceSelectors`. With this name, it is more clear that a single service is the binding target resource and that multiple selectors may be required in order to capture all the binding information.

WDYT?

isutton:
>
> Up until now, it has been my understanding that a single
> ServiceBindingRequest CR was intended to bind one or more applications to a
> *single* backing service. The motivation for backingServiceSelectors
> being plural was to support complex backing services that had needed
> binding properties spread around in multiple resources.
>
This behavior has already been changed in
https://github.com/redhat-developer/service-binding-operator/pull/322, by
accepting more than one service selector or direct reference to a service.

The rationale behind it is that typically applications depend on multiple
services (databases, cache services, etc) and the configuration values
exposed by those services are, again typically, required for the
well-functioning of said applications. This simplifies the operator's work,
since it has less related resources to manage (secrets or config maps).

> Is the name change from backingServiceSelectors to services intended to
> change the scope of the SBR CR such that now it binds one or more
> applications to *multiple* backing services rather than a single backing
> service? If so, then I think it would be good to have a discussion about
> the pros and cons of such a scope change. To me, this is much bigger than a
> simple name change.
>
The behavior you've mentioned hasn't changed; one can still use one service
entry with a service selector. It was added the ability to have more of
those, although I agree that a change such as removing the service selector
functionality should be discussed, although I'm fine with removing it since
I would like to control which dependencies are being included in my
application, which I don't have in the case of supporting service selectors.

> OTOH, if this name change does not change the scope and a SBR CR still
> represents binding to a *single* backing service with potentially
> multiple selectors, then maybe the name change should be from
> backingServiceSelectors to the more simple serviceSelectors. With this
> name, it is more clear that a single service is the binding target resource
> and that multiple selectors may be required in order to capture all the
> binding information.
>
We had some internal discussions regarding this topic, and I personally
consider it fine to have selectors for applications but not so much for
services.

I would love to hear the use cases you have in mind for service selectors,
it'd be really helpful.

ron1:
@isutton Thanks for your reply.

In complex binding scenarios as is the case for [Kafka with mutual TLS](https://github.com/strimzi/strimzi-kafka-operator/blob/790793f351b508118a6681a5dbe742aabc4955e6/design/service-binding.md#consuming-clients-servicebindingrequest-4), the use of two independent service "entries" for the same Kafka service seems unintuitive. In the binding secret at least, shouldn't the fact that I needed two service "entries" in order to bind to this one complex Kafka service be hidden? Is the thought that each service "entry" will have its own optional envvarprefix or id or mountpath and the two independent Kafka service "entries" will be associated with one another by assigning them the same envvarprefix/id/mountpath?

arthurdm:
>we've been discussing the idea of
aggregate the configuration values collected from the services, and
grouping those accordingly.

@isutton - in this case, would you not just simply add the service's ID, as per [this section](https://github.com/application-stacks/service-binding-specification#id-prefix)? This section of the spec will apply also to injected values once [this issue](https://github.com/application-stacks/service-binding-specification/issues/15) is written down, so it's consistent with mounted values.  Would make sense to store these values with their IDs in the secret themselves.

ron1:
My point is simply that Kafka and KafkaUser should.not be treated as separate services. Instead their properties should be globbed together in the binding secret. How does the SBR know to glob these two services together?

isutton:
>
> @isutton <https://github.com/isutton> - in this case, would you not just
> simply add the service's ID, as per this section
> <https://github.com/application-stacks/service-binding-specification#id-prefix>?
> This section of the spec will apply also to injected values once this
> issue
> <https://github.com/application-stacks/service-binding-specification/issues/15>
> is written down, so it's consistent with mounted values. Would make sense
> to store these values with their IDs in the secret themselves.
>
Each k8s resource already has an identifier, even if verbose: apiGroup,
kind and name. The "id" field is, from my understanding, a shortcut for
that tuple, and it might be error prone due to lack of type safety in both
the value assignment and template definition.

Adding this feature to be able to expose an environment variable DBUSER_HOST
instead of DATABASE_USER_HOST, or a mount point dbuser instead of
database/user seems overkill to me in this particular, but I would love to
hear more about your thoughts on this and perhaps other use cases.

isutton:
>
> My point is simply that Kafka and KafkaUser should.not be treated as
> separate services. Instead their properties should be globbed together in
> the binding secret. How does the SBR know to glob these two services
> together?
>
To be more clear, following up the previous Kafka and KafkaUser examples,
all the exposed configuration values specified in their CRDs will become
available to all applications bound to those services in order to be used
in Go template expressions to compose the intermediate resources that are
then referenced in application resources.

For example, in Go template those services will be accessible through the
path .Kafka.my-cluster.status.listeners, or
.KafkaUser.my-user.status.username, and
KAFKA_MY_CLUSTER_STATUS_LISTENERS_0_ADDRESSES_0_HOST,
KAFKA_MY_CLUSTER_STATUS_LISTENERS_0_ADDRESSES_0_PORT
and so on, and KAFKAUSER_MY_USER_STATUS_USERNAME.

The binding secret contains the configuration values that have been exposed
by all the declared services that currently exist in the cluster, so I tend
to think we're on the same page and this direction meets your needs. Please
let me know if I missed something.

ron1:
I believe an intended goal is to provide a canonical [Service Binding Schema](https://github.com/application-stacks/service-binding-specification#service-binding-schema) for backing services. In the above example, Kafka is a single service but requires two service "entries" to accomplish its binding. Some of the binding schema elements are extracted from the Kafka CR and others are extracted from the KafkaUser CR. From a binding schema perspective, all binding properties should appear within a single "binding root" regardless of the CR from which they were extracted.

In other words, I do not want a Kafka "binding root" in the resulting binding secret that has the following subset of binding schema elements:

type - kafka
endpoints.<listener_name> - bootstrap server information for all listeners (TBD)
ca.p12 - CA certificate PKCS #12 archive file for storing certificates and keys (binding:env:object:secret:ca.p12)
ca.password - password for protecting the CA certficate PKCS #12 archive file (binding:env:object:secret:ca.password)

and another KafkaUser "binding root" that has the remaining subset of binding schema elements:

username - username for the consuming client (binding:env:object:secret:username)
user.p12 - user certificate for the consuming client PKCS #12 archive file for storing certificates and keys (binding:env:object:secret:user.p12)
user.password - password for protecting the user certficate PKCS #12 archive file (binding:env:object:secret:user.password)

Does this clarify the issue?

deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
```diff
@@ -99,10 +99,78 @@ spec:
...
+                - resource
+                - version
+                type: object
+              type: array
             backingServiceSelector:
-              description: 'BackingServiceSelector is used to identify the backing
-                service operator. Deprecation Notice: In the upcoming release, this
-                field would be depcreated. It would be mandatory to set "backingServiceSelectors".'
+              description: BackingServiceSelector is used to identify the backing
+                service operator. Deprecated. In the upcoming release, this field
```
Avni-Sharma, line 172: backing service operator or any other k8s resource

deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
```diff
@@ -99,10 +99,78 @@ spec:
               - resource
               - version
               type: object
+            applications:
+              description: Applications are used to identify the application connecting
```
Avni-Sharma, line 103: multiple applications

deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
```diff
@@ -251,8 +319,34 @@ spec:
               description: EnvVarPrefix is the prefix for environment variables
               type: string
             mountPathPrefix:
-              description: MountPathPrefix is the prefix for volume mount
+              description: "Important: Run \"operator-sdk generate k8s\" to regenerate
```
Avni-Sharma, line 322: description here should not change, the previous one looks good ` MountPathPrefix is the prefix for volume mount`

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -25,28 +25,44 @@ type ServiceBindingRequestSpec struct {
...
+
+	// Applications are used to identify the application connecting to the
+	// backing service.
+	// +optional
+	Applications *[]ApplicationSelector `json:"applications,omitempty"`
+
+	// DetectBindingResources is flag used to bind all non-bindable variables from
+	// different subresources owned by backing operator CR.
+	// +optional
+	DetectBindingResources bool `json:"detectBindingResources"`
```
Avni-Sharma, line 43: omitempty should be included for `+optional` fiellds

deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
sbose78: This is auto-generated from
```
        // Applications are used to identify the application connecting to the
	// backing service.
	// +optional
	Applications *[]ApplicationSelector `json:"applications,omitempty"`
```

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
sbose78: Do you mind if I fixed it in a different PR, since the change isn't related?

deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
sbose78: fixed.

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
Avni-Sharma: sure

pkg/controller/servicebindingrequest/planner.go
```diff
@@ -85,6 +85,9 @@ func (p *Planner) Plan() (*Plan, error) {
 	if p.sbr.Spec.BackingServiceSelectors != nil {
 		selectors = append(selectors, *p.sbr.Spec.BackingServiceSelectors...)
 	}
+	if p.sbr.Spec.Services != nil {
```
Avni-Sharma, line 88: A similar condition for p.sbr.Spec.Applications is missing that should be checked in a similar was as for https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/controller/servicebindingrequest/planner.go#L83

sbose78: https://github.com/redhat-developer/service-binding-operator/pull/414/files#r407707455 
On this, could we please have a chat here ^ @Avni-Sharma @DhritiShikhar

sbose78:
Replaced by #495

Issue #354, opened 2020-03-02
customEnvVar templates getting incorrectly resolved to "<no value>"


navidsh:
I installed Service Binding Operator `v0.1.0-249` from OperatorHub with the default configuration (watching all namespaces). I tried the `examples/nodejs_postgresql` example and added two env under `customEnvVar`. 

```yaml
apiVersion: apps.openshift.io/v1alpha1
kind: ServiceBindingRequest
metadata:
  name: binding-request
  namespace: service-binding-demo
spec:
  applicationSelector:
    group: apps
    resource: deployments
    resourceRef: nodejs-rest-http-crud
    version: v1
  backingServiceSelector:
    group: postgresql.baiju.dev
    kind: Database
    resourceRef: db-demo
    version: v1alpha1
  customEnvVar:
    - name: MY_DB_NAME
      value: '{{ .spec.dbName }}'
    - name: MY_DB_IP
      value: '{{ .status.dbConnectionIP }}'
```

A secret got created and has the custom environment variables but with `<no value>`  instead of actual values from my backing service CR:
```yaml
data:
...
  MY_DB_IP: <no value>
  MY_DB_NAME: <no value>
```

I also tried on running the operator locally based on [`master@24a8d0e`](https://github.com/redhat-developer/service-binding-operator/tree/24a8d0e8b9658dfc88c9b66c39915a8cc03d273e), and see the same behavior. This used to work on `v0.0.24-214`.

sbose78:
Ouch, let me check cc @Avni-Sharma  @isutton

isutton:
Hi, 

First, thanks for reporting this, it's very much appreciated :)

Recently we introduced the ability of binding multiple services with an application, and might have not covered this case after the change.

We'll work to solve that shortly.

sbose78:
This is being actively worked upon. As part of the effort, we'll slightly change how a user specifies the customEnvVar template, to bring into consideration multiple backing services

https://issues.redhat.com/browse/APPSVC-478

**Bug Description**
On adding multiple backing services, we've end up causing a issue for customEnvVar. Therefore, we need to update the approach to collect data, and allow users to find all data regardless of how many backends are informed.


**Solution**
Currently the expressions are having the .spec.<attribute>, or .status.<attribute>, and therefore it needs to support multiple backing services. To add support, the expressions need to take Kubernetes Resource Kind and resourceRef in consideration, like the following:

[...]

PR #357, opened 2020-03-03
Add example for using the Service Binding Operator with the IBM Cloud Operator

pdettori:
### Motivation

Add example for using the Service Binding Operator with an out-of-cluster service managed by the IBM Cloud Operator

### Changes

Documentation and yaml templates for the example have been added to the examples directory. 

### Testing

No code has been changed since this is just documenting an example of usage, so no new e2e tests have been added.


sbose78:
Thanks a lot, we'll fix https://github.com/redhat-developer/service-binding-operator/issues/354 and then have this tested

pdettori:
> Thanks a lot, we'll fix #354 and then have this tested

Thanks @sbose78, yes, we'll need that fixed and will also need to update the custom env vars with the new syntax.

pdettori:
@sbose78 I opened #358 to track a blocking issue that is preventing me from retesting the scenario.

sbose78:
I'll make a PR to your branch with the updated syntax, Paolo.

sbose78:
https://github.com/redhat-developer/service-binding-operator/pull/360
fix is merged, existing syntax is intact with `backingServiceSelectors`, could you validate this and let me know?

pdettori:
@sbose78 I just finished testing using an image built from master. It works ! However I had to do some workaround to make it work for my OLM installed operator. First, I had to install it in a single namespace, second, I updated the image in csv and deployment to use the image I built from master, third, I had to do some hacking of roles - see https://github.com/redhat-developer/service-binding-operator/issues/358

sbose78:
Nice!

The easiest way would be to run "make local" to test master.

pdettori:
> The easiest way would be to run "make local" to test master.
@sbose78 good point ! I just wanted to test if #358 was still there using the master build. It seems now it might be related to the specifics of the RBAC rules for the OLM install.

sbose78:
Ah, missed updating the README https://github.com/redhat-developer/service-binding-operator#example-scenarios

PR #359, opened 2020-03-04
Add test for conflicting resource ref and labels

pratikjagrut:
### Motivation

https://issues.redhat.com/browse/APPSVC-248

###
testing:
`make test-unit`
or
`go test ./pkg/controller/servicebindingrequest -run TestReconcilerReconcileWithConflictingAppSelc -v `

### DIFF
--- a/pkg/controller/servicebindingrequest/reconciler_test.go
+++ b/pkg/controller/servicebindingrequest/reconciler_test.go
@@ -296,7 +296,6 @@ func TestReconcilerReconcileWithConflictingAppSelc(t *testing.T) {
 	f.AddMockedUnstructuredDeployment(applicationResourceRef1, matchLabels1)
 	f.AddMockedUnstructuredDeployment(applicationResourceRef2, nil)
 	f.AddMockedUnstructuredServiceBindingRequest(reconcilerName, backingServiceResourceRef, applicationResourceRef2, deploymentsGVR, matchLabels1)
-	f.AddMockedUnstructuredCSV("cluster-service-version-list")
 	f.AddMockedUnstructuredDatabaseCRD()
 	f.AddMockedUnstructuredDatabaseCR(backingServiceResourceRef)
 	f.AddMockedSecret("db-credentials")
--- a/pkg/controller/servicebindingrequest/reconciler_test.go
+++ b/pkg/controller/servicebindingrequest/reconciler_test.go
@@ -315,7 +315,6 @@ func TestReconcilerReconcileWithConflictingAppSelc(t *testing.T) {
 		sbrOutput, err := reconciler.getServiceBindingRequest(namespacedName)
 		require.NoError(t, err)
 
-		require.Equal(t, BindingSuccess, sbrOutput.Status.BindingStatus)
 		expectedStatus := v1alpha1.BoundApplication{
 			GroupVersionKind: v1.GroupVersionKind{
 				Group:   deploymentsGVR.Group,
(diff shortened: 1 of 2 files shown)


sbose78:
> Skipping this test due to an unresolved issue #319

So, this test wouldn't actually run ?

pratikjagrut:
> > Skipping this test due to an unresolved issue #319
> 
> So, this test wouldn't actually run ?

Yes, it will not, until #319 is fixed.

sbose78:
Could you fix #319 ?

sbose78:
( ask @baijum for assistance if needed..)

pratikjagrut:
> Could you fix #319 ?

I'm digging into it to find the crux of this issue, but I might gonna need help.

pratikjagrut:
Ok, cool, I'll get in touch with @baijum.

sbose78:
How's this going?

pratikjagrut:
@sbose78 Looks like search() of binder.go is the crux.
[binder.go->L76](https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/controller/servicebindingrequest/binder.go#L76)
The list is returning everything for the given GVR even if options only contains resourceRef.

sbose78:
Can we fix it ?

pratikjagrut:
> Can we fix it ?

I'm working on it?

pratikjagrut:
@sbose78 To close the story we can merge this PR, and once #319 is fixed I'll send another PR to remove the skip. I'm not sure how much time is going to take that bug to fix. As @baijum is busy with something I asked @akashshinde to help me with this.

pratikjagrut:
@sbose78 #319 is not fixed yet but I found a workaround now this test works fine, this PR is oK to merge.

pkg/controller/servicebindingrequest/reconciler_test.go
```diff
@@ -184,3 +184,43 @@ func TestReconcilerReconcileUsingVolumes(t *testing.T) {
...
+
+		res, err := reconciler.Reconcile(reconcileRequest())
+		require.NoError(t, err)
+		require.False(t, res.Requeue)
+
+		namespacedName := types.NamespacedName{Namespace: reconcilerNs, Name: reconcilerName}
+		sbrOutput, err := reconciler.getServiceBindingRequest(namespacedName)
+		require.NoError(t, err)
+
+		require.Equal(t, BindingSuccess, sbrOutput.Status.BindingStatus)
```
DhritiShikhar: Can you add a check for `sbrOutput.Status.Conditions` as well?

pkg/controller/servicebindingrequest/reconciler_test.go
```diff
@@ -184,3 +184,43 @@ func TestReconcilerReconcileUsingVolumes(t *testing.T) {
...
+		expectedStatus := fmt.Sprintf("%s/%s", reconcilerNs, applicationResourceRef2)
+		require.Equal(t, expectedStatus, sbrOutput.Status.ApplicationObjects[0])
```
DhritiShikhar: Can you add a check that length of `sbrOutput.Status.ApplicationObjects` is 1 and there are no other applicationObjects?

pkg/controller/servicebindingrequest/reconciler_test.go
pratikjagrut: If we have `BindingSuccess` asserted then is there any use of asserting conditions?

DhritiShikhar: @pratikjagrut We want to support conditions and deprecate BindingStatus in future

Avni-Sharma:
@pratikjagrut  can you add a case like this one https://github.com/redhat-developer/service-binding-operator/commit/c0f53c17f7bbb85b8bae6df48b1147643165e385#diff-14a37dadc390fc3ec23459efc088de8eR256 . Apart from this, it looks good to me

Avni-Sharma:
Cleanup of tests will happen in another PR as mentioned by @pratikjagrut . Merging this

pkg/controller/servicebindingrequest/reconciler_test.go
pratikjagrut: I added condition assertion.

pratikjagrut:
> Cleanup of tests will happen in another PR as mentioned by @pratikjagrut . Merging this

I removed mock CSV, so no other PR is needed.

PR #360, opened 2020-03-04
Use visited paths as input for CustomEnvParser instead of processed environment variables

isutton:
### Motivation

This PR adds a test case for the #354, and provides a solution for it without taking in consideration changes in the custom environment variables syntax to accommodate the multiple services feature.

### Changes

The SBR's secret data was being produced prior to persisting it in the API server whenever a service binding was being reconciled but the underlying data doesn't change after it is initially read. 

This PR moves CustomEnvParser from secret to retriever (which owns the visited paths used as input for the CustomEnvParser), and generates the final secret data right after services have been discovered.

### DIFF
--- a/pkg/controller/servicebindingrequest/binding_test.go
+++ b/pkg/controller/servicebindingrequest/binding_test.go
@@ -1,9 +1,11 @@
 package servicebindingrequest
 
 import (
+	"encoding/base64"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	"github.com/redhat-developer/service-binding-operator/pkg/conditions"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


sbose78:
Curious, are you planning to push a fix with the old behaviour, with the assumption that the user has only provided one backing Service?

I like that plan too. We could make the enhancements separately.

Anyway, it's totally up to you :-)

sbose78:
CC @pdettori  / keeping you in the loop.

Avni-Sharma:
Screenshot from 2020-03-05 00-26-36

Avni-Sharma:
Steps to recreate -
- Import an application- https://github.com/ldimaggi/java-rest-http-crud
- Install postgres db
- Run operator locally and apply 
```
kind: ServiceBindingRequest
metadata:
  name: binding-request
spec:
  applicationSelector:
    resourceRef: java-app
    group: apps
    version: v1
    resource: deployments
  backingServiceSelector:
    group: postgresql.baiju.dev
    version: v1alpha1
    kind: Database
    resourceRef: db-demo
  customEnvVar:
    - name: JDBC_URL
      value: 'jdbc:postgresql://{{ .status.dbConnectionIP }}:{{ .status.dbConnectionPort }}/{{ .status.dbName }}'   # yamllint disable-line rule:line-length
    - name: DB_USER
      value: '{{ index .status.dbConfigMap "db.username" }}'
    - name: DB_PASSWORD
      value: '{{ index .status.dbConfigMap "db.password" }}'
```

For more detailed guidelines please follow https://github.com/redhat-developer/service-binding-operator/tree/master/examples/java_postgresql_customvar

PR #353, opened 2020-02-28
Associate binding secret to a custom schema path in the workload

sbose78:
### Motivation

- Many Application workloads do not support podSpec, and might need association of the binding secret in an arbitrary location.
- Since we are really concerned about `containers` and `volumes`, users should be able to provide a custom path to them as well.

Example, in the below workload, the binding secret has to be associated in 
`spec.eventSourceSecret` instead of `spec.template.spec.containers`

```
apiVersion: sources.eventing.knative.dev/v1alpha1
 kind: KafkaSource
 metadata:
  name: my-kafka-source
 spec:
  consumerGroup: knative-group
  eventSourceSecret: eventsource-binding-foo-bar-binding-secret <--- Association made by service binding operator
  sink:
    ref:
      apiVersion: serving.knative.dev/v1alpha1
      kind: Service
      name: event-display
```

To accomplish the same, the `ServiceBindingRequest`'s  `applicationSelector` could look like the following:

```
apiVersion: apps.openshift.io/v1alpha1
kind: ServiceBindingRequest

metadata: 
  name: eventsource-binding-foo-bar
  namespace: service-binding-demo

spec: 
  applications: 
    - bindingPath:
         customSecretPath: spec.eventSourceSecret # <--- In  KafkaSource
      group: sources.eventing.knative.dev
      resource: kafkasources
      resourceRef: my-kafka-source
      version: v1alpha1

  services:                                                                                                         
    - group: kafka.strimzi.io
      version: v1beta1
      kind: KafkaTopic
      resourceRef: my-kafka-topic 
    - group: kafka.strimzi.io
      version: v1beta1
      kind: Kafka
      resourceRef: my-kafka-cluster
```

### Changes

New non-mandatory API field `bindingPath` introduced.

```
applications:
  - bindingPath:
       podSpecPath:
         containers: spec.template.spec.containers
         volumes: spec.template.spec.volumes
       customSecretPath: spec.custom.path
     group: apps.openshift.io
     resource: deploymentconfigs
     resourceRef: nodejs-rest-http-crud
     version: v1
```

**Default behaviour**
If `bindingPath` is not set, default containers/volumes path is used as per podSpec.

### Testing
Ensured that binding is successful with/without podSpec path set.

PR #366, opened 2020-03-07
WIP Optionally support "binding configmap" instead of "binding secret"

sbose78:
_Feature request from the @matzew._ , @lberk  ( knative eventing )

**What**

Currently, the binding information is written to a `secret` and then injected into a workload.

This PR proposes that the binding information be optionally written into a `configmap`, and then the `configmap` be injected into the workload

```
  envFrom:
     - configMapRef:
         name: binding-data
```

**Why?**

> fields like "topics", "consumerGroup" and "bootstrapServers" - none of them really needs to be a secretly protected thing.
> 
> mounting in their values via vanilla configmap seems more reasonable, instead of doing it via a `secret`, since the names of the topics, for instance, do not usually need to be protected

https://issues.redhat.com/browse/SRVKE-257

**How**

Introduced the `spec.binding` field which asks the user to optionally provide the GVK of the object into which the binding data would be written to.

```
kind: ServiceBindingRequest
metadata:
  name: binding-request-3
  namespace: sbose
spec:
  applications:
    -  group: apps.openshift.io
       resource: deploymentconfigs
       resourceRef: nodejs-rest-http-crud-3
       version: v1
  services:
    - group: postgresql.baiju.dev
       kind: Database
      resourceRef: demo-database
      version: v1alpha1
  binding:
     objectType:
        group: ''
        kind: ConfigMap
        version: v1
  detectBindingResources: false
```

Defaults:
If `bindingRef == nil`, create a binding secret


sbose78:
@otaviof Good point, we've also got https://github.com/redhat-developer/service-binding-operator/pull/353 .

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -303,31 +304,67 @@ func (b *Binder) appendEnvVar(
...
+	dataType := b.sbr.Spec.BindingReferenceType.Kind
+
 	for _, env := range envList {
-		if env.SecretRef.Name == secret {
-			b.logger.Debug("Directive 'envFrom' is already present!")
-			// secret name is already referenced
-			return envList
+		if dataType == ConfigMapKind {
+			if env.ConfigMapRef.Name == binddata {
+				b.logger.Debug("Directive 'envFrom' is already present!")
```
Avni-Sharma: We should log whether envFrom here is for `ConfigMap` or `secret`

sbose78: Done, thank you!

sbose78:
```
binding:
     objectType:
        group: ''
        kind: ConfigMap
        version: v1
     paths: < will add in a different PR >
```

@otaviof , this is how I have it now.

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -303,31 +304,67 @@ func (b *Binder) appendEnvVar(
 
 // appendEnvFrom based on secret name and list of EnvFromSource instances, making sure secret is
 // part of the list or appended.
-func (b *Binder) appendEnvFrom(envList []corev1.EnvFromSource, secret string) []corev1.EnvFromSource {
+func (b *Binder) appendEnvFrom(envList []corev1.EnvFromSource, binddata string) []corev1.EnvFromSource {
```
Avni-Sharma: we should follow camelCase for `bindData`

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -303,31 +304,67 @@ func (b *Binder) appendEnvVar(
...
+		}
+	}
+
+	b.logger.Debug("Adding 'envFrom' directive...")
+	return append(envList, dataRef)
 }
 
 // removeEnvFrom remove bind related entry from slice of "EnvFromSource".
-func (b *Binder) removeEnvFrom(envList []corev1.EnvFromSource, secret string) []corev1.EnvFromSource {
+func (b *Binder) removeEnvFrom(envList []corev1.EnvFromSource, data string) []corev1.EnvFromSource {
```
Avni-Sharma: Shouldn't this `data` be `bindData` as well(just to maintain consistency :))

pkg/controller/servicebindingrequest/secret.go
```diff
@@ -52,21 +54,61 @@ func (s *Secret) createOrUpdate(payload map[string][]byte) (*unstructured.Unstru
...
 	}
 
-	gvk := corev1.SchemeGroupVersion.WithKind(SecretKind)
-	u, err := converter.ToUnstructuredAsGVK(secretObj, gvk)
+	objectMeta := metav1.ObjectMeta{
+		Namespace: ns,
+		Name:      name,
+	}
+
+	/*
```
Avni-Sharma: Do we need this commented portion?

pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -162,6 +164,71 @@ func TestBinderNew(t *testing.T) {
...
+		f.FakeClient(),
+		f.FakeDynClient(),
+		sbr,
+		[]string{},
+	)
+
+	require.NotNil(t, binder)
+
+	t.Run("appendEnvFrom-removeEnvFrom", func(t *testing.T) {
+		secretName := "secret"
```
Avni-Sharma: since the function is for testing binder config map, here `secretName` should be `configMapName`

pkg/controller/servicebindingrequest/binding.go
```diff
@@ -220,6 +220,15 @@ func (b *ServiceBinder) onError(
...
+		b.Secret.plan.SBR.Spec.BindingReference = &v1alpha1.BindingReference{
+			ObjectType: v1.GroupVersionKind{
+				Kind:    SecretKind,
+				Version: "v1",
+			},
+		}
+	}
+
 	b.Logger.Info("Saving data on intermediary secret...")
 	secretObj, err := b.Secret.Commit(b.Data)
```
Avni-Sharma: We can track in a different issue to show the configmap in the status when intermediate secret is not being used. Currently, we have info for binding secret

pkg/controller/servicebindingrequest/secret.go
```diff
@@ -52,21 +54,61 @@ func (s *Secret) createOrUpdate(payload map[string][]byte) (*unstructured.Unstru
...
-			Namespace: ns,
-			Name:      name,
-		},
-		Data: payload,
+
+	var bindingRef interface{}
+	resource := SecretResource
+
+	if s.plan.SBR.Spec.BindingReference == nil {
+		return nil, err.New("BindingReference is empty")
```
Avni-Sharma: If ` bindingRef == nil,` a binding secret should be created , `resource = SecretResource` . It should not pass errors, IMO

pkg/controller/servicebindingrequest/secret.go
sbose78: I was trying to announce what I was trying to do, and what the issue I was seeing with that :)

pkg/controller/servicebindingrequest/binder_test.go
sbose78: Thank you! Done https://github.com/redhat-developer/service-binding-operator/pull/366/commits/c8566ef6c187e26e9e987ffe184925e1e10b4fe0#diff-d422caad0c9d74644056915cb438c96a

pkg/controller/servicebindingrequest/binding.go
sbose78: Very good point.

sbose78: Here you go https://github.com/redhat-developer/service-binding-operator/pull/366/commits/1c07fe2bde943393054d1c76c4119a2fc5687b15 . It would be unfair to not include the changes in this PR, I think.

pkg/controller/servicebindingrequest/secret.go
sbose78: Done, added a test too https://github.com/redhat-developer/service-binding-operator/pull/366/commits/c8566ef6c187e26e9e987ffe184925e1e10b4fe0

pkg/controller/servicebindingrequest/binder.go
sbose78: Done!

pkg/controller/servicebindingrequest/binder.go
sbose78: Right , done :)

pkg/controller/servicebindingrequest/bindingdata.go
```diff
@@ -0,0 +1,162 @@
...
+// Returns an error when the Kubernetes client returns an error.
+func (s *BindingDataHandler) createOrUpdate(payload map[string][]byte) (*unstructured.Unstructured, error) {
+	ns := s.plan.Ns
+	name := s.plan.Name
+	logger := s.logger.WithValues("Namespace", ns, "Name", name)
+
+	var bindingRef interface{}
+
+	// Default binding resource type is Secret
+	resource := SecretResource
```
Avni-Sharma: this assigning `resource := SecretResource` should be after the `s.plan.SBR.Spec.BindingReference.ObjectType.Kind == ConfigMapKind` check in an `else`. So currently everytime secret is being created even before checking.

sbose78: Yeah, the flow is more like "Start with secret, and continue with secret unless specified otherwise"

1) Assume, we need to have a `Secret`

```
// Default binding resource type is Secret
resource := SecretResource
gvk := corev1.SchemeGroupVersion.WithKind(SecretKind)
```

2) If something is explicitly specified, and that something is a `configmap`, we'll go for configmap. Therefore, if someone specifies ObjectType.Kind = Service, we'll ignore it and create a binding secret.
```
// Only if otherwise specified as configmap, we'll use a binding
// configmaps
if s.plan.SBR.Spec.Binding != nil &&
	s.plan.SBR.Spec.Binding.ObjectType.Kind == ConfigMapKind {
	gvk = corev1.SchemeGroupVersion.WithKind(s.plan.SBR.Spec.Binding.ObjectType.Kind)
		resource = ConfigMapResource
}
```

sbose78:
https://github.com/redhat-developer/service-binding-operator/pull/366/commits/a54f4ce440920750130e097dbacb338517999880 is not related to my PR but I had to do it because `make lint` seems to have discovered **just now** that it is unused code.

pkg/controller/servicebindingrequest/binding.go
```diff
@@ -220,13 +221,31 @@ func (b *ServiceBinder) onError(
...
 		return b.onError(err, b.SBR, sbrStatus, nil)
 	}
-	sbrStatus.Secret = secretObj.GetName()
+	sbrStatus.BindingData = v1alpha1.BindingData{
+		GroupVersionKind: metav1.GroupVersionKind{
+			Group:   bindingData.GroupVersionKind().Group,
+			Version: bindingData.GroupVersionKind().Version,
+			Kind:    bindingData.GroupVersionKind().Kind,
+		},
+		LocalObjectReference: v1.LocalObjectReference{
```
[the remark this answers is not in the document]
sbose78: I did consider that :) but it uses `API Group` & `Kind` instead of `GVK`. Are you okay with that?

pkg/controller/servicebindingrequest/binding.go
```diff
@@ -220,13 +221,31 @@ func (b *ServiceBinder) onError(
 func (b *ServiceBinder) Bind() (reconcile.Result, error) {
 	sbrStatus := b.SBR.Status.DeepCopy()
 
+	if b.bindingDataHandler.plan.SBR.Spec.Binding == nil {
```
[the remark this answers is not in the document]
sbose78: I've actually removed this check since due diligence is being done in `BindingDataHandler.createOrUpdate` already. e641143

pkg/controller/servicebindingrequest/binding.go
```diff
@@ -7,7 +7,8 @@ import (
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	"gotest.tools/assert/cmp"
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
sbose78: ef703f8a29d3a92818030df6ae0de88e7bea62c1

sbose78:
@Avni-Sharma would you please give these a try locally and let me know if tests are failing for you? For me, they aren't.

Avni-Sharma:
> @Avni-Sharma would you please give these a try locally and let me know if tests are failing for you? For me, they aren't.

It shows 
```
pkg/controller/servicebindingrequest/reconciler_test.go:400:53: sbrOutput.Status.Secret undefined (type v1alpha1.ServiceBindingRequestStatus has no field or method Secret) (typecheck)
		require.Equal(t, reconcilerName, sbrOutput.Status.Secret)
```

isutton:
> I did consider that :) but it uses `API Group` & `Kind` instead of `GVK`. Are you okay with that? 

I'm ok with that.

From the user's perspective, it might be shorter and convey the same information; for example:

```yaml
- apiGroup: apps/v1
  kind: Deployment
```

vs

```yaml
- group: apps
  version: v1
  kind: Deployment
```

From the technical perspective, it might be an advantage to use the primitives Kubernetes provides, since the value would not require conversions from and to an internal representation to be used by its APIs.

Below is an excerpt from the [Kubernetes role binding example](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-example)  with the same type apparently being used:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
# This role binding allows "jane" to read pods in the "default" namespace.
# You need to already have a Role named "pod-reader" in that namespace.
kind: RoleBinding
metadata:
  name: read-pods
  namespace: default
subjects:
# You can specify more than one "subject"
- kind: User
  name: jane # "name" is case sensitive
  apiGroup: rbac.authorization.k8s.io
roleRef:
  # "roleRef" specifies the binding to a Role / ClusterRole
  kind: Role #this must be Role or ClusterRole
  name: pod-reader # this must match the name of the Role or ClusterRole you wish to bind to
  apiGroup: rbac.authorization.k8s.io
```

sbose78:
From a technical perspective, isn't metav1.GroupVersionKind also a Kubernetes primitive type?

Avni-Sharma:
@sbose78  The unit test is not starting at all for quite few days now. The unit tests are passing locally for me. 
/lgtm

Avni-Sharma:
> From a technical perspective, isn't metav1.GroupVersionKind also a Kubernetes primitive type?

Yes

isutton:
It is, but we’re implementing our own version of a namespaces local
resource, so it makes sense to use K8s instead.

sbose78:
Hi @isutton , 
I gave it a shot, is this what you are looking for https://github.com/sbose78/service-binding-operator/pull/4/files ?

sbose78:
wrong, need TypeMeta..

isutton:
>
> Hi @isutton <https://github.com/isutton> ,
> I gave it a shot, is this what you are looking for
> https://github.com/sbose78/service-binding-operator/pull/4/files ?
>
>  Yes, that's the direction.

What do you think about removing the related BindingData type and using
TypedLocalObjectReference directly, since the former's sole responsibility
is to encapsulate the latter?

sbose78:
That sounds good!

sbose78:
Closing this as I wouldn't get a chance to take a look at this in the next couple of weeks. Feel free to re-open this and push to this branch, or create a fresh one.

PR #371, opened 2020-03-09
(bug) Fix normalization of annotations to CRDDescription

sbose78:
### Motivation

Parsing the annotations into the data structures which contained spec|status descriptors had a bug where only one spec descriptor and one status descriptor was being persisted irrespective of the total number of annotations present.

Example, the following 
```
"servicebindingoperator.redhat.io/status.dbCredentials-db.password": "binding:env:object:secret",
"servicebindingoperator.redhat.io/spec.dbName":                      "binding:env:attribute",
"servicebindingoperator.redhat.io/spec.image":                       "binding:env:attribute",
"servicebindingoperator.redhat.io/status.dbConfigMap-db.host":       "binding:env:object:configmap",
```
was being mapped to only one spec descriptor and one status descriptor, instead of two each.

### Changes

Fix the 

```
func buildDescriptorsFromAnnotations(annotations map[string]string) (
	[]olmv1alpha1.SpecDescriptor,
	[]olmv1alpha1.StatusDescriptor,
	error, 
``` 
function to append parsed annotations into a list. 

### Testing

Ideally, use of a backing service which has no CSV should validate this.

- [x] Added tests
- [x] The fix was put in https://github.com/redhat-developer/service-binding-operator/pull/367 so that we could test use of backing service objects which had no CSV/CRD. The change has been validated there.

### DIFF
--- a/pkg/controller/servicebindingrequest/olm.go
+++ b/pkg/controller/servicebindingrequest/olm.go
@@ -198,11 +198,11 @@ func buildCRDDescriptionFromCRD(crd *unstructured.Unstructured) (*olmv1alpha1.CR
 	}
 
 	if specDescriptor != nil {
-		crdDescription.SpecDescriptors = append(crdDescription.SpecDescriptors, *specDescriptor)
+		crdDescription.SpecDescriptors = append(crdDescription.SpecDescriptors, specDescriptor...)
 	}
 
 	if statusDescriptor != nil {
-		crdDescription.StatusDescriptors = append(crdDescription.StatusDescriptors, *statusDescriptor)
+		crdDescription.StatusDescriptors = append(crdDescription.StatusDescriptors, statusDescriptor...)
 	}
 
 	return crdDescription, nil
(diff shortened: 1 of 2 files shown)


pkg/controller/servicebindingrequest/olm.go
```diff
@@ -198,11 +198,11 @@ func buildCRDDescriptionFromCRD(crd *unstructured.Unstructured) (*olmv1alpha1.CR
 	}
 
 	if specDescriptor != nil {
-		crdDescription.SpecDescriptors = append(crdDescription.SpecDescriptors, *specDescriptor)
+		crdDescription.SpecDescriptors = append(crdDescription.SpecDescriptors, specDescriptor...)
```
pedjak: how about to rename the arg name to plural form, i.e. `specDescriptors`?

sbose78: Good point, done https://github.com/redhat-developer/service-binding-operator/pull/371/commits/e75799f9324edc0fd8a02788c6397186c37c0a1f

pkg/controller/servicebindingrequest/olm_test.go
```diff
@@ -75,3 +76,55 @@ func TestOLMNew(t *testing.T) {
...
+		require.Len(t, crdDescription.SpecDescriptors, 2)
+
+		expected := map[string]string{
+			"dbName":        "binding:env:attribute:spec.dbName",
+			"image":         "binding:env:attribute:spec.image",
+			"dbCredentials": "binding:env:object:secret:db.password",
+			"dbConfigMap":   "binding:env:object:configmap:db.host",
+		}
+
+		for _, value := range crdDescription.StatusDescriptors {
```
Avni-Sharma: One loop can be added for SpecDescriptors as well

sbose78: Done, thank you for the close review 87eff7b

PR #370, opened 2020-03-09
Validate SBO releases

baijum:
Ref. https://issues.redhat.com/browse/APPSVC-470

### Motivation

After making SBO release through operatorhub.io, validate these releases.

PR #378, opened 2020-03-12
reverted version back to 0.1.0

pedjak:
It was increased by mistake in PR #370


sbose78:
Does quay applications allow going to an older semantic version?

pedjak:
If understand correctly @sbose78 0.1.1 is not published to quay

DhritiShikhar:
@pedjak I can see `0.1.1-256` and `0.1.1-257` here https://quay.io/application/redhat-developer/service-binding-operator?tab=releases

navidsh:
@pedjak I'm not sure if my issue is relevant to the version change, but I'm not seeing any extra info in the CSV for `0.1.1-257`:

image

I'll check again once `0.1.0-258` shows up in OperatorHub.

baijum:
> It was increased by mistake in PR #370

@pedjak That was an intentional change to make the dev releases from for every commit to master branch. This is required to make the dev release number higher than the current official release.
Can you let me if I missed something?

pedjak:
@baijum we did not release any `0.1.0` yet, and all dev releases have suffix. Why does the dev release number need to be higher?

baijum:
@pedjak Dev releases are pre-releases as per the SemVer spec:
https://semver.org/#spec-item-9
(A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version.)

After making a formal release, we should increment the number in Makefile so that the Dev releases will become pre-releases.

Issue #373, opened 2020-03-11
API rule violations 


sbose78:
```
 operator-sdk generate k8s && operator-sdk generate openapi
```

gives us

```
API rule violation: list_type_missing,./pkg/apis/apps/v1alpha1,ServiceBindingRequestSpec,CustomEnvVar
API rule violation: list_type_missing,./pkg/apis/apps/v1alpha1,ServiceBindingRequestStatus,ApplicationObjects
API rule violation: list_type_missing,./pkg/apis/apps/v1alpha1,ServiceBindingRequestStatus,Conditions
API rule violation: names_match,./pkg/apis/apps/v1alpha1,ServiceBindingRequestStatus,ApplicationObjects
```

sbose78:
https://github.com/kubernetes/kube-openapi/blob/master/pkg/generators/rules/names_match.go#L55

Avni-Sharma:
Ran the command on `master` . It works fine with the following as output
```
INFO[0000] Running deepcopy code-generation for Custom Resource group versions: [apps:[v1alpha1], ] 
INFO[0007] Code-generation complete.                    
[Deprecation notice] The 'operator-sdk generate openapi' command is deprecated!

 - To generate CRDs, use 'operator-sdk generate crds'.
 - To generate Go OpenAPI code, use 'openapi-gen'. For example:

      # Build the latest openapi-gen from source
      which ./bin/openapi-gen > /dev/null || go build -o ./bin/openapi-gen k8s.io/kube-openapi/cmd/openapi-gen

      # Run openapi-gen for each of your API group/version packages
      ./bin/openapi-gen --logtostderr=true -o "" -i ./pkg/apis/<group>/<version> -O zz_generated.openapi -p ./pkg/apis/<group>/<version> -h ./hack/boilerplate.go.txt -r "-"

INFO[0000] Running OpenAPI code-generation for Custom Resource group versions: [apps:[v1alpha1], ] 
API rule violation: list_type_missing,github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1,ServiceBindingRequestSpec,CustomEnvVar
API rule violation: list_type_missing,github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1,ServiceBindingRequestStatus,ApplicationObjects
API rule violation: list_type_missing,github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1,ServiceBindingRequestStatus,Conditions
[...]

PR #379, opened 2020-03-12
Fix API rule violations

Avni-Sharma:
Fix issue #373


deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
```diff
@@ -259,7 +259,7 @@ spec:
         status:
           description: ServiceBindingRequestStatus defines the observed state of ServiceBindingRequest
           properties:
-            applications:
+            applicationObjects:
```
sbose78: How about flipping this, call everything `applications`?

Avni-Sharma:
Currently, we have 
```
INFO[0000] Running OpenAPI code-generation for Custom Resource group versions: [apps:[v1alpha1], ] 
API rule violation: list_type_missing,github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1,ServiceBindingRequestStatus,Applications
API rule violation: list_type_missing,github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1,ServiceBindingRequestStatus,Conditions
```

To solve this we will have to do
```
Conditions *[]conditionsv1.Condition `json:"conditions"`

```

However, Ideally, we should not have `*[]conditionsv1` . I believe we can ignore such violations because, on a quick check with upstream operators, I found that `[]conditionsv1` is being used.

sbose78:
@Avni-Sharma Can I take this in?

Avni-Sharma:
> @Avni-Sharma Can I take this in?

Yes

PR #90, opened 2019-07-29
[ODC-1329] Kinds Supported for ApplicationSelector

otaviof:
Using a more generic way of handling application resources. Adding back support to:

- DeploymentConfigs;
- Deployments;
- StatefulSets;
- DaemonSets;

### Changes in client-go's Fixture

In order to simulate the usage of `UnstructuredList`, as it would work in a real cluster, the following changes have been applied, they are present in this PR's vendor folder for the time being.

- https://github.com/kubernetes/kubernetes/pull/80723
- https://github.com/kubernetes/kubernetes/issues/80722
- https://github.com/kubernetes/kubernetes/issues/80609

PR #382, opened 2020-03-13
Remove osappsv1 from scheme

Avni-Sharma:
Remove osappsv1 from the scheme and as a dependency as now we use Unstructured and osappsv1 was added for deploymentconfigs. (https://github.com/redhat-developer/service-binding-operator/pull/90/files#diff-a8480c104199c9b04be874f1d87ef9abR44)

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -7,8 +7,7 @@ import (
 	"os"
 	"runtime"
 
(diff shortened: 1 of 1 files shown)


baijum:
@Avni-Sharma Can we remove the dependency from `go.mod` & `go.sum` files also?

Avni-Sharma:
openshift/api is being used in fake.go, mocks.go and bind_custom_resources_test.go, hence it should not be removed.

PR #386, opened 2020-03-17
update go version

Avni-Sharma:
update go version to 1.14

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/redhat-developer/service-binding-operator
 
-go 1.13
+go 1.14
 
 require (
 	cloud.google.com/go v0.45.1 // indirect
--- a/openshift-ci/Dockerfile.tools
+++ b/openshift-ci/Dockerfile.tools
@@ -28,7 +28,7 @@ WORKDIR /tmp
 RUN mkdir -p $GOPATH/bin
 RUN mkdir -p /tmp/goroot
 
-RUN curl -Lo go1.13.7.linux-amd64.tar.gz https://dl.google.com/go/go1.13.7.linux-amd64.tar.gz && tar -C /tmp/goroot -xzf go1.13.7.linux-amd64.tar.gz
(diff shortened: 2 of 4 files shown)


sbose78:
Do we have a 1.14 builder?

Avni-Sharma:
> Do we have a 1.14 builder?
> https://github.com/redhat-developer/service-binding-operator/blob/master/Dockerfile.rhel#L4

Not as of now :/

sbose78:
:(
/hold

baijum:
@Avni-Sharma There is Go 1.14.1 version available.

Avni-Sharma:
https://github.com/openshift/release/tree/master/projects/origin-release 1.14 is not available yet :/

openshift-ci/Dockerfile.tools
```diff
@@ -28,7 +28,7 @@ WORKDIR /tmp
 RUN mkdir -p $GOPATH/bin
 RUN mkdir -p /tmp/goroot
 
-RUN curl -Lo go1.13.7.linux-amd64.tar.gz https://dl.google.com/go/go1.13.7.linux-amd64.tar.gz && tar -C /tmp/goroot -xzf go1.13.7.linux-amd64.tar.gz
+RUN curl -Lo go1.14.linux-amd64.tar.gz https://dl.google.com/go/go1.14.linux-amd64.tar.gz && tar -C /tmp/goroot -xzf go1.14.linux-amd64.tar.gz
```
baijum, line 31: Use Go 1.14.3 (latest bugfix release).
https://dl.google.com/go/go1.14.3.linux-amd64.tar.gz

pedjak:
should not we update `go.mod` as well?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/redhat-developer/service-binding-operator
 
-go 1.13
+go 1.14.3
```
pedjak, line 3: we should not be so precise here, 1.14 should work.

DhritiShikhar: @pedjak 

Hi, Just to understand, Is there a technical reason why we don't provide the patch number?

pedjak: this here is about go compiler version, not the bugfix version. It is signal to compiler what language features will be available:

https://blog.ksub.org/bytes/2019/09/15/go.mods-go-directive/

DhritiShikhar: Oh thanks for explanation!

PR #391, opened 2020-03-23
Include past releases in the bundle

### LINKED ISSUE #390: Failing e2e bundle validation tests
image

baijum:
Also specify the version in the subscription with Manual upgrade.

Fixes #390

The current issue is coming because we decrement the version of the Quay application bundle. Since we are using a periodic job there is a chance for false positives if we merge to master within short intervals.

These issues are addressed with these changes:

1. Include all past releases in the Quay application bundle (Without this only the latest will be available in the catalog)

2. Specify the start version of CSV and use Manual mode for install plan approval (This prevent automatic upgrade)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -462,10 +462,6 @@ push-bundle-quay:
 	$(Q)$(OUTPUT_DIR)/venv3/bin/operator-courier verify $(SBR_MANIFESTS)
 	$(Q)$(OUTPUT_DIR)./venv3/bin/operator-courier push $(SBR_MANIFESTS) redhat-developer service-binding-operator $(BUNDLE_VERSION) "$(QUAY_BUNDLE_TOKEN)"
 
-.PHONY: aa
-aa:
-	echo $(SBR_MANIFESTS)
-
 ## -- Target for validating the operator --
 .PHONY: dev-release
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -440,19 +443,29 @@ push-to-manifest-repo:
...
 	rm -rf deploy/olm-catalog/$(GO_PACKAGE_REPO_NAME)/$(BUNDLE_VERSION)
 
+## -- Target to push bundle to quay
+.PHONY: push-bundle-to-quay
+## Push manifest bundle to quay application
+push-bundle-quay:
+	$(Q)$(OUTPUT_DIR)/venv3/bin/operator-courier verify $(SBR_MANIFESTS)
+	$(Q)$(OUTPUT_DIR)./venv3/bin/operator-courier push $(SBR_MANIFESTS) redhat-developer service-binding-operator $(BUNDLE_VERSION) "$(QUAY_BUNDLE_TOKEN)"
+
+.PHONY: aa
```
Avni-Sharma: Is this target required? I guess this was just for echo

baijum: Removed. dd9e556bbdfd9f43792576f6e772e38417639487

PR #395, opened 2020-03-25
Increment version to 0.1.1

baijum:
This is required to fix Dev release.

"It looks like OperatorSource is always pulling the latest version from the Quay application registry. Since we can't remove releases from Quay, we should increment the version."


sbose78:
Thought so :)

/lgtm
/approve

Issue #396, opened 2020-03-25
Support multiple backing services in customEnvVar / dataMapping


sbose78:
JIRA story https://issues.redhat.com/browse/APPSVC-481

Extracted out https://github.com/redhat-developer/service-binding-operator/issues/356#issuecomment-596108001

```
backingServiceSelectors:
    - group: postgres.dev
      kind: Service
      resourceRef: user-db
      version: v1beta1
      id: X
    - group: postgres.dev
      kind: Credentials
      resourceRef: admin-creds
      version: v1beta1
      id: Y
customEnvVar:
    - name: CONNECTION_URL
       value: {{  X.status.username }} / {{ Y.status.url }}
```

sbose78:
We kinda decided on this, @ZhuangYuZY to make it easy for folks to write :) What do you think?

sbose78:
@otaviof did you want `value: {{  X.status.username }} / {{ Y.status.url }` to look different?

qibobo:
@sbose78 
To implement this, the retriever.cache will be changed from map[string]interface{} to map[string]map[string]interface{}?
For map[string]interface{}, the key is "status", or "spec" ....
For map[string]map[string]interface{}, the key are X, Y and the value are the same as current cache(map[string]interface{}).
Please correct me if I am wrong. Thanks.

otaviof:
@sbose78, @qibobo regarding the data collection, lets sync with @isutton and @Avni-Sharma, they are currently busy with those changes.

ZhuangYuZY:
@sbose78 Thank you. Will take some time for internal discussion to see if it works.

sbose78:
FYI @arthurdm @navidsh tracking this separately here.

ZhuangYuZY:
@sbose78 and @otaviof following our discussion on to-json requirement in #388 .

After discussion with team, we want to go with following way, which is using to-json in a template string as customEnvVar. I think it aligns with our discussion and operator design that extending customEnvVar to introduce a new to-json function, and also multiple backing services which identified by `id`.  Example is below. Let me know your comment. If agree, then let's discussion on implementation. 
There are three changes needed 1) multiple backing services no override on customEnvVar, 2) id for backing services, 3) to-json function based on id. My question is do you have any ongoing work for above ? I'd like understand and think how to collaborate to submit PR for 1), 2), and 3).
Let me know your comment. Thank you.

```
apiVersion: apps.openshift.io/v1alpha1
[...]

Issue #387, opened 2020-03-18
Feedback: support defining envVarPrefix per backing service


navidsh:
Currently, it is only possible to define one `envVarPrefix` in the `ServiceBindingRequest` CR. And if multiple backing services are defined, they all get the same prefix.
Being able to use different prefixes per backing service is important since different backing services might use the same name for the binding information.

https://github.com/redhat-developer/service-binding-operator/issues/356 has a similar discussion on how the SBR could look like to get around this.

ZhuangYuZY:
@sbose78 What I understand, we plan to do following to support envVarPrefix for multiple backing service CRs. Is it correct, thank you.

```
backingServiceSelectors:
    - group: serving.knative.dev
      kind: Service
      resourceRef: demo-1
      version: v1beta1
      envVarPrefix: demo-1-prefix
    - group: serving.knative.dev
      kind: Service
      resourceRef: demo-2
      version: v1beta1
      envVarPrefix: demo-2-prefix
```

sbose78:
@ZhuangYuZY , roughly yes.

ZhuangYuZY:
@sbose78 Do we have anyone working on this, I can get PR prepared. Thank you.

sbose78:
Go for it @ZhuangYuZY !

One thing to note, there might be a global env var prefix ( spec.envVarPrefix ) too. Maybe the combined prefix should be `spec.envVarPrefix` + "_" + `spec.backingServiceSelectors[i].envVarPrefix` ?

isutton:
>
> @sbose78 <https://github.com/sbose78> What I understand, we plan to do
> following to support envVarPrefix for multiple backing service CRs. Is it
> correct, thank you.
>
> backingServiceSelectors:
>     - group: serving.knative.dev
>       kind: Service
>       resourceRef: demo-1
>       version: v1beta1
>       envVarPrefix: demo-1-prefix
>     - group: serving.knative.dev
>       kind: Service
>       resourceRef: demo-2
>       version: v1beta1
>       envVarPrefix: demo-2-prefix
>
> Ideally we should strive to require the interaction of the developer as
little as possible by offering good enough defaults. In the case of
multiple services, using at least *kind* and *resourceRef* is a wonderful
start.

Building on your example, I see that each configuration value exposed in
those resources would have the following prefix (already munged to look
like an environment variable): SERVICE_DEMO-1_ and SERVICE_DEMO-2_ .

[...]

PR #410, opened 2020-04-02
Add "id" and "envVarPrefix" to backingServiceSelector to enable multiple service binding

qibobo:
### Motivation

1 Current when there are multiple services in backingServiceSelectors, the last service's credential will override all the previous' which need to handle.
2 Sometimes we need a json format custom env var.

### Changes

1 Change retriever.cache from map[string]interface{} to map[string]map[string]interface{} and each backingService has an id to identify itself in the cache.
2 Add a function to_json into custom_env_parser to do the json marshal.

More details can be found in 
https://github.com/redhat-developer/service-binding-operator/issues/396

### Testing

Add the necessary unit tests for it.

```
kind: ServiceBindingRequest
metadata:
  name: helloworld-binding
spec:
  applicationSelector:
    resourceRef: helloworld
    group: serving.knative.dev
    version: v1
    resource: services
  backingServiceSelectors:
    - group: ibmcloud.ibm.com
      version: v1alpha1
      kind: Binding
      resourceRef: coligo-cos-service-credential
      id: coligo-cos-service-credential
  customEnvVar:
    - name: VCAP_SERVICE
      value: {
          "cloud-object-storage": [
            {
              "credentials": {
                {{to_json coligo-cos-service-credential.status.secretName}} 
              },
              "name": "coligo-cos",
              "plan": "standard"                      
            }
          ]
        }
```

### DIFF
--- a/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
+++ b/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
@@ -82,8 +82,9 @@ type BackingServiceSelector struct {
 	// customEnvVar:
 	// 	- name: CONNECTION_URL
 	// 	value: {{  X.status.username }} / {{ Y.status.url }}
-	Id        string  `json:"id,omitempty"`
-	Namespace *string `json:"namespace,omitempty"`
+	Id           string  `json:"id,omitempty"`
+	Namespace    *string `json:"namespace,omitempty"`
+	EnvVarPrefix string  `json:"envVarPrefix,omitempty"`
 }
 
 // BoundApplication defines the application workloads to which the binding secret has
--- a/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
+++ b/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
@@ -66,6 +66,22 @@ type BackingServiceSelector struct {
 	metav1.GroupVersionKind `json:",inline"`
 	ResourceRef             string `json:"resourceRef"`
 	// +optional
+	// The id is provided by user to identify the backing service in the custom env var tempalte.
+	// Below is an example:
+	// backingServiceSelectors:
+	//   - group: postgres.dev
+	// 	   kind: Service
+	// 	   resourceRef: user-db
+	// 	   version: v1beta1
+	// 	   id: X
+	//   - group: postgres.dev
+	// 	   kind: Credentials
+	// 	   resourceRef: admin-creds
+	// 	   version: v1beta1
+	// 	   id: Y
+	// customEnvVar:
+	// 	- name: CONNECTION_URL
+	// 	value: {{  X.status.username }} / {{ Y.status.url }}
 	Id        string  `json:"id,omitempty"`
 	Namespace *string `json:"namespace,omitempty"`
 }
--- a/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
+++ b/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
@@ -66,6 +66,7 @@ type BackingServiceSelector struct {
 	metav1.GroupVersionKind `json:",inline"`
 	ResourceRef             string `json:"resourceRef"`
 	// +optional
+	Id        string  `json:"id,omitempty"`
 	Namespace *string `json:"namespace,omitempty"`
 }
 
--- a/pkg/controller/servicebindingrequest/custom_env_parser.go
+++ b/pkg/controller/servicebindingrequest/custom_env_parser.go
@@ -11,11 +11,11 @@ import (
 // CustomEnvParser is responsible to interpolate a given EnvVar containing templates.
 type CustomEnvParser struct {
 	EnvMap []corev1.EnvVar
-	Cache  map[string]interface{}
+	Cache  map[string]map[string]interface{}
 }
 
 // NewCustomEnvParser returns a new CustomEnvParser.
-func NewCustomEnvParser(envMap []corev1.EnvVar, cache map[string]interface{}) *CustomEnvParser {
+func NewCustomEnvParser(envMap []corev1.EnvVar, cache map[string]map[string]interface{}) *CustomEnvParser {
 	return &CustomEnvParser{
 		EnvMap: envMap,
 		Cache:  cache,
--- a/pkg/controller/servicebindingrequest/retriever_test.go
+++ b/pkg/controller/servicebindingrequest/retriever_test.go
@@ -9,7 +9,7 @@ import (
 	"github.com/redhat-developer/service-binding-operator/test/mocks"
 )
 
-func TestRetriever(t *testing.T) {
+func TestRetrieverWithDifferentCrIds(t *testing.T) {
 	logf.SetLogger(logf.ZapLogger(true))
 	var retriever *Retriever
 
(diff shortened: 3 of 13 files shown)


pkg/controller/servicebindingrequest/custom_env_parser.go
```diff
@@ -42,3 +43,10 @@ func (c *CustomEnvParser) Parse() (map[string]interface{}, error) {
 	}
 	return data, nil
 }
+func MarshalToJSON(m interface{}) (string, error) {
```
Avni-Sharma, line 46: Nit pick -This function is not an exported function. We can have it as `marshalToJson`  to maintain consistency

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -67,6 +67,7 @@ type BackingServiceSelector struct {
 	metav1.GroupVersionKind `json:",inline"`
 	ResourceRef             string `json:"resourceRef"`
 	// +optional
+	Id        string  `json:"id,omitempty"`
```
Avni-Sharma, line 70: Can we have a short description for the optional `id` field here as it has been provided for other optional fields as well.

isutton:
>
> In pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
> <https://github.com/redhat-developer/service-binding-operator/pull/410#discussion_r402928860>
> :
>
> > @@ -67,6 +67,7 @@ type BackingServiceSelector struct {
>  	metav1.GroupVersionKind `json:",inline"`
>  	ResourceRef             string `json:"resourceRef"`
>  	// +optional
> +	Id        string  `json:"id,omitempty"`
>
> Can we have a short description for the optional id field here as it has
> been provided for other optional fields as well.
>
> Shall we leave this out from this PR? I don't believe the JSON
functionality depends on this change.

pkg/controller/servicebindingrequest/custom_env_parser.go
```diff
@@ -25,7 +26,7 @@ func NewCustomEnvParser(envMap []corev1.EnvVar, cache map[string]interface{}) *C
 func (c *CustomEnvParser) Parse() (map[string]interface{}, error) {
 	data := make(map[string]interface{})
 	for _, v := range c.EnvMap {
-		tmpl, err := template.New("set").Parse(v.Value)
+		tmpl, err := template.New("set").Funcs(template.FuncMap{"to_json": MarshalToJSON}).Parse(v.Value)
```
isutton, line 29: I believe `toJson` or just `json` would be more idiomatic.

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
isutton: I'd prefer to have this change in a different PR.

qibobo:
@isutton @sbose78 
As you suggested, I rename this PR and have submitted another PR to do the json feature.
https://github.com/redhat-developer/service-binding-operator/pull/412

pkg/controller/servicebindingrequest/retriever_test.go
```diff
@@ -16,6 +16,8 @@ func TestRetriever(t *testing.T) {
 	ns := "testing"
 	backingServiceNs := "backing-servicec-ns"
 	crName := "db-testing"
+	crId1 := "testingCrId1"
+	crId2 := "testingCrId2"
```
DhritiShikhar, line 20: There should be tests for these scenarios as well:

[1] What happens when id is an empty string `""`
[2] What happens when id is not unique like:
```
crId1 = "testingCrId1"
crId2 = "testingCrId1"
```

sbose78:
@qibobo See https://github.com/redhat-developer/service-binding-operator/issues/387#issuecomment-610069526 . I think it's reasonable to have an `envVarPrefix`.

Could you refactor your PR to have an `envVarPrefix` instead of an `id`? The `id` is more of a way to refer to a backing service in `customEnvVar` which is a separate discussion, and there might be other ways to accomplish the same :)

qibobo:
@sbose78 
Per my understanding, current if we just add an envVarPrefix to each of the service in service selectors, only the last one will take affect because it will override all the previous ones, right?

sbose78:
If "service 1" has an env var called "MASTER_DB" for example, then all environment variables will be prefixed with that.

If "service 2" has an env var called "AGENT_DB" for example, then all environment variables will be prefixed with that.

I ideally wouldn't mix up `id` and `envVarPrefix` because the latter is a very explicit ask to have environment variables created in a certain way whereas `id` doesn't really say much about how the environment would look eventually. `id` at the most is helpful while referring to the service in the customEnvVar whereas `envVarPrefix` would be consumed with annotation-based environment variables too.

ZhuangYuZY:
@sbose78,  Now @qibobo is preparing two PRs. This is the first one to support `id` as backing service reference. The second one is in local for `envPrefix` pre backing service. 
To get these PRs ongoing, I think we need to get an agreement on design. I put a comment about my thinking on https://github.com/redhat-developer/service-binding-operator/issues/387#issuecomment-611319646
Please comment with your thought. Thank you.

sbose78:
Sounds good!

pkg/controller/servicebindingrequest/retriever_test.go
qibobo: fixed

qibobo:
@sbose78 According to our talk, so can we continue this PR?

sbose78:
Please go ahead with the PR for `envVarPrefix` @qibobo .

qibobo:
@sbose78 added the envVarPrefix changes.

sbose78:
Could you move `id` to a different PR please ?

qibobo:
> Could you move `id` to a different PR please ?

@sbose78 


1.  According to the comment https://github.com/redhat-developer/service-binding-operator/issues/387#issuecomment-614209441, so we get the conclusion that both "id" and "envVarPerfix per service" are necessary, am I right? I think the two changes have dependency on each other, so now I combine them in one PR. And I can divide it into 2 ones. Then we can merge one first and rebase the other. So which one do you think should be submitted first? The "id" one or the "envVarPerfix" one?

2. Sometimes the e2e tests fails easily in travis, but seems I can not get why does it fail? Can I run the e2e test in my local?

sbose78:
Apologies if it wasn't clear.

* `id` was being proposed to refer to backing services inside a customEnvVar template.
* service-specific `envVarPrefix` is used to add prefixes to environment variables.

baijum:
> 2. Sometimes the e2e tests fails easily in travis, but seems I can not get why does it fail? Can I run the e2e test in my local?

To run e2e tests locally, create an OpenShift cluster and set KUBECONFIG, then run:
```
make test-e2e
```

qibobo:
@sbose78 
Create a new PR for envVarPrefix https://github.com/redhat-developer/service-binding-operator/pull/428

Avni-Sharma:
@isutton do we need to close this PR as a similar one is being captured here https://github.com/redhat-developer/service-binding-operator/pull/468

isutton:
> @isutton do we need to close this PR as a similar one is being captured here #468

@Avni-Sharma thanks for point this out!

@qibobo can we close this PR in favor of #468?

qibobo:
@isutton 
I will close it. Thank you.

PR #412, opened 2020-04-03
Introduce "json" function in templating to generate inline "json" as custom env var

qibobo:
### Motivation

Sometimes we need a json format custom env var

### Changes

Add a function json into custom_env_parser to do the json marshal.


sbose78:
Nice @qibobo !

PR #428, opened 2020-04-18
Add an envVarPrefix field to backingServiceSelector

qibobo:
### Motivation

According to issue https://github.com/redhat-developer/service-binding-operator/issues/387, an envVarPrefix will be added to backingServiceSelector to support multiple service binding as below shows:

  ```
backingServiceSelectors:
    - group: groupname
      version: v1alpha1
      kind: Binding
      resourceRef: resourceName
      envVarPrefix: testprefix
```

### Changes

Add a field envVarPrefix.

### DIFF
--- a/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
+++ b/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
@@ -66,7 +66,8 @@ type BackingServiceSelector struct {
 	metav1.GroupVersionKind `json:",inline"`
 	ResourceRef             string `json:"resourceRef"`
 	// +optional
-	Namespace *string `json:"namespace,omitempty"`
+	Namespace    *string `json:"namespace,omitempty"`
+	EnvVarPrefix string  `json:"envVarPrefix,omitempty"`
 }
 
 // BoundApplication defines the application workloads to which the binding secret has
--- a/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
+++ b/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
@@ -67,7 +67,7 @@ type BackingServiceSelector struct {
 	ResourceRef             string `json:"resourceRef"`
 	// +optional
 	Namespace    *string `json:"namespace,omitempty"`
-	EnvVarPrefix string  `json:"envVarPrefix,omitempty"`
+	EnvVarPrefix *string `json:"envVarPrefix,omitempty"`
 }
 
 // BoundApplication defines the application workloads to which the binding secret has
--- a/pkg/controller/servicebindingrequest/retrieve_decoupled.go
+++ b/pkg/controller/servicebindingrequest/retrieve_decoupled.go
@@ -36,11 +36,11 @@ func (r *Retriever) Get() (map[string][]byte, error) {
 // ReadBindableResourcesData reads all related resources of a given sbr
 func (r *Retriever) ReadBindableResourcesData(
 	sbr *v1alpha1.ServiceBindingRequest,
-	crs []*unstructured.Unstructured,
+	relatedResources RelatedResources,
 ) error {
 	r.logger.Info("Detecting extra resources for binding...")
-	for _, cr := range crs {
-		b := NewDetectBindableResources(sbr, cr, []schema.GroupVersionResource{
+	for _, rs := range ([]*RelatedResource)(relatedResources) {
+		b := NewDetectBindableResources(sbr, rs.CR, []schema.GroupVersionResource{
 			{Group: "", Version: "v1", Resource: "configmaps"},
 			{Group: "", Version: "v1", Resource: "services"},
 			{Group: "route.openshift.io", Version: "v1", Resource: "routes"},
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by sbose78** [comment]
Thank you!

/ok-to-test

File: pkg/controller/servicebindingrequest/retriever.go [hunk 1]
```diff
@@ -235,21 +235,29 @@ func (r *Retriever) readConfigMap(cr *unstructured.Unstructured, name string, it
...
+func (r *Retriever) store(envPrefix string, u *unstructured.Unstructured, key string, value []byte) {
 	key = strings.ReplaceAll(key, ":", "_")
 	key = strings.ReplaceAll(key, ".", "_")
-	if r.bindingPrefix == "" {
-		key = fmt.Sprintf("%s_%s", u.GetKind(), key)
+	if envPrefix == "" {
+		if r.bindingPrefix == "" {
+			key = fmt.Sprintf("%s_%s", u.GetKind(), key)
+		} else {
+			key = fmt.Sprintf("%s_%s_%s", r.bindingPrefix, u.GetKind(), key)
```
sbose78, line 253: So,
* if no envVarPrefix is provided, then we use the existing format of prefix-ing the `Kind`, else, we use the envVarPrefix.
* if an envVarPrefix is provided, we don't use the `Kind`, rather we use the provided envVarPrefix.

pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -235,21 +235,29 @@ func (r *Retriever) readConfigMap(cr *unstructured.Unstructured, name string, it
...
+		r.store(envPrefix, cr, fmt.Sprintf("configMap_%s", k), []byte(value))
 	}
 
 	r.Objects = append(r.Objects, u)
 	return nil
 }
 
 // store key and value, formatting key to look like an environment variable.
-func (r *Retriever) store(u *unstructured.Unstructured, key string, value []byte) {
+func (r *Retriever) store(envPrefix string, u *unstructured.Unstructured, key string, value []byte) {
```
sbose78, line 246: I wonder if it's better to have `envPrefix *string` ? There's a chance that if the `envVarPrefix` was skipped by the user in the CR, the value would actually resolve to an empty string.

pkg/apis/apps/v1alpha1/servicebindingrequest_types.go
```diff
@@ -66,7 +66,8 @@ type BackingServiceSelector struct {
 	metav1.GroupVersionKind `json:",inline"`
 	ResourceRef             string `json:"resourceRef"`
 	// +optional
-	Namespace *string `json:"namespace,omitempty"`
+	Namespace    *string `json:"namespace,omitempty"`
+	EnvVarPrefix string  `json:"envVarPrefix,omitempty"`
```
sbose78, line 70: Maybe use `EnvVarPrefix *string` so that we can distinguish between 
* a skipped field, and
* an empty string
in our code.

Asking for this because:
* an empty string explicitly implies that the environment variable need to looks like `HOST`
* a skipped field implicitly implies that the environment variable needs to look like `KIND_HOST`

pkg/controller/servicebindingrequest/retrieve_decoupled.go
```diff
@@ -51,36 +51,36 @@ func (r *Retriever) ReadBindableResourcesData(
...
+			r.storeInto(rs.EnvVarPrefix, rs.CR, k, []byte(fmt.Sprintf("%v", v)))
 		}
 	}
 
 	return nil
 }
 
-func (r *Retriever) storeInto(cr *unstructured.Unstructured, key string, value []byte) {
-	r.store(cr, key, value)
+func (r *Retriever) storeInto(envPrefix string, cr *unstructured.Unstructured, key string, value []byte) {
```
sbose78, line 61: `envPrefix` --> `envVarPrefix` for consistency.

qibobo:
@sbose78 
Thank you for your comments. I have updated the code and added the relevant unit tests.

sbose78:
What does the behaviour look like if there's a global env var prefix and there's a local env var prefix?

qibobo:
@sbose78 
When global env prefix and local env prefix are both not empty, the result is GlobalEnvPrefix_LocalEnvPrefix_Key, just as https://github.com/qibobo/service-binding-operator/blob/396316955c397446a17a05a075353c95d34c3682/pkg/controller/servicebindingrequest/retriever.go#L261.
And the ut for this is https://github.com/qibobo/service-binding-operator/blob/396316955c397446a17a05a075353c95d34c3682/pkg/controller/servicebindingrequest/retriever_test.go#L126.

qibobo:
rebase the master.

Avni-Sharma:
Case 1- when envVarPrefix is provided
Screenshot from 2020-04-20 21-46-27

Avni-Sharma:
Case 2 - when envVarPrefix is not mentioned (is omitted), then append with KIND_
Screenshot from 2020-04-20 21-53-29

Avni-Sharma:
Case 3- when envVarPrefix is given as an empty string `envVarPrefix: ""`

Screenshot from 2020-04-20 21-55-25

pkg/controller/servicebindingrequest/retriever.go
```diff
@@ -197,7 +197,7 @@ func (r *Retriever) readSecret(cr *unstructured.Unstructured, name string, items
 		// update cache after reading configmap/secret in cache
 		r.cache[fromPath].(map[string]interface{})[path].(map[string]interface{})[k] = string(data)
 		// making sure key name has a secret reference
-		r.store(cr, fmt.Sprintf("secret_%s", k), data)
+		r.store(envVarPrefix, cr, fmt.Sprintf("secret_%s", k), data)
```
sbose78, line 200: May I request an associated change? 
If the user has specified an `envVarPrefix : ""`, could we even skip the `secret_xyz` / `configmap_xyz` ?

sbose78:
Thank you @Avni-Sharma for the validation.

qibobo:
@sbose78 Per my understanding, the target result is as below, right?
|  GlobalEnvVarPrefix   | ServiceEnvVarPrefix | Key|Result|
|  ----  | ----  |  ----  | ----  |
|  globalprefix |  nil|  apikey |  GLOBALPREFIX_SECRET_APIKEY|
| ""  | nil |  apikey | SECRET_APIKEY |
|  globalprefix |  serviceprefix|  apikey |  GLOBALPREFIX_SERVICEPREFIX_SECRET_APIKEY|
|  globalprefix |  ""|  apikey |  GLOBALPREFIX_APIKEY|
|  "" |  serviceprefix|  apikey |  SERVICE_SECRET_APIKEY|
|  "" | ""|  apikey |  APIKEY|

sbose78:
Yes!

qibobo:
@sbose78 
Thank you for your comment. Updated the code.

qibobo:
@sbose78 @Avni-Sharma 
any comments?

Issue #429, opened 2020-04-19
SBR cannot get deleted if customEnvVar has a bad template


int8dev:
If a sbr is created with a bad customEnvVar template, the sbr cannot be deleted.

e.g.
Create SBR as following
```
apiVersion: apps.openshift.io/v1alpha1
kind: ServiceBindingRequest
metadata:
  name: hello-bindingrequest
  namespace: default
spec:
  applicationSelector:
    group: serving.knative.dev
    resource: services
    resourceRef: hello
    version: v1
  backingServiceSelectors:
  - group: ibmcloud.ibm.com
    kind: Binding
    resourceRef: yingcos-binding
    version: v1alpha1
  customEnvVar:
  - name: INSTANCE_ID
    value: "{{ .status.instanceId "     # bad template
```
We will get error while reconciling even though deletion is triggered by `kubectl`.
```
level:error ts:1587310932.8735907 logger:controller-runtime.controller msg:Reconciler error controller:servicebindingrequest-controller request:default/hello-bindingrequest error:template: set:1: unclosed action stacktrace:github.com/go-logr/zapr.(*zapLogger).Error
	service-binding-operator/vendor/github.com/go-logr/zapr/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	service-binding-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:258
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	service-binding-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:232
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
	service-binding-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:211
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
	service-binding-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:152
k8s.io/apimachinery/pkg/util/wait.JitterUntil
	service-binding-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:153
k8s.io/apimachinery/pkg/util/wait.Until
	service-binding-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88
```

[...]

PR #430, opened 2020-04-19
check SBR deletion before calling retriever.Get()

int8dev:
### Motivation

If a sbr is created with a bad customEnvVar template, the sbr cannot be deleted. Issue https://github.com/redhat-developer/service-binding-operator/issues/429

### Changes

Check the SBR's `DeletionTimestamp` before calling `Retriever.Get()` as the data is not used in deletion path.

### Testing
Unit test cases added


sbose78:
Thank you very much!

PR #414, opened 2020-04-07
Status should be updated when application selector is missing

DhritiShikhar:
fixes #408 

## Motivation

When `applicationSelector` was missing from SBR, the `status` field would not get updated. This PR fixes this issue.

Sample SBR:

```
apiVersion: apps.openshift.io/v1alpha1
kind: ServiceBindingRequest
metadata:
  name: binding-request
  namespace: sbo-try-1
spec:
  backingServiceSelectors:
        - resourceRef: db-demo
          group: postgresql.baiju.dev
          version: v1alpha1
          kind: Database
```

PR #432, opened 2020-04-22
fixes generic binding

DhritiShikhar:
### Motivation

- Fixes generic binding bug
- Refer -> https://github.com/redhat-developer/service-binding-operator/pull/414/files#r407707455

### Changes

- Reverts some changes introduced in this PR ->  https://github.com/redhat-developer/service-binding-operator/pull/414

### How to test

- Import an application
- Create a DB instance 

```
kind: ServiceBindingRequest
metadata:
  name: binding-request
spec:
 backingServiceSelector:
    group: postgresql.baiju.dev
    version: v1alpha1
    kind: Database
    resourceRef: db-demo
```

3. Use the following command to check if the binding secret has been generated:

The output should have:

```
  secret: binding-request
```

### DIFF
--- a/pkg/controller/servicebindingrequest/binding_test.go
+++ b/pkg/controller/servicebindingrequest/binding_test.go
@@ -545,7 +545,7 @@ func TestServiceBinder_Bind(t *testing.T) {
 			SBR:                    sbrEmptyAppSelector,
 			Client:                 f.FakeClient(),
 		},
-		wantBuildErr: EmptyApplicationSelectorErr,
+		wantErr: EmptyApplicationSelectorErr,
 		wantConditions: []wantedCondition{
 			{
 				Type:    conditions.BindingReady,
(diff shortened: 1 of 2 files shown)


Avni-Sharma:
Making me wonder if we really need the err EmptyApplicationSelectorErr, at all https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/controller/servicebindingrequest/reconciler.go#L153 
The generic binding should not log this as an error

DhritiShikhar:
We need `EmptyApplicationSelectorErr` to return proper error message in our `Status.Conditions`.
With this PR, generic binding updates the status with `EmptyApplicationSelectorErr` reason and keeps on reconciling until the user updates the CR.

However, the reconcile method needs a bit of cleanup as this function is never reached -> https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/controller/servicebindingrequest/reconciler.go#L152

We could create another issue for this cleanup.

This PR is about reverting some changes made in #414

PR #434, opened 2020-04-22
Add a choice of installing SBO either from community or master in examples

pmacik:
### Motivation

Currently, the examples use the version of the SBO installed from `community-operators`. I was not easy to try those examples with the latest version of the examples that are built from `master`.

Providing means to run those examples with the latest `master` version of SBO also supports automated tests that implement the example's scenarios.

This is a dependency for #441.

### Changes

This PR adds make target for installing the latest SBO version from `master` using a custom `OperatorSource` and renames the current make targets (for the community version) by adding `community` suffix.

### Testing

Tests introduced in #441 use this update.

### DIFF
--- a/examples/commons.mk
+++ b/examples/commons.mk
@@ -82,23 +82,55 @@ help: ## Credit: https://gist.github.com/prwhite/8168133#gistcomment-2749866
 
 ## --- Service Binding Operator ---
 
-.PHONY: install-service-binding-operator-subscription
+## ---- Community version ----
+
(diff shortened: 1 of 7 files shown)


sbose78:
If I had to do manually with yamls, how would I do so?

pmacik:
> If I had to do manually with yamls, how would I do so?

@sbose78 Same as before. I added the instructions for manual steps into the docs, if that's what you were asking for

PR #435, opened 2020-04-23
[WIP] Removes finalizer, updates status when applicationSelector is removed

### LINKED ISSUE #384: Generic binding: SBR needs to be cleaned up if applicationSelector is removed
Once `applicationSelector` is removed from a `ServiceBindingRequest`, the operator should update the CR to:

1. remove the `finalizer`
1. update `status.applications`

 ``` yaml
...
metadata:
  finalizers:
    - finalizer.servicebindingrequest.openshift.io
...
status:
  applications:
    - group: apps
      kind: Deployment
      name: barista-kafka
      version: v1
...
```

Also, the binding status is incorrect which causes operator to keep reconciling. There's already an issue open for this: https://github.com/redhat-developer/service-binding-operator/issues/381.

### PR DESCRIPTION
**Pull request by DhritiShikhar**
### Motivation

To fix  #384  

### Changes

When `applicationSelector` is removed from `ServiceBindingRequest`, the operator updates the CR to:

- Remove `finalizer` from SBR metadata
- update `status.applications` 

### How to test

1. Follow all the steps given in this [example](https://github.com/redhat-developer/service-binding-operator/tree/master/examples/nodejs_postgresql) to:
- Import an application
- Create a DB instance
- Create a ServiceBindingRequest

2. Use the following command to check the `ServiceBindingRequest`:

Notice the presence of `finalizer` in metadata:

```
metadata:
 finalizers:
   - finalizer.servicebindingrequest.openshift.io
```

Also, notice the application name in `status.applications`

```
status:
  applications:
  - group: apps
    kind: Deployment
    name: nodejs-rest-http-crud
    version: v1

```

3. Update the `ServiceBindingRequest` to remove the `applicationSelector`:

```
kind: ServiceBindingRequest
metadata:
  name: binding-request
spec:
 backingServiceSelector:
    group: postgresql.baiju.dev
    version: v1alpha1
    kind: Database
    resourceRef: db-demo
```

4. Use the following command to check if the status has been updated:

The output should not have:

```
metadata:
 finalizers:
   - finalizer.servicebindingrequest.openshift.io
```

and output should not have `status.applications`

### DIFF
--- a/pkg/controller/servicebindingrequest/binding.go
+++ b/pkg/controller/servicebindingrequest/binding.go
@@ -231,6 +231,17 @@ func (b *ServiceBinder) Bind() (reconcile.Result, error) {
 	updatedObjects, err := b.Binder.Bind()
 	if err != nil {
 		b.Logger.Error(err, "On binding application.")
+
+		// Delete Finalizer when applicationSelector is removed
+		if err == EmptyApplicationSelectorErr {
+
+			b.SBR.SetFinalizers(removeStringSlice(b.SBR.GetFinalizers(), Finalizer))
+			if _, err := b.updateServiceBindingRequest(b.SBR); err != nil {
+				return NoRequeue(err)
+			}
+
+		}
+
 		return b.onError(err, b.SBR, sbrStatus, updatedObjects)
 	}
 	b.setApplicationObjects(sbrStatus, updatedObjects)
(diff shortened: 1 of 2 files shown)


isutton:
Are we sure finalizers need to be removed when no application selector is informed? The finalizer is required to give the operator time to remove all resources associated with the service binding request before the API server removes the entry from etcd.

Without the finalizer the SBR will be promptly removed and any failure during this last reconciliation will leave resources behind.

Avni-Sharma:
When applicationSelector is not present, it is a generic binding case and in generic binding, the secret should be available, so that it can be referenced elsewhere. So we actually don't need to delete the intermediate secret in a case when there is no applicationSelector. In that case the CollectionReady status value can be shown as True

DhritiShikhar:
Agreed. Will close this PR.

Issue #422, opened 2020-04-14
Controller doesn't reconcile to changes to application workloads


cdlliuy:
```
spec:
  applicationSelector:
    group: serving.knative.dev
    resource: services
    resourceRef: mydemo
    version: v1
.....
```
Once the sbr created,  I tried to delete `mydemo` , then create it again with the same name, but  SBR didn't  reconcile to have the secrets injected into the new `mydemo` container anymore.   

Similar to backingEndServiceSelectors as well,  I delete a specific backingservice, then create it again with the same name but different secret content.  Anyway, the new secret didn't inject into the application container. 

In short:
1.  create sbr successfully
2.  delete application  
3.  create application again with the same name.  Keep SBR unchanged
4.   no env injected in the new application container. 

OR
1.  create sbr successfully
2.  delete the backingservice  
3.  create backingservice again with the same name.  Keep SBR unchanged
4.  With the new backingservice, the service credential is changed.   But the injected env in application containers is not updated.

sbose78:
Thanks for reporting this.

@isutton @baijum @DhritiShikhar @Avni-Sharma  could you please take a look?

DhritiShikhar:
Yes @sbose78 I am looking into this..

cdlliuy:
@DhritiShikhar  Thanks and eager to know your findings :-)

int8dev:
AFAIK currently operator only watch the following resources

- ServiceBindingRequest CRD 
- Configmaps
- Secrets
- CRDs owned by ClusterServiceVersion resources.

So If we specify other CRDs in `applicationSelector` or `backingServiceSelectors`, reconcile would not be triggered as they are not watched by informer. 

And I also noticed another potential issue that after a success reconcile for knative service, there are no SBR informations in its annotations field. So I think even we added operator to watch knative services the reconcile will not working as expected. 

Checking knative service after creating SBR successsfully. No annotations for SBR information.
```
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  annotations:
    serving.knative.dev/lastModifier: system:serviceaccount:ibm-service-binding:service-binding-operator
  creationTimestamp: "2020-04-22T13:31:11Z"
  generation: 3
  name: helloworld
  namespace: 930852e2-7e33
  resourceVersion: "48100370"
  selfLink: /apis/serving.knative.dev/v1/namespaces/930852e2-7e33/services/helloworld
[...]

PR #442, opened 2020-04-29
Add watching for workloads

lanceliuu:
### Motivation
Fix issue https://github.com/redhat-developer/service-binding-operator/issues/422

### Changes
Besically align https://github.com/redhat-developer/service-binding-operator/issues/422#issuecomment-619260714 Except:
Instead of hard-coded watching typical workloads at starting, this change will add watching for GVKs defined in application selector at runtime. As for typical workloads with status as subresources,  the existing predicate function which checks the`metadata.generation` is able to avoid noises.

Also split `Set/Remove SBRAnnotations` function into two seperate functions. It is useful to combine the updating for application CR. 

### Testing
Unit tests added and manual testing passed.

### DIFF
--- a/pkg/controller/servicebindingrequest/binder.go
+++ b/pkg/controller/servicebindingrequest/binder.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 
 	"gotest.tools/assert/cmp"
 	corev1 "k8s.io/api/core/v1"
@@ -480,6 +481,9 @@ func (b *Binder) update(objs *unstructured.UnstructuredList) ([]*unstructured.Un
 		log := b.logger.WithValues("Obj.Name", name, "Obj.Kind", obj.GetKind())
 		log.Debug("Inspecting object...")
 
+		sbrNamespacedName := types.NamespacedName{Namespace: b.sbr.GetNamespace(), Name: b.sbr.GetName()}
+		updatedObj = SetSBRAnnotations(sbrNamespacedName, updatedObj)
+
 		updatedObj, err := b.updateSpecContainers(updatedObj)
 		if err != nil {
 			return nil, err
@@ -491,7 +495,7 @@ func (b *Binder) update(objs *unstructured.UnstructuredList) ([]*unstructured.Un
 			}
 		}
 
-		if specsAreEqual, err := nestedMapComparison(&obj, updatedObj, "spec"); err != nil {
+		if specsAreEqual, err := nestedMapComparison(&obj, updatedObj); err != nil {
 			log.Error(err, "")
 			continue
 		} else if specsAreEqual {
@@ -535,6 +539,8 @@ func (b *Binder) remove(objs *unstructured.UnstructuredList) error {
 		if err != nil {
 			return err
 		}
+		updatedObj = RemoveSBRAnnotations(updatedObj)
+
 		if len(b.volumeKeys) > 0 {
 			if updatedObj, err = b.removeSpecVolumes(updatedObj); err != nil {
 				return err
--- a/pkg/controller/servicebindingrequest/reconciler.go
+++ b/pkg/controller/servicebindingrequest/reconciler.go
@@ -9,6 +9,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -27,9 +28,10 @@ const (
 
 // Reconciler reconciles a ServiceBindingRequest object
 type Reconciler struct {
-	dynClient  dynamic.Interface // kubernetes dynamic api client
-	scheme     *runtime.Scheme   // api scheme
-	RestMapper meta.RESTMapper   // restMapper to convert GVK and GVR
+	dynClient       dynamic.Interface // kubernetes dynamic api client
+	scheme          *runtime.Scheme   // api scheme
+	RestMapper      meta.RESTMapper   // restMapper to convert GVK and GVR
+	resourceWatcher ResourceWatcher   // ResourceWatcher to add watching for specific GVK/GVR
 }
 
 // reconcilerLog local logger instance
@@ -184,6 +186,18 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		// just bail out without re-queueing nor updating conditions.
 		logger.Error(err, "Building ServiceBinder")
 		return NoRequeue(err)
+	}
+
+	gvrSpec := sbr.Spec.ApplicationSelector.GroupVersionResource
+	gvr := schema.GroupVersionResource{
+		Group:    gvrSpec.Group,
+		Version:  gvrSpec.Version,
+		Resource: gvrSpec.Resource,
+	}
+
+	err = r.resourceWatcher.AddWatchForGVR(gvr)
+	if err != nil {
+		logger.Error(err, "Error add watching application GVR")
 	}
 
 	if sbr.GetDeletionTimestamp() != nil {
--- a/pkg/controller/servicebindingrequest/sbrcontroller.go
+++ b/pkg/controller/servicebindingrequest/sbrcontroller.go
@@ -6,6 +6,7 @@ import (
 
 	olmv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -22,13 +23,22 @@ import (
 	"github.com/redhat-developer/service-binding-operator/pkg/log"
 )
 
+// ResourceWatcher add watching for GroupVersionKind/GroupVersionResource
+type ResourceWatcher interface {
+	AddWatchForGVR(schema.GroupVersionResource) error
+	AddWatchForGVK(schema.GroupVersionKind) error
+}
+
 // SBRController hold the controller instance and methods for a ServiceBindingRequest.
 type SBRController struct {
 	Controller   controller.Controller            // controller-runtime instance
 	Client       dynamic.Interface                // kubernetes dynamic api client
+	RestMapper   meta.RESTMapper                  // restMapper to convert GVK and GVR
 	watchingGVKs map[schema.GroupVersionKind]bool // cache to identify GVKs on watch
 	logger       *log.Log                         // logger instance
 }
+
+var _ ResourceWatcher = (*SBRController)(nil)
 
 // controllerName common name of this controller
 const controllerName = "servicebindingrequest-controller"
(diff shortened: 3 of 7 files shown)


int8dev:
@DhritiShikhar Please help to review and comment. Thanks!

pkg/controller/servicebindingrequest/annotations.go
```diff
@@ -100,45 +100,71 @@ func updateUnstructuredObj(client dynamic.Interface, obj *unstructured.Unstructu
 	return err
 }
 
-// SetSBRAnnotations update existing annotations to include operator's. The annotations added are
+// SetSBRAnnotations set nnotations to include SBR information. Return true if object modified.
```
DhritiShikhar, line 103: typo 

/s/nnotations/annotations

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -477,6 +477,9 @@ func (b *Binder) update(objs *unstructured.UnstructuredList) ([]*unstructured.Un
 		log := b.logger.WithValues("Obj.Name", name, "Obj.Kind", obj.GetKind())
 		log.Debug("Inspecting object...")
 
+		sbrNamespacedName := types.NamespacedName{Namespace: b.sbr.GetNamespace(), Name: b.sbr.GetName()}
+		annotationChanged := SetSBRAnnotations(sbrNamespacedName, &obj)
+		log.Debug("anno changed")
```
DhritiShikhar, line 482: Would be good to have `annotation` instead of `anno`

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -19,9 +22,11 @@ import (
...
 // Reconciler reconciles a ServiceBindingRequest object
 type Reconciler struct {
-	client    client.Client     // kubernetes api client
-	dynClient dynamic.Interface // kubernetes dynamic api client
-	scheme    *runtime.Scheme   // api scheme
+	client            client.Client     // kubernetes api client
+	dynClient         dynamic.Interface // kubernetes dynamic api client
+	scheme            *runtime.Scheme   // api scheme
+	ensureWatchingGVR func(gvr schema.GroupVersionResource) error
+	restMapper        meta.RESTMapper
```
DhritiShikhar, line 29: Why is this `restMapper` added here? I don't see it anywhere being used for `Reconcile`.

pkg/controller/servicebindingrequest/binder.go
int8dev: I will remove this debugging as it is not useful.

pkg/controller/servicebindingrequest/reconciler.go
int8dev: You are right. We should add `restMapper` to `SBRController` only.

DhritiShikhar:
@int8dev I tested this PR using this [example](https://github.com/redhat-developer/service-binding-operator/tree/master/examples/nodejs_postgresql):

[1] Creating a PostgreSQL DB
[2] Creating an SBR
[3] Creating a nodejs application
[4] Deleting the nodejs application
[5] Re-creating the nodejs application with the same name

But I don't see secret injected in the nodejs application.

Here is the video for the same -> https://youtu.be/q6pFTaD-WWM

int8dev:
Hi @DhritiShikhar, thanks for your feedback.

I think the reason is that the delete event is ignored in step 4). So controller does not enqueue the SBR for reconciling. We can remediate this in the to return true in `DeleteFunc` of `predicates`. After this change the SBR resource will be in a fail-reconcile loop as the application CR is not found until the application CR is recreated.
For other cases if we try to update the application object (e.g, remove the envs injected by operator), the reconcile should be triggered and get the injection fixed.

This should be easy fix but I found another issue in my local testing. After I deleted the application CR, I can confirm the reconcile triggered as expected. But the reconcile ended up with success even the application no longer exist. This can also be observed If we create a SBR directly with a non-existing `resourceRef` in `applicationSelector`. I think it is introduced in the recent fixes for deletion issues.

DhritiShikhar:
@int8dev 

Yes, I have tested [this]( https://github.com/redhat-developer/service-binding-operator/issues/422#issuecomment-618484986) scenario and the case is solved by your PR.

[Here](https://youtu.be/h2Vh0Yt35gA) is a video of my test. 

However, We need to add a fix for the [original comment ](https://github.com/redhat-developer/service-binding-operator/issues/422#issue-599321019) also which is the deletion issue.

We can handle that in another PR. Meanwhile, I will merge this PR when the CI tests pass.

int8dev:
For specific, I have no idea why this error is ignored. 
https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/controller/servicebindingrequest/binder.go#L569

DhritiShikhar:
@int8dev 

Yes. There have been some issues that have been created by a commit -> https://github.com/redhat-developer/service-binding-operator/commit/f5dfc45e069ef32d7d6fee04e1f575f068d04ae5

We are working on those.

isutton:
@DhritiShikhar Can you hold this for a moment so I can review it?

I'd like to also ask to separate the changes in `SetSBRAnnotations` and the watch feature itself.

DhritiShikhar:
@isutton Sure

int8dev:
Please hold this PR and I created another PR https://github.com/redhat-developer/service-binding-operator/pull/444 as request by @isutton. 
I will rebase the changes in this PR with better interface abstraction for watching.
Thanks you! @isutton

Avni-Sharma:
> Hi @DhritiShikhar, thanks for your feedback.
> 
> I think the reason is that the delete event is ignored in step 4). So controller does not enqueue the SBR for reconciling. If the delete event is handled properly the SBR resource will be in a fail-reconcile loop as the application CR is not found until the application CR is recreated.
> 
> For other cases if we try to update the application object (e.g, remove the envs injected by operator), the reconcile should be triggered and get the injection fixed.
> 
> I did some test from my side. After I deleted the application CR. I can confirm the reconcile triggered as expected. But the reconcile ended up with success even the application no longer exist. This can also be observed If we create a SBR directly with a non-existing `resourceRef` in `applicationSelector`. I think it is introduced in the recent fixes for deletion issues.

@int8dev  So if we delete an application, and then try to delete the sbr instance, then the sbr won't get deleted as it will keep searching for the application to come up again, and since the application is not available, it will be requed and the deletion of the sbr goes into a deadlock.

int8dev:
@Avni-Sharma I think we should check the deletionTimestamp of SBR before we can actually ignore the application not found error. If the SBR is not in a deletion path, it is expected that the SBR goes into a fail-requeue state.

DhritiShikhar:
1. Created a DB, application, SBR
2. Deleted application successfully
3. Deleted SBR successfully

[Here](https://youtu.be/_OGQqUGuqoo) is a video for the same

sbose78:
Thanks!

@isutton , why do we need to make two PRs for this? Given the refactor was needed for a specific capability to work, I feel it should be a part of the same PR. It helps account for the reasons behind a refactor later on if they are in the same commit in master.

int8dev:
@isutton 
I like the idea of watching all kinds of resources registered in apiserver as discussed in the meeting. But I don't think it is possible for the following reasons.
1) There is no "magic" behind the controllers/informers. For specific, the cotroller does not pull all resources in the apiserver and generate event for reconciling which is too much overhead. Basically the underlying behaviour of informer is a listing/watching request for specific GVK such as. `GET https://api.some-kube-apiserver/apis/apps.openshift.io/v1alpha1/servicebindingrequests?resourceVersion=xxx&watch=true`. That is what a `Reflector` does in client-go. There is no API to list/watch all kinds of resources in apiserver.
For operator-sdk which use controller-runtime to create informers, the functions are defined in https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/informers_map.go#L230-L302. We can tell that each informer for GVK is attatched to a list/watch request for a dynamicClient
2) There should be RBAC rules for controllers to limit the access to apiserver. That meas even we could get all GVKs installed in the cluster, it is barely impossible for controller to list/watch all kinds of resources.

isutton:
> @isutton
> I like the idea of watching all kinds of resources registered in apiserver as discussed in the meeting. But I don't think it is possible for the following reasons.
> 
> 1. There is no "magic" behind the controllers/informers. For specific, the cotroller does not pull all resources in the apiserver and generate event for reconciling which is too much overhead. Basically the underlying behaviour of informer is a listing/watching request for specific GVK such as. `GET https://api.some-kube-apiserver/apis/apps.openshift.io/v1alpha1/servicebindingrequests?resourceVersion=xxx&watch=true`. That is what a `Reflector` does in client-go. There is no API to list/watch all kinds of resources in apiserver.
>    For operator-sdk which use controller-runtime to create informers, the functions are defined in https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/informers_map.go#L230-L302. We can tell that each informer for GVK is attatched to a list/watch request for a dynamicClient.

IIRC, the first reconciliation when the operator starts, all resources for a given GVK are *pulled* from the API server and locally enqueued to wait for a worker to process.

If I gave the impression that there were some ready API for doing that, I apologize; it is required to [list all APIResources](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#apiresourcelist-v1-meta) and individually watch for each known GVK using the aforementioned APIs.

> 2. There should be RBAC rules for controllers to limit the access to apiserver. That means even we could get all GVKs installed in the cluster, it is barely impossible for controller to list/watch all kinds of resources.

That's an interesting point that should be further discussed IMO.

Are you planning to have one Operator running per project or cluster wide?

isutton:
> @isutton , why do we need to make two PRs for this? :) Given the refactor was needed for a specific capability to work, I feel it should be a part of the same PR. It helps account for the reasons behind a refactor later on if they are in the same commit in master.

From my understanding, the functionality this PR provides are present only in `reconciler.go`, `controller.go` and `sbrcontroller.go`.

IMHO the other changes are not related and should have its own discussion; keeping it will only add noise to this PR.

int8dev:
Thank you for the clarification @isutton.

> Are you planning to have one Operator running per project or cluster wide?

For our current design we decide to run service-binding-operator at cluster level. And the cluster might be pretty big (i.e >100 nodes, >10k namespaces).

sbose78:
Hi @int8dev Anything to be done here?

lanceliuu:
I will rebase this PR after some fixes of the testing. That will be in another PR.

Avni-Sharma:
Hi @lanceliuu this PR is waiting on which PR, can you link that PR here

lanceliuu:
@Avni-Sharma This now depends on PR https://github.com/redhat-developer/service-binding-operator/pull/457

Avni-Sharma:
Are we considering on watching over all  backing services as well or this PR is focused only on the application workload?
cc @isutton

lanceliuu:
@Avni-Sharma  @isutton 
I have rebased this PR on master branch.

isutton:
> Are we considering on watching over all backing services as well or this PR is focused only on the application workload?
> cc @isutton

I believe this can be considered finished, and a new PR should be created for the services' watcher implementation.

DhritiShikhar:
@isutton @Avni-Sharma Some extra support is needed to add service watcher. I will raise a PR for it.

pkg/controller/servicebindingrequest/sbrcontroller.go
```diff
@@ -144,6 +154,15 @@ func (s *SBRController) AddWatchForGVK(gvk schema.GroupVersionKind) error {
 	return s.Controller.Watch(src, s.newEnqueueRequestsForSBR(), buildGVKPredicate(logger))
 }
 
+// AddWatchForGVR creates a watch on a given GVR
+func (s *SBRController) AddWatchForGVR(gvr schema.GroupVersionResource) error {
```
Avni-Sharma, line 158: logs should be added here 
```
	logger := s.logger.WithValues("GVK", gvk)
	logger.Debug("Adding watch for GVK...")
```

DhritiShikhar:
The code coverage is failing because there are no tests for `AddWatchForGVK()` method in 

.../controller/servicebindingrequest/sbrcontroller.go


https://codecov.io/gh/redhat-developer/service-binding-operator/pull/442/diff?src=pr&el=tree#diff-cGtnL2NvbnRyb2xsZXIvc2VydmljZWJpbmRpbmdyZXF1ZXN0L3NicmNvbnRyb2xsZXIuZ28=

lanceliuu:
@DhritiShikhar test cases added. Please help to re-run tests as there is no `ok-to-test` label and I cannot trigger tests. Thanks.

DhritiShikhar:
The e2e tests are failing for me on master.
I have created an issue for this -> https://github.com/redhat-developer/service-binding-operator/issues/497

Can we merge this PR as an exception?
and fix the e2e test then?

Avni-Sharma:
manually tested, passed for me
/lgtm

DhritiShikhar:
@isutton PR looks good. Can we approve it?

PR #444, opened 2020-04-29
Separate SetSBRAnnotations/RemoveSBRAnnotations functions

int8dev:
### Motivation
Part of https://github.com/redhat-developer/service-binding-operator/pull/442 as requested by @isutton  

### Changes
Seperate `SetSBRAnnotations`/`RemoveSBRAnnotations` functions.
After this change `SetSBRAnnotations`/`RemoveSBRAnnotations` is only used to modify the object without updating it using clientset. The original behaviour is changed to `SetAndUpdateSBRAnnotations`/`RemoveAndUpdateSBRAnnotations`.

This is helpful for functions which want to combine set/remove SBRAnnotations with other modifications as a single updating request.

### Testing
Test cases included.

### DIFF
--- a/pkg/controller/servicebindingrequest/annotations.go
+++ b/pkg/controller/servicebindingrequest/annotations.go
@@ -100,45 +100,68 @@ func updateUnstructuredObj(client dynamic.Interface, obj *unstructured.Unstructu
 	return err
 }
 
-// SetSBRAnnotations update existing annotations to include operator's. The annotations added are
+// SetSBRAnnotations set annotations to include SBR information and return a new object.
+func SetSBRAnnotations(namespacedName types.NamespacedName,
+	obj *unstructured.Unstructured) *unstructured.Unstructured {
+	newObj := obj.DeepCopy()
+	annotations := newObj.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+	annotations[sbrNamespaceAnnotation] = namespacedName.Namespace
+	annotations[sbrNameAnnotation] = namespacedName.Name
+	newObj.SetAnnotations(annotations)
+	return newObj
+}
+
+// SetAndUpdateSBRAnnotations update existing annotations to include operator's. The annotations added are
 // referring to a existing SBR namespaced name.
-func SetSBRAnnotations(
+func SetAndUpdateSBRAnnotations(
 	client dynamic.Interface,
 	namespacedName types.NamespacedName,
 	objs []*unstructured.Unstructured,
 ) error {
 	for _, obj := range objs {
-		annotations := obj.GetAnnotations()
-		if annotations == nil {
-			annotations = make(map[string]string)
-		}
-
-		annotations[sbrNamespaceAnnotation] = namespacedName.Namespace
-		annotations[sbrNameAnnotation] = namespacedName.Name
-		obj.SetAnnotations(annotations)
-
-		if err := updateUnstructuredObj(client, obj); err != nil {
+		newObj := SetSBRAnnotations(namespacedName, obj)
+		equal, err := nestedMapComparison(obj, newObj, []string{"metadata", "annotations"}...)
+		if err != nil {
 			return err
+		}
+		if !equal {
+			if err := updateUnstructuredObj(client, newObj); err != nil {
+				return err
+			}
 		}
 	}
 	return nil
 }
 
-// RemoveSBRAnnotations removes SBR related annotations from all the objects and updates them using
+// RemoveSBRAnnotations removes SBR related annotations and return a new object.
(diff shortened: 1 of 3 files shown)


pkg/controller/servicebindingrequest/annotations.go
```diff
@@ -100,45 +100,71 @@ func updateUnstructuredObj(client dynamic.Interface, obj *unstructured.Unstructu
...
-			annotations = make(map[string]string)
-		}
-
-		annotations[sbrNamespaceAnnotation] = namespacedName.Namespace
-		annotations[sbrNameAnnotation] = namespacedName.Name
-		obj.SetAnnotations(annotations)
-
-		if err := updateUnstructuredObj(client, obj); err != nil {
-			return err
+		if SetSBRAnnotations(namespacedName, obj) {
```
isutton, line 132: I'd rather see the comparison in here using [this existing function](https://github.com/int8dev/service-binding-operator/blob/e822e2340e9cdcf8a899f691900ea1c49a1492de/pkg/controller/servicebindingrequest/binder.go#L441-L441).

```go
newObj := SetSBRAnnotations(namespacedName, obj)
ok, err := nestedMapComparison(obj, newObj, []string{"metadata", "annotations"})

```

pkg/controller/servicebindingrequest/annotations.go
```diff
@@ -100,45 +100,71 @@ func updateUnstructuredObj(client dynamic.Interface, obj *unstructured.Unstructu
...
+	obj.SetAnnotations(annotations)
+
+	return changed
+}
+
+// RemoveAndUpdateSBRAnnotations removes SBR related annotations from all the objects and updates them using
+// the given client.
+func RemoveAndUpdateSBRAnnotations(client dynamic.Interface, objs []*unstructured.Unstructured) error {
+	for _, obj := range objs {
+		if RemoveSBRAnnotations(obj) {
```
isutton, line 164: The same discussion in `SetSBRAnnotations` apply here.

pkg/controller/servicebindingrequest/annotations.go
int8dev: Thank you! I assume that the in-place map operations would be more cost-efficient. I will do the changes accordingly.

isutton: To be clear, the problem with the original code is the change management might be much more error prone than comparing the data structures for changes, even if performance wise it might be worse.

pkg/controller/servicebindingrequest/annotations_test.go
```diff
@@ -105,7 +105,55 @@ func TestAnnotationsSetAndRemoveSBRAnnotations(t *testing.T) {
 	objs = append(objs, u)
 
 	t.Run("SetSBRAnnotations", func(t *testing.T) {
-		err := SetSBRAnnotations(client, namespacedName, objs)
+		originCopy := u.DeepCopy()
```
isutton, line 108: Perhaps `originCopy` can be simplified since `SetSBRAnnotations` is already doing a deep copy.

pkg/controller/servicebindingrequest/annotations_test.go
```diff
@@ -105,7 +105,55 @@ func TestAnnotationsSetAndRemoveSBRAnnotations(t *testing.T) {
...
+
+		// assert nothing else is changed
+		newObj.SetAnnotations(nil)
+		equal, err = nestedMapComparison(u, newObj)
+		require.NoError(t, err)
+		require.True(t, equal)
+	})
+
+	t.Run("RemoveSBRAnnotations", func(t *testing.T) {
+		originCopy := u.DeepCopy()
```
isutton, line 132: Perhaps `originCopy` can be simplified since `RemoveSBRAnnotations` is already doing a deep copy.

pkg/controller/servicebindingrequest/annotations_test.go
int8dev: I don't think we can deduce the details of `SetSBRAnnotations` in test cases. So this test case can insure that func `SetSBRAnnotations` is really doing a deep copy.

isutton:
Thanks for your contribution!

/lgtm
/approve

pkg/controller/servicebindingrequest/annotations.go
```diff
@@ -100,46 +100,69 @@ func updateUnstructuredObj(client dynamic.Interface, obj *unstructured.Unstructu
 	return err
 }
 
-// SetSBRAnnotations update existing annotations to include operator's. The annotations added are
+// SetSBRAnnotations set annotations to include SBR information and return a new object.
+func SetSBRAnnotations(namespacedName types.NamespacedName,
+	obj *unstructured.Unstructured) *unstructured.Unstructured {
+	newObj := obj.DeepCopy()
```
baijum, line 106: What is the need for deep copy here?

PR #447, opened 2020-04-30
Add example for any resource as a backing service

Avni-Sharma:
Add examples doc for PR https://github.com/redhat-developer/service-binding-operator/pull/367


examples/route_k8s_resource/README.md
```diff
@@ -0,0 +1,126 @@
+# Use any resource as a backing service
```
sbose78, line 1: Maybe give it a more specific name like "Binding a nodejs app to a Route/Ingress".

Issue #450, opened 2020-05-08
Potential issues for unit tests caused by fake clients


lanceliuu:
There are two types of client in concrete type `Binder` in https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/controller/servicebindingrequest/binder.go#L43-L44.
```
// Binder executes the "binding" act of updating different application kinds to use intermediary
// secret. Those secrets should be offered as environment variables.
type Binder struct {
	ctx        context.Context                 // request context
	client     client.Client                   // kubernetes API client
	dynClient  dynamic.Interface               // kubernetes dynamic api client
	sbr        *v1alpha1.ServiceBindingRequest // instantiated service binding request
	volumeKeys []string                        // list of key names used in volume mounts
	modifier   ExtraFieldsModifier             // extra modifier for CRDs before updating
	logger     *log.Log                        // logger instance
}
```
In unit tests both of them are initliazed with provided objects. But the underneath behavior of both is to create deep copies of the objects. So the two client are holding two different objects and CRUDs might cause discrepancy. And the test results might also be affected.
https://github.com/kubernetes/client-go/blob/master/testing/fixture.go#L373-L376

Why we are holding two type of clients in `Binder`? I assume the dynamic client is enough for binder to interact with API server.

isutton:
Indeed, the dynamic client is enough for binder to interact with the API server.

This is a left over from before we have introduced the dynamic client, it's fine to remove it.

For reference`*unstructured.Unstructured` are fed to the test scenario to be used by the fake dynamic client.

PR #457, opened 2020-05-11
Use dynamic client to interact with the API server

lanceliuu:
### Motivation
Discussion https://github.com/redhat-developer/service-binding-operator/issues/450
Mixed using of clientsets will cause trouble for testing. As dynamic client is enough for binder to  interact with the API server, it is better to use a single dynamic client.
### Changes

- Remove dependency of client.Client in binder.
- Change the search behavior of binder for cases that `ResourceRef` is defined expilictly. Fake client will not filter objects by fieldselector.
- Add RESTMapper to binder to convert GVK and GVR

### Testing
Fixed potential test issues.

### DIFF
--- a/pkg/controller/servicebindingrequest/binder.go
+++ b/pkg/controller/servicebindingrequest/binder.go
@@ -7,7 +7,7 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
@@ -17,9 +17,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1"
 	"github.com/redhat-developer/service-binding-operator/pkg/log"
@@ -40,11 +38,11 @@ const ChangeTriggerEnv = "ServiceBindingOperatorChangeTriggerEnvVar"
 // secret. Those secrets should be offered as environment variables.
 type Binder struct {
 	ctx        context.Context                 // request context
-	client     client.Client                   // kubernetes API client
 	dynClient  dynamic.Interface               // kubernetes dynamic api client
 	sbr        *v1alpha1.ServiceBindingRequest // instantiated service binding request
 	volumeKeys []string                        // list of key names used in volume mounts
 	modifier   ExtraFieldsModifier             // extra modifier for CRDs before updating
+	restMapper meta.RESTMapper                 // RESTMapper to convert GVR from GVK
 	logger     *log.Log                        // logger instance
 }
 
@@ -67,7 +65,7 @@ var EmptyApplicationSelectorErr = errors.New("application ResourceRef or MatchLa
 var ApplicationNotFound = errors.New("Application is already deleted")
 
 // search objects based in Kind/APIVersion, which contain the labels defined in ApplicationSelector.
-func (b *Binder) search() (*unstructured.UnstructuredList, error) {
+func (b *Binder) search() ([]*unstructured.Unstructured, error) {
 	ns := b.sbr.GetNamespace()
 	gvr := schema.GroupVersionResource{
 		Group:    b.sbr.Spec.ApplicationSelector.GroupVersionResource.Group,
@@ -75,15 +73,18 @@ func (b *Binder) search() (*unstructured.UnstructuredList, error) {
 		Resource: b.sbr.Spec.ApplicationSelector.GroupVersionResource.Resource,
 	}
 
+	var result []*unstructured.Unstructured
 	var opts metav1.ListOptions
-
 	// If Application name is present
 	if b.sbr.Spec.ApplicationSelector.ResourceRef != "" {
-		fieldName := make(map[string]string)
-		fieldName["metadata.name"] = b.sbr.Spec.ApplicationSelector.ResourceRef
-		opts = metav1.ListOptions{
-			FieldSelector: fields.Set(fieldName).String(),
+		object, err := b.dynClient.Resource(gvr).Namespace(ns).
+			Get(b.sbr.Spec.ApplicationSelector.ResourceRef, metav1.GetOptions{})
+
+		if err != nil {
+			return nil, err
 		}
+		result = append(result, object)
+		return result, nil
 	} else if b.sbr.Spec.ApplicationSelector.LabelSelector != nil {
 		matchLabels := b.sbr.Spec.ApplicationSelector.LabelSelector.MatchLabels
 		opts = metav1.ListOptions{
@@ -95,11 +96,12 @@ func (b *Binder) search() (*unstructured.UnstructuredList, error) {
 
 	objList, err := b.dynClient.Resource(gvr).Namespace(ns).List(opts)
 	if err != nil {
-		return nil, ApplicationNotFound
-
+		return nil, err
 	}
-
-	return objList, err
+	for idx := range objList.Items {
+		result = append(result, &objList.Items[idx])
+	}
+	return result, nil
 }
 
 // extractSpecVolumes based on volume path, extract it unstructured. It can return error on trying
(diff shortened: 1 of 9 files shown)


sbose78:
@lanceliuu do you want to fix up history a bit?

lanceliuu:
I suggest hold this PR for now as using dynamic client will lost the cache feature from client of controller-runtime.  I noticed that there is issue to get a unstructured list using the fake client from controller-runtime as described at https://github.com/redhat-developer/service-binding-operator/tree/master/test/mocks#unstructured-list-vs-typed-resource.
It is more like a bug from the fake client and I am able to get desired result from a real client.

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -501,48 +502,46 @@ func (b *Binder) update(objs *unstructured.UnstructuredList) ([]*unstructured.Un
 		}
 
 		log.Debug("Updating object...")
-		if err := b.client.Update(b.ctx, updatedObj); err != nil {
-			return nil, err
-		}
+		gvr, _ := meta.UnsafeGuessKindToResource(updatedObj.GroupVersionKind())
```
isutton, line 505: It should be possible to propagate the Reconciler's `RestMapper` field here and use instead of `UnsafeGuessKindToResource`.

lanceliuu: Good point. I will change it accordingly.

lanceliuu:
@isutton  
What do you think we will lost the cache capability from informers for get/list actions if we change the typed client to unstructured client?

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -501,48 +503,51 @@ func (b *Binder) update(objs *unstructured.UnstructuredList) ([]*unstructured.Un
...
 		if len(b.volumeKeys) > 0 {
-			if updatedObj, err = b.removeSpecVolumes(&obj); err != nil {
+			if updatedObj, err = b.removeSpecVolumes(updatedObj); err != nil {
 				return err
 			}
 		}
 
-		logger.Debug("Updating object...")
-		if err = b.client.Update(b.ctx, updatedObj); err != nil {
+		gvr, _ := meta.UnsafeGuessKindToResource(updatedObj.GroupVersionKind())
```
isutton, line 541: Missing `b.restMapper` call here.

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -67,23 +65,26 @@ var EmptyApplicationSelectorErr = errors.New("application ResourceRef or MatchLa
 var ApplicationNotFound = errors.New("Application is already deleted")
 
 // search objects based in Kind/APIVersion, which contain the labels defined in ApplicationSelector.
-func (b *Binder) search() (*unstructured.UnstructuredList, error) {
+func (b *Binder) search() ([]*unstructured.Unstructured, error) {
```
isutton, line 68: Is this change from `*unstructured.UnstructuredList` to `[]*unstructured.Unstructured` essential for this change?

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -465,28 +467,28 @@ func nestedMapComparison(a, b *unstructured.Unstructured, fields ...string) (boo
 // update the list of objects informed as unstructured, looking for "containers" entry. This method
 // loops over each container to inspect "envFrom" and append the intermediary secret, having the same
 // name than original ServiceBindingRequest.
-func (b *Binder) update(objs *unstructured.UnstructuredList) ([]*unstructured.Unstructured, error) {
+func (b *Binder) update(objs []*unstructured.Unstructured) ([]*unstructured.Unstructured, error) {
```
isutton, line 470: Is this change from `*unstructured.UnstructuredList` to `[]*unstructured.Unstructured` essential for this change?

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -501,48 +503,51 @@ func (b *Binder) update(objs *unstructured.UnstructuredList) ([]*unstructured.Un
...
+		updatedObjs = append(updatedObjs, updated)
 	}
 
 	return updatedObjs, nil
 }
 
 // remove attempts to update each given object without any service binding related information.
-func (b *Binder) remove(objs *unstructured.UnstructuredList) error {
-	for _, obj := range objs.Items {
+func (b *Binder) remove(objs []*unstructured.Unstructured) error {
```
isutton, line 526: Is this change from `*unstructured.UnstructuredList` to `[]*unstructured.Unstructured` essential for this change?

pkg/controller/servicebindingrequest/controller.go
```diff
@@ -23,8 +23,8 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager, client dynamic.Interface) (reconcile.Reconciler, error) {
+	mgr.GetClient()
```
isutton, line 26: A commend should state the reason this method is being executed for its side-effect.

lanceliuu: this should be deleted. thanks!

lanceliuu:
This PR replaces listing resource using fieldselector to get method. The latter method returns a single unstructured object instead of unstructuredList. We can definitely construct and return a unstructuredList with single object. But I don't think it is necessary to do the wrap as what we are dealing with is the items in it. 
Besides unstructuredList has its own Kind as it implements runtime.Object though it is not useful for most of the cases. The real API server returns it with kind `List` and the fake client left it empty. So in order to keep consistancy I prefer to return a slice of unstructured object.
Anyway I can revert this change as it won't affect the behaviour of these functions.

isutton:
> This PR replaces listing resource using fieldselector to get method. The latter method returns a single unstructured object instead of unstructuredList. We can definitely construct and return a unstructuredList with single object. But I don't think it is necessary to do the wrap as what we are dealing with is the items in it.
> Besides unstructuredList has its own Kind as it implements runtime.Object though it is not useful for most of the cases. The real API server returns it with kind `List` and the fake client left it empty. So in order to keep consistancy I prefer to return a slice of unstructured object.
> Anyway I can revert this change as it won't affect the behaviour of these functions.

I agree with the change, but as a separate PR.

lanceliuu:
I've revert the changes for unstructuredList in this PR.

pkg/controller/servicebindingrequest/reconciler_test.go
```diff
@@ -222,15 +207,8 @@ func TestReconcilerGenericBinding(t *testing.T) {
...
-
-	fakeClient := f.FakeClient()
-	reconciler := &Reconciler{
-		RestMapper: restMapper,
-		client:     fakeClient,
-		dynClient:  f.FakeDynClient(),
-		scheme:     f.S,
-	}
+	fakeDynClient := f.FakeDynClient()
+	reconciler := &Reconciler{dynClient: fakeDynClient, RestMapper: testutils.BuildTestRESTMapper(), scheme: f.S}
```
Avni-Sharma, line 211: Do we need the RestMapper as an exported value here?

pkg/controller/servicebindingrequest/reconciler_test.go
```diff
@@ -277,30 +253,27 @@ func TestReconcilerGenericBinding(t *testing.T) {
...
+	require.NoError(t, err)
 
+	// Update Credentials
+	s.Data["password"] = []byte("abc123")
+	obj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&s)
+	require.NoError(t, err)
+	updated := unstructured.Unstructured{Object: obj}
+	_, err = fakeDynClient.Resource(secretsGVR).Namespace(updated.GetNamespace()).Update(&updated, metav1.UpdateOptions{})
+	require.NoError(t, err)
+	time.Sleep(1 * time.Second)
```
Avni-Sharma, line 269: Is the sleep necessary here :thinking:

lanceliuu: Remove this will cause flaky as we inject a RFC3339 format timestamp such as "2020-05-14T08:16:42Z" to pod template. For reconciling it might fail to detect template changes since it has only second precision.

pkg/testutils/testutils.go
```diff
@@ -19,5 +19,9 @@ func BuildTestRESTMapper() meta.RESTMapper {
 		schema.GroupVersionKind{Kind: "ConfigMap", Version: "v1"},
 		meta.RESTScopeNamespace,
 	)
+	restMapper.Add(
+		schema.GroupVersionKind{Kind: "Deployment", Version: "v1", Group: "apps"},
```
Avni-Sharma, line 23: should we have one for a deploymentConfig as well?

lanceliuu: It is not required in this PR. But I think it's fine to add that.

isutton: In this particular case, I think it is fine to omit `DeploymentConfig` until we add tests covering its usage.

What do you think?

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -469,24 +468,24 @@ func (b *Binder) update(objs *unstructured.UnstructuredList) ([]*unstructured.Un
 	updatedObjs := []*unstructured.Unstructured{}
 
 	for _, obj := range objs.Items {
-		// store a copy of the original object to later be used in a comparison
-		originalObj := obj.DeepCopy()
+		// modify the copy of the original object and use the original one later for comparison
+		updatedObj := obj.DeepCopy()
```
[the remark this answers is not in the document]
lanceliuu: It is not just about renaming. The original code takes the address obj and return it as result. And the address will be overwritten if there are mutiple objects in objs.Items. So I changes the behavior to modify a copy of obj to avoid the shadowing.

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -28,10 +27,9 @@ const (
 
 // Reconciler reconciles a ServiceBindingRequest object
 type Reconciler struct {
-	client     client.Client     // kubernetes api client
 	dynClient  dynamic.Interface // kubernetes dynamic api client
-	scheme     *runtime.Scheme   // api scheme
 	RestMapper meta.RESTMapper   // restMapper to convert GVK and GVR
+	scheme     *runtime.Scheme   // api scheme
```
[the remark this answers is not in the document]
lanceliuu: Reverted.

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -76,14 +74,17 @@ func (b *Binder) search() (*unstructured.UnstructuredList, error) {
 	}
 
 	var opts metav1.ListOptions
-
```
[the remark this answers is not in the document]
lanceliuu: reverted

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -76,14 +74,17 @@ func (b *Binder) search() (*unstructured.UnstructuredList, error) {
...
-			FieldSelector: fields.Set(fieldName).String(),
+		object, err := b.dynClient.Resource(gvr).Namespace(ns).
+			Get(b.sbr.Spec.ApplicationSelector.ResourceRef, metav1.GetOptions{})
+
+		if err != nil {
+			return nil, err
 		}
+
+		objList := &unstructured.UnstructuredList{Items: []unstructured.Unstructured{*object}}
+		return objList, nil
```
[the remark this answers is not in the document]
lanceliuu: Do you mean by adding function to convert []unstructured.Unstructured to unstructured.UnstructuredList?

pkg/controller/servicebindingrequest/binder.go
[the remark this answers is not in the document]
isutton: Oh, thanks for clarifying!

In this case, it could also be possible to shadow `obj` [with itself](https://play.golang.org/p/TNVUm6RPalH) to avoid renaming in this unit:

```go
obj := obj
```

I'm fine with keeping the change though :)

pkg/controller/servicebindingrequest/binder.go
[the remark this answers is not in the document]
lanceliuu: refactored accordingly.

PR #489, opened 2020-05-26
[Fix #447]  correctly detect multiple secret created by a service operator

### LINKED ISSUE #477: pkg/controller/servicebindingrequest/service.go bug
I'm building a POC with Strimzi binding. I defined a service binding request thus:
```
  ...
  backingServiceSelectors:
  - group: kafka.strimzi.io
    kind: KafkaUser
    resourceRef: rcopoc
    version: v1beta1
    namespace: poc-rco
  ...
```
I modified the Kafka User Secret to be owned by the Kafka User:
```
$ oc get secret rcopoc -o yaml
apiVersion: v1
data:
  password: UnlPM0QyeDZzcG9i
kind: Secret
metadata:
  ...
  ownerReferences:
  - apiVersion: kafka.strimzi.io/v1beta1
    blockOwnerDeletion: false
    controller: false
    kind: KafkaUser
    name: rcopoc
    uid: 0c407fe4-6ce5-4b88-bc98-50c6b0b7f458
  ...
```
Goal is to have the binding service operator process that secret because it is owned by the kafka user. No matter what I tried, the binding would not find the secret. Stepping through the binding operator, I found code in `pkg/controller/servicebindingrequest/service.go`:
```
157 func getOwnedResources(
158 	client dynamic.Interface,
159 	ns string,
160 	gvk schema.GroupVersionKind,
161 	name string,
162 	uid types.UID,
163 ) (
164 	[]*unstructured.Unstructured,
165 	error,
166 ) {
167 	var resources []*unstructured.Unstructured
168 	for _, br := range bindableResources {
169 		lst, err := client.Resource(br.gvr).Namespace(ns).List(metav1.ListOptions{})
170 		if err != nil {
171 			return resources, err
172 		}
173 		for _, item := range lst.Items {
174 			owners := item.GetOwnerReferences()
175 			for _, owner := range owners {
176 				if owner.UID == uid {
177 					resources = append(resources, &item)
178 				}
179 			}
180 		}
181 	}
182 	return resources, nil
183 }
```
The issue appears to me to be line 177. The `item` var is overwritten by the range statement. This one-file  program illustrates the issue:
```
package main

import "fmt"

func main() {
	items := []string{"1", "2", "3"}
	resources := []*string{}
	for _, item := range items {
		if item == "2" {
			resources = append(resources, &item)
		}
	}
	if *resources[0] != "2" {
		fmt.Printf("ERROR: Expected '2', got %v\n", *resources[0])
	}
}
```
Program outputs: `ERROR: Expected '2', got 3`

akashshinde:
### Motivation
Fixes: #477 

### Changes

currently secret is being added copy by reference of `&item` var, which is being changed after each for loop iteration,
so eventually resource slice will have same address elements. 
```
for _, item := range lst.Items {
  if owner.UID == uid {
     resources = append(resources, &item)
  }
}
```

This is being changed in PR where addresses passed to slice would be different.
```
for idx, _ := range lst.Items {
  if owner.UID == uid {
      resources = append(resources, &lst.Items[idx])
  }
}
```

### Testing
Run all tests with `make test-unit`

### DIFF
--- a/pkg/controller/servicebindingrequest/service.go
+++ b/pkg/controller/servicebindingrequest/service.go
@@ -170,11 +170,11 @@ func getOwnedResources(
 		if err != nil {
 			return resources, err
 		}
-		for _, item := range lst.Items {
+		for idx, item := range lst.Items {
 			owners := item.GetOwnerReferences()
 			for _, owner := range owners {
 				if owner.UID == uid {
-					resources = append(resources, &item)
+					resources = append(resources, &lst.Items[idx])
 				}
 			}
 		}
--- a/pkg/controller/servicebindingrequest/servicecontext_test.go
+++ b/pkg/controller/servicebindingrequest/servicecontext_test.go
@@ -128,46 +128,69 @@ func TestFindOwnedResourcesCtxs_ConfigMap(t *testing.T) {
 	})
 }
 
-func TestFindOwnedResourcesCtxs_Secret(t *testing.T) {
-	f := mocks.NewFake(t, "test")
-	cr := mocks.DatabaseCRMock("test", "test")
-	reference := metav1.OwnerReference{
-		APIVersion:         cr.APIVersion,
-		Kind:               cr.Kind,
-		Name:               cr.Name,
-		UID:                cr.UID,
-		Controller:         &trueBool,
-		BlockOwnerDeletion: &trueBool,
+func TestFindOwnedResourcesCtxs_Secrets(t *testing.T) {
+	testCases := []struct {
+		desc    string
+		secrets []string
+	}{
+		{
+			desc:    "backend cr creating only one secret should returns only one child",
+			secrets: []string{"test_database"},
+		},
+		{
+			desc:    "backend cr creating multiple secrets should returns only multiple children",
+			secrets: []string{"test_database", "test_database2"},
+		},
 	}
-	secret := mocks.SecretMock("test", "test_database")
-	us := &unstructured.Unstructured{}
-	uc, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&secret)
-	require.NoError(t, err)
-	us.Object = uc
-	us.SetOwnerReferences([]metav1.OwnerReference{reference})
-	route, err := runtime.DefaultUnstructuredConverter.ToUnstructured(mocks.RouteCRMock("test", "test"))
-	require.NoError(t, err)
-	usRoute := &unstructured.Unstructured{Object: route}
-	usRoute.SetOwnerReferences([]metav1.OwnerReference{reference})
-	f.S.AddKnownTypes(pgv1alpha1.SchemeGroupVersion, &pgv1alpha1.Database{})
-	f.S.AddKnownTypes(routev1.SchemeGroupVersion, &routev1.Route{})
-	f.AddMockResource(cr)
-	f.AddMockResource(us)
-	f.AddMockResource(&unstructured.Unstructured{Object: route})
 
-	restMapper := testutils.BuildTestRESTMapper()
+	for _, tC := range testCases {
+		t.Run(tC.desc, func(t *testing.T) {
+			f := mocks.NewFake(t, "test")
+			cr := mocks.DatabaseCRMock("test", "test")
+			reference := metav1.OwnerReference{
+				APIVersion:         cr.APIVersion,
+				Kind:               cr.Kind,
+				Name:               cr.Name,
+				UID:                cr.UID,
+				Controller:         &trueBool,
+				BlockOwnerDeletion: &trueBool,
+			}
+			route, err := runtime.DefaultUnstructuredConverter.ToUnstructured(mocks.RouteCRMock("test", "test"))
+			require.NoError(t, err)
+			usRoute := &unstructured.Unstructured{Object: route}
+			usRoute.SetOwnerReferences([]metav1.OwnerReference{reference})
+			f.S.AddKnownTypes(pgv1alpha1.SchemeGroupVersion, &pgv1alpha1.Database{})
+			f.S.AddKnownTypes(routev1.SchemeGroupVersion, &routev1.Route{})
+			f.AddMockResource(cr)
+			f.AddMockResource(&unstructured.Unstructured{Object: route})
 
-	t.Run("existing selectors", func(t *testing.T) {
-		ownedResourcesCtxs, err := findOwnedResourcesCtxs(
-			f.FakeDynClient(),
-			cr.GetNamespace(),
-			cr.GetName(),
-			cr.GetUID(),
-			cr.GroupVersionKind(),
-			nil,
-			restMapper,
-		)
-		require.NoError(t, err)
-		require.NotEmpty(t, ownedResourcesCtxs)
-	})
+			restMapper := testutils.BuildTestRESTMapper()
+
+			for _, secret := range tC.secrets {
+				secret := mocks.SecretMock("test", secret)
+				us := &unstructured.Unstructured{}
+				uc, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&secret)
+				require.NoError(t, err)
+				us.Object = uc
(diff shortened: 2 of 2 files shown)


pkg/controller/servicebindingrequest/service.go
```diff
@@ -170,11 +170,11 @@ func getOwnedResources(
 		if err != nil {
 			return resources, err
 		}
-		for _, item := range lst.Items {
+		for idx, item := range lst.Items {
```
pedjak, line 173: The change fixes truly a bug, it does not introduce a new behaviour, more details about it on this [stackoverflow answer](https://stackoverflow.com/a/48826629). As stated there, the proposed solution might have an issue with garbage collection, and perhaps we could solve slightly different:

```diff
diff --git a/pkg/controller/servicebindingrequest/service.go b/pkg/controller/servicebindingrequest/service.go
index 9895a94..6d79132 100644
--- a/pkg/controller/servicebindingrequest/service.go
+++ b/pkg/controller/servicebindingrequest/service.go
@@ -161,10 +161,10 @@ func getOwnedResources(
 	name string,
 	uid types.UID,
 ) (
-	[]*unstructured.Unstructured,
+	[]unstructured.Unstructured,
 	error,
 ) {
-	var resources []*unstructured.Unstructured
+	var resources []unstructured.Unstructured
 	for _, br := range bindableResources {
 		lst, err := client.Resource(br.gvr).Namespace(ns).List(metav1.ListOptions{})
 		if err != nil {
@@ -174,7 +174,7 @@ func getOwnedResources(
 			owners := item.GetOwnerReferences()
 			for _, owner := range owners {
 				if owner.UID == uid {
-					resources = append(resources, &item)
+					resources = append(resources, item)
 				}
 			}
 		}
@@ -202,7 +202,7 @@ func buildOwnedResourceContext(
 
 func buildOwnedResourceContexts(
 	client dynamic.Interface,
-	objs []*unstructured.Unstructured,
+	objs []unstructured.Unstructured,
 	ownerEnvVarPrefix *string,
 	restMapper meta.RESTMapper,
 ) ([]*ServiceContext, error) {
@@ -215,7 +215,7 @@ func buildOwnedResourceContexts(
 			}
 			svcCtx, err := buildOwnedResourceContext(
 				client,
-				obj,
+				&obj,
 				ownerEnvVarPrefix,
 				restMapper,
 				br.inputPath,


```

pkg/controller/servicebindingrequest/servicecontext_test.go
```diff
@@ -169,5 +169,32 @@ func TestFindOwnedResourcesCtxs_Secret(t *testing.T) {
 		)
 		require.NoError(t, err)
 		require.NotEmpty(t, ownedResourcesCtxs)
+		require.EqualValues(t, 1, len(ownedResourcesCtxs))
+		require.Equal(t, "test_database", ownedResourcesCtxs[0].Service.Object["metadata"].(map[string]interface{})["name"].(string))
 	})
+
+	t.Run("multiple secrets", func(t *testing.T) {
```
pedjak, line 176: since we are adding a new test here, we could definitely make some nice table tests, to avoid code copy/pasting.

pkg/controller/servicebindingrequest/service.go
pedjak: as a followup, we should inspect the code in order to see if we have the same issue elsewhere.

sbose78:
@akashshinde 

> Testing
Run all tests with make test-unit

Note, this testing section is supposed to let the PR author specify how to test this 'as a user' :)

pkg/controller/servicebindingrequest/servicecontext_test.go
```diff
@@ -169,5 +169,32 @@ func TestFindOwnedResourcesCtxs_Secret(t *testing.T) {
 		)
 		require.NoError(t, err)
 		require.NotEmpty(t, ownedResourcesCtxs)
+		require.EqualValues(t, 1, len(ownedResourcesCtxs))
+		require.Equal(t, "test_database", ownedResourcesCtxs[0].Service.Object["metadata"].(map[string]interface{})["name"].(string))
```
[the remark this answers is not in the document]
akashshinde: Done

pkg/controller/servicebindingrequest/servicecontext_test.go
```diff
@@ -169,5 +169,32 @@ func TestFindOwnedResourcesCtxs_Secret(t *testing.T) {
...
+			cr.GetName(),
+			cr.GetUID(),
+			cr.GroupVersionKind(),
+			nil,
+			restMapper,
+		)
+		require.NoError(t, err)
+		require.NotEmpty(t, ownedResourcesCtxs)
+		require.EqualValues(t, 2, len(ownedResourcesCtxs))
+		require.Equal(t, "test_database", ownedResourcesCtxs[0].Service.Object["metadata"].(map[string]interface{})["name"].(string))
```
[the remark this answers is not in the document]
akashshinde: Done

pkg/controller/servicebindingrequest/servicecontext_test.go
```diff
@@ -169,5 +169,32 @@ func TestFindOwnedResourcesCtxs_Secret(t *testing.T) {
...
+		require.Equal(t, "test_database2", ownedResourcesCtxs[1].Service.Object["metadata"].(map[string]interface{})["name"].(string))
```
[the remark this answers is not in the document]
akashshinde: Done

pkg/controller/servicebindingrequest/servicecontext_test.go
akashshinde: @pedjak I was wondering how would we have table driven tests for this scenario since both the tests has different test logic. implementation of table driven tests should be there when tests uses same logic and but different parameters.

pkg/controller/servicebindingrequest/service.go
akashshinde: raised jira ticket https://issues.redhat.com/browse/APPSVC-611

pkg/controller/servicebindingrequest/servicecontext_test.go
pedjak: well,  the setup for both tests are pretty much similar, the same steps are created for both secrets:
```
secret := mocks.SecretMock("test", "test_database")
	us := &unstructured.Unstructured{}
	uc, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&secret)
	require.NoError(t, err)
	us.Object = uc
	us.SetOwnerReferences([]metav1.OwnerReference{reference})
	route, err := runtime.DefaultUnstructuredConverter.ToUnstructured(mocks.RouteCRMock("test", "test"))
	require.NoError(t, err)
	usRoute := &unstructured.Unstructured{Object: route}
	usRoute.SetOwnerReferences([]metav1.OwnerReference{reference})
	f.S.AddKnownTypes(pgv1alpha1.SchemeGroupVersion, &pgv1alpha1.Database{})
	f.S.AddKnownTypes(routev1.SchemeGroupVersion, &routev1.Route{})
	f.AddMockResource(cr)
	f.AddMockResource(us)
	f.AddMockResource(&unstructured.Unstructured{Object: route})
```

in the new test you have in essence copied the above block and renamed a few variables, but is really different is that the first tests has as input one secret named `test_database`, and the new test has two secrets. The above block could be iterated of a list of input params. Hence, the table test struct could look like:

```
tests := []struct {
		name                string
		secretNames    []string
	}{
  name: "existing selector",
  secretNames: { "test_database" },
},
{
  name: "multiple secrets",
  secretNames: { "test_database", "test_database2" },
}
```

akashshinde: Thanks @pedjak makes sense 
pushed changes accordingly here https://github.com/redhat-developer/service-binding-operator/pull/489/commits/5e0e1aa1a283fdd224024b08bca0750166d397fc

akashshinde:
@isutton  @pedjak need lgtm

Issue #419, opened 2020-04-10
Once the selected application is deleted, the sbr can't be deleted


cdlliuy:
After a sbr created successfully,  I deleted the selected application directly.

```
spec:
  applicationSelector:
    group: serving.knative.dev
    resource: services
    resourceRef: mydemo
    version: v1
.....
```

That means I deleted `mydemo`. 

Then, I realized and tried to delete sbr , but I can't delete it at all.

```
 >> kubectl get sbr
NAME                    AGE
mydemo-bindingrequest   3h24m
>>  kubectl delete sbr mydemo-bindingrequest
servicebindingrequest.apps.openshift.io "mydemo-bindingrequest" deleted
(Then it hangs here ...)




^C
>> $ kubectl get sbr
NAME                    AGE
mydemo-bindingrequest   3h24m
```

Any comment on this?  Expected behavior? 

It makes me to think about whether the selected `application` should be the owner of the sbr , since the sbr is only serving for the specific application only.   If  the ownership is set,  the sbr can be deleted when the application is removed.

sbose78:
I think the finalizer we create has a bug?

Avni-Sharma:
> After a sbr created successfully, I deleted the selected application directly.
> 
> ```
> spec:
>   applicationSelector:
>     group: serving.knative.dev
>     resource: services
>     resourceRef: mydemo
>     version: v1
> .....
> ```
> 
> That means I deleted `mydemo`.
> 
> Then, I realized and tried to delete sbr , but I can't delete it at all.
> 
> ```
>  >> kubectl get sbr
> NAME                    AGE
> mydemo-bindingrequest   3h24m
> >>  kubectl delete sbr mydemo-bindingrequest
> servicebindingrequest.apps.openshift.io "mydemo-bindingrequest" deleted
> (Then it hangs here ...)
> 
> 
> 
> 
> ^C
> >> $ kubectl get sbr
> NAME                    AGE
> mydemo-bindingrequest   3h24m
> ```
> 
> Any comment on this? Expected behavior?
> 
> Furthermore, I created `mydemo` again, SBR didn't reconcile to have the secrets injected into `mydemo` container anymore. Anyway, with the recreation of `application resourceRef`, I can delete sbr.
> 
> It makes me to think about whether the selected `application` should be the owner of the sbr , since the sbr is only serving for the specific application only. If the ownership is set, the sbr can be deleted when the application is removed.

[...]

PR #498, opened 2020-05-28
Split BindingReady into CollectionReady and InjectionReady

### LINKED ISSUE #381: Generic binding : BindingFail status is reported
Even though the binding secret gets created fine, the status is marked as `BindingFail`

```
kind: ServiceBindingRequest
metadata:
  creationTimestamp: '2020-03-12T21:05:44Z'
  generation: 1
  name: node-todo-git
  namespace: sbose
  resourceVersion: '221728'
  selfLink: >-
    /apis/apps.openshift.io/v1alpha1/namespaces/sbosse/servicebindingrequests/node-todo-git
  uid: 3413f95b-7cd6-44b7-9a62-daaec834be18
spec:
  backingServiceSelectors:
    - group: etcd.database.coreos.com
      kind: EtcdCluster
      resourceRef: example
      version: v1beta2
  detectBindingResources: true
status:
  bindingStatus: BindingFail
  conditions:
    - lastHeartbeatTime: '2020-03-12T21:05:59Z'
      lastTransitionTime: '2020-03-12T21:05:45Z'
      message: application ResourceRef or MatchLabel not found
      reason: BindingFail
      status: 'False'
      type: Ready
  secret: node-todo-git
```

DhritiShikhar:
### Motivation

To fix  #381 

### Changes

Splits the `BindingReady` type into more specific types:
- CollectionReady which would be related to the collection and persistence of intermediate manifests (secrets, config maps, etc)
- InjectionReady which would be related to change the applications' manifests to use those intermediate manifests:

### How to test

- Import an application
- Create a DB instance 

```
kind: ServiceBindingRequest
metadata:
  name: binding-request
spec:
 backingServiceSelector:
    group: postgresql.baiju.dev
    version: v1alpha1
    kind: Database
    resourceRef: db-demo
```

3. Use the following command to check if the status has been updated:

The output should have `status.conditions.type` as 
- `CollectionReady` with `status: "True"`
-  `InjectionReady` with `status: "False"`

```
status:
  conditions:
  - lastHeartbeatTime: 2020-04-23T06:25:16Z
    lastTransitionTime: 2020-04-23T06:25:16Z
    message: application ResourceRef or MatchLabel not found
    reason: EmptyApplicationSelector
    status: "True"
    type: CollectionReady
  - lastHeartbeatTime: 2020-04-23T06:25:16Z
    lastTransitionTime: 2020-04-23T06:25:16Z
    message: application ResourceRef or MatchLabel not found
    reason: EmptyApplicationSelector
    status: "False"
    type: InjectionReady
```

## Demo

https://youtu.be/NI6Fd5l65rw


pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -258,10 +277,55 @@ func (b *ServiceBinder) Bind() (reconcile.Result, error) {
...
+		}
+
+		// appending finalizer, should be later removed upon resource deletion
+		sbr.SetFinalizers(append(removeStringSlice(b.SBR.GetFinalizers(), Finalizer), Finalizer))
+		if _, err = b.updateServiceBindingRequest(sbr); err != nil {
+			return NoRequeue(err)
+		}
+
+		b.Logger.Info(ErrEmptyApplicationSelector.Error())
+		return Done()
```
isutton, line 305: This block deserves its own function:

```go
if isApplicationSelectorEmpty(b.SBR.Spec.ApplicationSelector) {
        return b.handleApplicationSelectorEmpty(b.SBR)
}
```

This modification has two important benefits:

* Introduces more coverage by making it easier to cover the intended use cases; and
* Provide supporting material in further refactoring and simplifications.

What do you think?

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -258,10 +277,55 @@ func (b *ServiceBinder) Bind() (reconcile.Result, error) {
...
+				Message: ErrApplicationNotFound.Error(),
+			})
+
+			// appending finalizer, should be later removed upon resource deletion
+			b.SBR.SetFinalizers(append(removeStringSlice(b.SBR.GetFinalizers(), Finalizer), Finalizer))
+			if _, updateErr := b.updateServiceBindingRequest(b.SBR); updateErr != nil {
+				return NoRequeue(updateErr)
+			}
+		}
+		return b.onError(err, b.SBR, sbrStatus, nil)
```
isutton, line 328: On similar note as another comment, this block might deserve its own function (sharing behavior with the `handleApplicationSelectorEmpty` method).

pkg/controller/servicebindingrequest/servicebinder_test.go
```diff
@@ -103,29 +103,27 @@ func TestServiceBinder_Bind(t *testing.T) {
...
+					if len(c.Reason) > 0 {
+						expected.Reason = c.Reason
+						got.Reason = sb.SBR.Status.Conditions[i].Reason
+					}
+					if len(c.Message) > 0 {
+						expected.Message = c.Message
+						got.Message = sb.SBR.Status.Conditions[i].Message
+					}
+
+					require.Equal(t, expected, got)
```
isutton, line 126: Can you elaborate the reason of this change?

On another note, I have the impression `conditionsv1.IsStatusConditionPresentAndEqual` could be used here instead; what do you think?

test/e2e/servicebindingrequest_test.go
```diff
@@ -186,6 +187,7 @@ func assertDeploymentEnvFrom(
 // assertSBRStatus will determine if SBR is on "success" state
 func assertSBRStatus(
 	ctx context.Context,
+	t *testing.T,
```
isutton, line 190: Test functions should always receive `t` as first argument.

```suggestion
	t *testing.T,
	ctx context.Context,
```

test/e2e/servicebindingrequest_test.go
```diff
@@ -450,7 +461,7 @@ func inspectSBRStatus(
 ) {
 	err := retry(10, 5*time.Second, func() error {
 		t.Logf("Inspecting SBR: '%s'", namespacedName)
-		err := assertSBRStatus(ctx, f, namespacedName)
+		err := assertSBRStatus(ctx, t, f, namespacedName)
```
isutton, line 464: Test functions should always receive `t` as first argument.

```suggestion
		err := assertSBRStatus(t, ctx, f, namespacedName)
```

test/e2e/servicebindingrequest_test.go
isutton: Please disregard this comment; `ctx` should always be the first argument per https://github.com/golang/go/wiki/CodeReviewComments#contexts.

test/e2e/servicebindingrequest_test.go
isutton: Please disregard this comment; `ctx` should always be the first argument per https://github.com/golang/go/wiki/CodeReviewComments#contexts.

pkg/controller/servicebindingrequest/servicebinder_test.go
DhritiShikhar: I kind of amended the existing code to check against multiple conditions.

But yes, I will re-use `requireConditionPresentAndTrue` here.

pkg/controller/servicebindingrequest/reconciler_test.go
```diff
@@ -272,8 +317,26 @@ func TestReconcilerGenericBinding(t *testing.T) {
...
-	require.Equal(t, corev1.ConditionTrue, sbrOutput2.Status.Conditions[0].Status)
+	require.True(t,
+		conditionsv1.IsStatusConditionPresentAndEqual(
+			sbrOutput2.Status.Conditions,
+			CollectionReady,
+			corev1.ConditionTrue,
+		),
+		"CollectionReady condition should exist and true; existing conditions: %+v",
+		sbrOutput2.Status.Conditions,
+	)
```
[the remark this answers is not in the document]
DhritiShikhar: Yes. This should be modular. I will make this change.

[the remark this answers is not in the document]
DhritiShikhar: Done here 88c15c5

pkg/controller/servicebindingrequest/servicebinder_test.go
DhritiShikhar: Done here 88c15c5

pkg/controller/servicebindingrequest/servicebinder.go
DhritiShikhar: Done 1b275f9

pkg/controller/servicebindingrequest/servicebinder.go
DhritiShikhar: Agreed.

Application Not Found and Empty Application scenario should be handled in different function.

Done 1b275f9

DhritiShikhar:
Will rebase the PR once PR review is done. Is that okay?

DhritiShikhar:
Would rebase the PR in the end.

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -246,6 +251,77 @@ func (b *ServiceBinder) onError(
...
+
+	b.Logger.Info(ErrEmptyApplicationSelector.Error())
+	return Done()
+}
+
+// handleApplicationSelectorEmpty handles scenario when application is not found.
+func (b *ServiceBinder) handleApplicationNotFound() (reconcile.Result, error) {
+	conditionsv1.SetStatusCondition(&b.SBR.Status.Conditions, conditionsv1.Condition{
+		Type:   CollectionReady,
+		Status: corev1.ConditionTrue,
```
Avni-Sharma, line 302: I believe the secret should be updated too https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/apis/apps/v1alpha1/servicebindingrequest_types.go#L55 in case of generic binding, or we can have the secret name in Message. If we add in Message then would we need status.secret ?

DhritiShikhar: @Avni-Sharma 

[1] Secret is handled here -> https://github.com/redhat-developer/service-binding-operator/pull/498/files#diff-fae660884df4e35af60a04c5ddf930daR335

[2] I don't see a point in adding a secret name in Condition's message when we already have it under status.secret.

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -246,6 +251,77 @@ func (b *ServiceBinder) onError(
...
+		application.ResourceRef == "" &&
+			application.LabelSelector.MatchLabels == nil {
+		return true
+	}
+	return false
+}
+
+// handleApplicationSelectorEmpty handles scenario when applicationSelector is not declared in
+// the Service Binding Request.
+func (b *ServiceBinder) handleApplicationSelectorEmpty() (reconcile.Result, error) {
```
isutton, line 270: This method is exactly the same as `handleApplicationNotFound`, with the only difference being what is returned at the end in case of success: either `Done()` as in the case of `handleApplicationSelectorEmpty` or `RequeueOnNotFound()` as in the case of `handleApplicationNotFound`.

I believe `handleApplicationNotFound` can consume the `handleApplicationSelectorEmpty` method and then deciding whether the work item should be retried:

```go
func (b *ServiceBinder) handleApplicationNotFound() (reconcile.Result, error) {
    // it is safe to ignore the first argument since a new result will be generated based 
    // on the propagated error
    _, err := b.handleApplicationSelectorEmpty();
   return RequeueOnNotFound(err)
}
```

DhritiShikhar: [1] method `handleApplicationSelectorEmpty()` and `handleApplicationNotFound()` are not exactly similar. They differ in two respect:

- reason and message of `InjectionReady` condition type is different
- what is returned at the end - Done() or RequeueOnNotFound()

[2] Based on your suggestion, If I make `handleApplicationNotFound` consume the `handleApplicationSelectorEmpty` method, then
I would have to remove these 3 lines -> https://github.com/redhat-developer/service-binding-operator/pull/498/files#diff-fae660884df4e35af60a04c5ddf930daR337-R339
Right?
that would mean that handleApplicationSelectorEmpty is called AFTER we're done searching the application -> https://github.com/redhat-developer/service-binding-operator/pull/498/files#diff-fae660884df4e35af60a04c5ddf930daR340

If the user hasn't specified any application in SBR, why should there be any search for an application at all?

isutton: To make it clearer: I wouldn't change the existing flow so I would keep the check you've mentioned in [2] before calling `Bind`.

On the other hand, those methods are almost identical with the exception of those two values mentioned in [1] that can (and should) be parameterized in the base function itself, regardless if it is `handleApplicationSelectorEmpty` or another method you choose to extract the base function.

I think this isn't a biggie for now if we don't want to pursue this, only seemed a good improvement and easy to tackle now :)

DhritiShikhar: I get your point now. I will make changes.

DhritiShikhar: @isutton Is this what you had in mind? -> bd70eaa

pkg/controller/servicebindingrequest/servicebinder.go
Avni-Sharma: The status.Secret?

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -129,13 +137,16 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-			Status:  corev1.ConditionFalse,
-			Reason:  EmptyServiceSelectorsReason,
-			Message: "The spec.backingServiceSelectors field is empty.",
-		})
-		_, updateErr := updateServiceBindingRequestStatus(r.dynClient, sbr)
+		_, updateErr := updateServiceBindingRequestStatus(
+			r.dynClient,
+			sbr,
+			conditionsv1.Condition{
+				Type:    InjectionReady,
```
Avni-Sharma, line 144: Should we have collection Ready and injection ready both as false in case of len(selectors) ==0 because ideally when we don't have any service then no collection is happening.

pkg/controller/servicebindingrequest/reconciler_test.go
```diff
@@ -212,22 +240,15 @@ func TestReconcilerGenericBinding(t *testing.T) {
 
 	// Reconcile without deployment
 	res, err := reconciler.Reconcile(reconcileRequest())
-	require.NoError(t, err)
+	require.EqualError(t, err, ErrApplicationNotFound.Error())
```
Avni-Sharma, line 243: Are we requeing when the application is not found, i.e application was there but then deleted, i.e I initially had an applicationSelector but now my deployment was deleted. If that happens then we will keep falling in requeue and this error will exist https://github.com/redhat-developer/service-binding-operator/issues/419

pkg/controller/servicebindingrequest/reconciler.go
DhritiShikhar: +1 I will add it.

pkg/controller/servicebindingrequest/reconciler_test.go
DhritiShikhar: I will look into this

pkg/controller/servicebindingrequest/reconciler.go
DhritiShikhar: Is this okay @Avni-Sharma 9b395a8 ?

Avni-Sharma: looks good :+1:

pkg/controller/servicebindingrequest/reconciler_test.go
DhritiShikhar: @Avni-Sharma review 7889bfa

pkg/controller/servicebindingrequest/servicebinder.go
isutton: Pretty much 👍 

I believe the whole error handling could be simplified even further but for the scope of this PR naming the method handling application errors sensibly should be as far as it goes, so I propose to call the method `handleApplicationError` instead of `handleApplication` as last small change request 😅 What do you think?

DhritiShikhar: @isutton It's not really an `Error` scenario, is it?
If something is not found, it's not actually an error scenario.

DhritiShikhar: Thinking again, maybe it's okay to rename it `handleApplicationError` for now since we already have `ErrEmptyApplicationSelector` and `ErrApplicationNotFound`.

DhritiShikhar: DOne 99272e7

DhritiShikhar:
Based on conversation with Predrag,
[1] An E2E test for the ApplicationNotFound scenario

DhritiShikhar:
@Avni-Sharma @isutton Should secret be deleted in `applicationNotFound` scenario?

DhritiShikhar:
Discussions by me and @isutton about this PR is captured here 
https://docs.google.com/document/d/1FlvJjE9UbgVsgxcraJNx16lJCIbP9YMCFsgcTzxYuwE

DhritiShikhar:
> Only the Service Binding Operator related finalizers should be removed from the Service Binding Request.

Once we have #524 , no need to address this comment, right @isutton ?

DhritiShikhar:
Will rebase this PR after #524 is merged.

pedjak:
> Based on conversation with Predrag,
[1] An E2E test for the ApplicationNotFound scenario

@DhritiShikhar what is the expected behaviour? https://github.com/redhat-developer/service-binding-operator/blob/master/test/e2e/servicebindingrequest_test.go#L607 covers only the case when SBR gets removed.

DhritiShikhar:
I am supposed to add an e2e test for "ApplicationNotFound" scenario in this PR.

Based on the conversation in standup today:
A new framework for acceptance tests is in progress. This is the PR for the same -> https://github.com/redhat-developer/service-binding-operator/pull/541

The [current e2e test scenarios](https://github.com/redhat-developer/service-binding-operator/blob/master/test/e2e/servicebindingrequest_test.go#L86) will be migrated to the new acceptance tests.

Writing an e2e for "ApplicationNotFound" scenario and then migrating it again based on the new acceptance tests would be redundant.

I propose:
[1] Let's create an issue for adding "ApplicationNotFound" acceptance test which can be taken up after we have this PR in -> https://github.com/redhat-developer/service-binding-operator/pull/541
[2] Meanwhile, let's merge this split binding PR without the "ApplicationNotFound" e2e test.

I need the team's opinion on this:

@pedjak @sbose78 @isutton @baijum @Avni-Sharma  @pmacik

sbose78:
+1 on the proposal, Dhriti.

pmacik:
@DhritiShikhar I aggree with the proposal, as discussed. It is not worthy spending time on implementing the test that would be converted to the new framework eventually, anyway. Lets create the followup task to add the e2e test as part of the acceptance suite as soon as the suite is in place.

Avni-Sharma:
+1 , instead of rewriting we can instead focus on the #541 for a speedy merge

isutton:
> I am supposed to add an e2e test for "ApplicationNotFound" scenario in this PR.
> 
> Based on the conversation in standup today:
> A new framework for acceptance tests is in progress. This is the PR for the same -> #541
> 
> The [current e2e test scenarios](https://github.com/redhat-developer/service-binding-operator/blob/master/test/e2e/servicebindingrequest_test.go#L86) will be migrated to the new acceptance tests.
> 
> Writing an e2e for "ApplicationNotFound" scenario and then migrating it again based on the new acceptance tests would be redundant.

In the case the "ApplicationNotFound" reconciliation scenario is being exercised in unit tests and there's a reconciliation test covering this case (which from my understanding there are), I'm fine postponing this E2e test until we have the acceptance tests infra-structure.

DhritiShikhar:
I created an issue for adding e2e test for "ApplicationNotFound" scenario -> https://github.com/redhat-developer/service-binding-operator/issues/547

I have also rebased this PR.

@isutton @Avni-Sharma Can you please re-review?

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -236,40 +241,102 @@ func (b *serviceBinder) onError(
...
+	})
+
+	// updating status of request instance
+	sbr, err := b.updateStatusServiceBindingRequest(b.sbr, &b.sbr.Status)
+	if err != nil {
+		return requeueOnConflict(err)
+	}
+
+	// appending finalizer, should be later removed upon resource deletion
+	sbr.SetFinalizers(append(removeStringSlice(b.sbr.GetFinalizers(), finalizer), finalizer))
```
Avni-Sharma, line 280: Not sure whether Finalisers should be set here. I think it is already being set here https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/controller/servicebindingrequest/servicebinder.go#L278

DhritiShikhar: @Avni-Sharma 
In case of applicationNotFound, the control doesn't reach there.

pratikjagrut:
Here 4.5-e2e is failed due to #549

test/e2e/servicebindingrequest_test.go
```diff
@@ -581,7 +504,7 @@ func serviceBindingRequestTest(
 	t.Log("Inspecting SBR status...")
 	require.Eventually(t, func() bool {
 		t.Logf("Inspecting SBR: '%s'", sbrNamespacedName)
-		err := assertSBRStatus(todoCtx, f, sbrNamespacedName)
+		_, err := assertSBRSecret(todoCtx, f, intermediarySecretNamespacedName, assertKeys)
```
DhritiShikhar, line 507: fixed this in next commit

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -236,40 +241,102 @@ func (b *serviceBinder) onError(
...
 func (b *serviceBinder) bind() (reconcile.Result, error) {
-	sbrStatus := b.sbr.Status.DeepCopy()
-
 	b.logger.Info("Saving data on intermediary secret...")
 
 	secretObj, err := b.secret.createOrUpdate(b.envVars, b.sbr.AsOwnerReference())
 	if err != nil {
 		b.logger.Error(err, "On saving secret data..")
-		return b.onError(err, b.sbr, sbrStatus, nil)
+		return b.onError(err, b.sbr, &b.sbr.Status, nil)
```
isutton, line 305: It seems that all usages of `b.onError` are using a reference to `b.sbr.Status`, so perhaps that parameter can be removed?

```go
return b.onError(err, b.sbr, nil)
```

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -236,40 +241,102 @@ func (b *serviceBinder) onError(
...
+	b.sbr.Status.Secret = secretObj.GetName()
 
+	if isApplicationSelectorEmpty(b.sbr.Spec.ApplicationSelector) {
+		return b.handleApplicationError(EmptyApplicationSelectorReason, errEmptyApplicationSelector)
+	}
 	updatedObjects, err := b.binder.bind()
 	if err != nil {
 		b.logger.Error(err, "On binding application.")
-		return b.onError(err, b.sbr, sbrStatus, updatedObjects)
+		if errors.Is(err, errApplicationNotFound) {
```
isutton, line 315: I believe @qibobo has mentioned something about this assertion not working as expected, could you double check whether this works as expected?

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -236,40 +241,102 @@ func (b *serviceBinder) onError(
...
+		Status: corev1.ConditionTrue,
+	})
+	conditionsv1.SetStatusCondition(&b.sbr.Status.Conditions, conditionsv1.Condition{
+		Type:   InjectionReady,
 		Status: corev1.ConditionTrue,
 	})
 
 	// updating status of request instance
-	sbr, err := b.updateStatusServiceBindingRequest(b.sbr, sbrStatus)
+	sbr, err := b.updateStatusServiceBindingRequest(b.sbr, &b.sbr.Status)
```
isutton, line 339: Please note that `updateStatusServiceBindingRequest` compares the first argument's `Status` field against the second argument's value before triggering the update; this change makes the comparison return always true, so the update won't be processed AFAICS.

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -279,7 +346,6 @@ func (b *serviceBinder) bind() (reconcile.Result, error) {
 	if _, err = b.updateServiceBindingRequest(sbr); err != nil {
 		return noRequeue(err)
```
isutton, line 347: The sbr should be requeued on any update error, otherwise the removal of the finalizers above won't be available in the API server. The side effect of this is the garbage collector won't be able to remove the resource.

```go
return requeueError(err)
```

pkg/controller/servicebindingrequest/servicebinder.go
DhritiShikhar: That can be done. It's a good idea. 
But the intention of raising this PR was not to refactor `b.onError(..)`
We can create an issue for that and tackle it in another PR.

pkg/controller/servicebindingrequest/servicebinder.go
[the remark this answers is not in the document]
DhritiShikhar: Done 69fb48e

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -236,40 +241,102 @@ func (b *serviceBinder) onError(
...
+	// appending finalizer, should be later removed upon resource deletion
+	sbr.SetFinalizers(append(removeStringSlice(b.sbr.GetFinalizers(), finalizer), finalizer))
+	if _, err = b.updateServiceBindingRequest(sbr); err != nil {
+		return noRequeue(err)
+	}
+
+	b.logger.Info(applicationError.Error())
+
+	if errors.Is(applicationError, errApplicationNotFound) {
+		sbr.SetFinalizers(nil)
```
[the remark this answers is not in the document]
DhritiShikhar: Done 69fb48e

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -236,40 +241,102 @@ func (b *serviceBinder) onError(
...
+		if _, err = b.updateServiceBindingRequest(sbr); err != nil {
+			return noRequeue(err)
+		}
+		return requeueOnNotFound(errApplicationNotFound, requeueAfter)
```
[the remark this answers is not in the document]
DhritiShikhar: Even if err doesn't have a value, the function would return from here and not proceed further.

pkg/controller/servicebindingrequest/servicebinder.go
DhritiShikhar: Checked. This scenario is working as expected

pkg/controller/servicebindingrequest/servicebinder.go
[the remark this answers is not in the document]
DhritiShikhar: Changed this a bit 0526003

pkg/controller/servicebindingrequest/servicebinder.go
DhritiShikhar: handled it here 0526003

pkg/controller/servicebindingrequest/servicebinder.go
DhritiShikhar: handled it here 
fcf6e52

Avni-Sharma:
> Here 4.5-e2e is failed due to #549

@pratikjagrut @DhritiShikhar . It seems it is failing for all steps.
```
--- FAIL: TestAddSchemesToFramework (1361.69s)
    --- FAIL: TestAddSchemesToFramework/end-to-end (1338.27s)
        --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-etcd-unannotated-app-db-sbr (163.19s)
        --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-db-app-sbr (144.33s)
        --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-app-db-sbr (144.33s)
        --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-db-sbr-app (154.34s)
        --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-app-sbr-db (144.60s)
        --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-sbr-db-app (154.29s)
        --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-sbr-app-db (144.33s)
        --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-csv-db-app-sbr (144.44s)
        --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-csv-app-db-sbr (144.42s)
FAIL
```
I guess it should fail only for the step when sbr is first applied.

Avni-Sharma:
I followed the steps to run the example described in the description, manually.

After I run the command `oc get ServiceBindingRequest binding-request -o yaml` . I get the following.

```
avni@localhost service-binding-operator (DhritiShikhar-new-split-status)=>>   k get sbr binding-request -o yaml
kind: ServiceBindingRequest
metadata:
  creationTimestamp: "2020-07-14T07:47:16Z"
  finalizers:
  - finalizer.servicebindingrequest.openshift.io
  generation: 2
  name: binding-request
  namespace: default
  resourceVersion: "128901"
  selfLink: /apis/apps.openshift.io/v1alpha1/namespaces/default/servicebindingrequests/binding-request
  uid: 235506e6-744c-467d-bf3c-a25e181adfb6
spec:
  applicationSelector:
    group: ""
    resource: ""
    version: ""
  backingServiceSelector:
    group: postgresql.baiju.dev
    kind: Database
    resourceRef: db-demo
    version: v1alpha1
status:
  conditions:
  - lastHeartbeatTime: "2020-07-14T07:47:43Z"
    lastTransitionTime: "2020-07-14T07:47:25Z"
    status: "True"
    type: CollectionReady
  - lastHeartbeatTime: "2020-07-14T07:47:43Z"
    lastTransitionTime: "2020-07-14T07:47:25Z"
    message: application selectors are empty
    reason: EmptyApplicationSelector
    status: "False"
    type: InjectionReady
  secret: binding-request
```

The thing here to notice is that somehow the API `resourceRef` in applicationSelector is changing to `resource`, so when I try to apply  this example https://github.com/redhat-developer/service-binding-operator/blob/master/examples/nodejs_postgresql/service-binding-request.yaml , the sbr does not acknowledge the resorceRef API. The same behaviour can be observed in the video too at the 20th second.

pratikjagrut:
> > Here 4.5-e2e is failed due to #549
> 
> @pratikjagrut @DhritiShikhar . It seems it is failing for all steps.
> 
> ```
> --- FAIL: TestAddSchemesToFramework (1361.69s)
>     --- FAIL: TestAddSchemesToFramework/end-to-end (1338.27s)
>         --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-etcd-unannotated-app-db-sbr (163.19s)
>         --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-db-app-sbr (144.33s)
>         --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-app-db-sbr (144.33s)
>         --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-db-sbr-app (154.34s)
>         --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-app-sbr-db (144.60s)
>         --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-sbr-db-app (154.29s)
>         --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-sbr-app-db (144.33s)
>         --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-csv-db-app-sbr (144.44s)
>         --- FAIL: TestAddSchemesToFramework/end-to-end/scenario-csv-app-db-sbr (144.42s)
> FAIL
> ```
> 
> I guess it should fail only for the step when sbr is first applied.

@Avni-Sharma This is not failing due to racing condition we saw, this failure is at 
```
// after deletion, deployment should not contain envFrom directive anymore
	require.Eventually(t, func() bool {
		t.Logf("Inspecting deployment '%s'", deploymentNamespacedName)
		_, err := assertDeploymentEnvFrom(todoCtx, f, deploymentNamespacedName, sbrName)
		if err != nil {
			t.Log("expect deployment not to be carrying envFrom directive")
			return true
		}
		return false
	}, 120*time.Second, 2*time.Second)
```
After delation envFrom is still found in deployment.
First log msg is the same `t.Logf("Inspecting deployment '%s'", deploymentNamespacedName)` we need to change it something else to avoid confusion.
@DhritiShikhar Can you confirm this scenario with your PR?

DhritiShikhar:
@pratikjagrut which scenario?

pratikjagrut:
> @pratikjagrut which scenario?
```
// after deletion, deployment should not contain envFrom directive anymore
	require.Eventually(t, func() bool {
		t.Logf("Inspecting deployment '%s'", deploymentNamespacedName)
		_, err := assertDeploymentEnvFrom(todoCtx, f, deploymentNamespacedName, sbrName)
		if err != nil {
			t.Log("expect deployment not to be carrying envFrom directive")
			return true
		}
		return false
	}, 120*time.Second, 2*time.Second)
```

Avni-Sharma:
e2e are looking for applicationSelector.resourceRef and here in SBR we are somehow having  applicationSleector.resource. Maybe that's why all steps are failing.

DhritiShikhar:
> The thing here to notice is that somehow the API resourceRef in applicationSelector is changing to resource, so when I try to apply this example

@Avni-Sharma 

This is the result on EmptyApplicationSelector
```
spec:                                                      
  applicationSelector:
    group: ""                                              
    resource: ""                                           
    version: "" 
```

This is the result when application is specified by user
```
  applicationSelector:
    group: apps
    resource: deployments
    resourceRef: nodejs-rest-http-crud
    version: v1
```

`resource` field is not converted to `resourceRef`

Avni-Sharma:
I guess I got confused between resource and resourceful. shouldn't we getting resourceRef as: "" then?

DhritiShikhar:
> I guess I got confused between resource and resourceRef. shouldn't we be getting resourceRef as: "" then?

Wouldn't address it in this PR :)

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -236,6 +241,64 @@ func (b *serviceBinder) onError(
...
+func isApplicationSelectorEmpty(
+	application v1alpha1.ApplicationSelector,
+) bool {
+	var emptyApplication v1alpha1.ApplicationSelector
+	if application == emptyApplication ||
+		application.ResourceRef == "" &&
+			application.LabelSelector.MatchLabels == nil {
+		return true
+	}
+	return false
```
isutton, line 255: Perhaps this can be simplified:

```suggestion
	// in the case application.LabelSelector is a pointer:
	// return application.ResourceRef == "" && application.LabelSelector != nil && application.LabelSelector.MatchLabels == nil
	return application.ResourceRef == "" && application.LabelSelector.MatchLabels == nil
```

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -236,6 +241,64 @@ func (b *serviceBinder) onError(
...
+	}
+
+	b.logger.Info(applicationError.Error())
+
+	if errors.Is(applicationError, errApplicationNotFound) {
+		removeFinalizer(b.sbr)
+		if _, err = b.updateServiceBindingRequest(sbr); err != nil {
+			return requeueError(err)
+		}
+		return requeueOnNotFound(errApplicationNotFound, requeueAfter)
```
isutton, line 296: The original line indicated to always requeue on the scenario an application can't be found, so the following might make the intention clearer:

```suggestion
		return requeue(applicationError, requeueAfter)
```

pkg/controller/servicebindingrequest/servicebinder.go
```diff
@@ -248,11 +315,18 @@ func (b *serviceBinder) bind() (reconcile.Result, error) {
 		return b.onError(err, b.sbr, sbrStatus, nil)
 	}
 	sbrStatus.Secret = secretObj.GetName()
+	b.sbr.Status.Secret = secretObj.GetName()
```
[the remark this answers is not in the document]
DhritiShikhar: This is outdated. 
This has already been addressed.

pkg/controller/servicebindingrequest/servicebinder.go
DhritiShikhar: Why not use requeueOnNotFound(...) ?

```suggestion
		return requeueOnNotFound(applicationError, requeueAfter)
```

pkg/controller/servicebindingrequest/servicebinder.go
DhritiShikhar: Not committing this directly.
Made some changes here 7f886cd

pkg/controller/servicebindingrequest/servicebinder.go
isutton: Your suggestion is equivalent to my suggestion within the condition wrapping this return statement, since within the `if errors.Is(applicationError, errApplicationNotFound) {` block `applicationError` will always itself be or be wrapped by an `errApplicationNotFound` error.

DhritiShikhar: addressed 1f5f505

PR #502, opened 2020-05-28
[WIP] Update ServiceBindingRequest.Status.Conditions when service not found

DhritiShikhar:
**This PR is on hold. Based on discussion with Predrag, I will pick up this PR in next sprint.**

### Motivation

When `Service` is not available, update `status.conditions` as
```
  conditions:
  - lastHeartbeatTime: "2020-05-28T20:07:51Z"
    lastTransitionTime: "2020-05-28T20:07:26Z"
    message: service not found
    reason: ServiceNotFound
    status: "False"
    type: Ready
```

### Changes

[1] Adds a check for Service Not Found case.

### How to test

- Import an application
- Create a DB instance 

```
kind: ServiceBindingRequest
metadata:
  name: binding-request
spec:
  applicationSelector:
        resourceRef: nodejs-rest-http-crud 
        group: apps
        version: v1
        resource: deployments
 backingServiceSelector:
    group: postgresql.baiju.dev
    version: v1alpha1
    kind: Database
    resourceRef: db-demo
```

3. Use the following command to check if the status has been updated:

The output should have `status.conditions.type` as 
-  `BindingReady` with `status: "False"`

```
status:
  conditions:
  - lastHeartbeatTime: "2020-05-28T20:07:51Z"
    lastTransitionTime: "2020-05-28T20:07:26Z"
    message: service not found
    reason: ServiceNotFound
    status: "False"
    type: Ready
  secret: ""

```

### Note

This PR does not deal with the scenario where `Service` is deleted and service is re-created with the same name.


Avni-Sharma:
According to this https://github.com/redhat-developer/service-binding-operator/pull/498 we won't be having BindingReady. In that case we can have changes related to conditions when service not found  in #498 itself IMO.

DhritiShikhar:
@Avni-Sharma I will rebase according to whichever PR gets in first.

Avni-Sharma:
> @Avni-Sharma I will rebase according to whichever PR gets in first.

It means that either way both PRs are interrelated(since it is also depending as to which goes in first), so in that case, it would be good if it goes in as a single PR

DhritiShikhar:
> It means that either way both PRs are interrelated(since it is also depending as to which goes in first), 

If it's about rebasing then in that case, every PR in this project is interrelated.
Are you suggesting that rebasing a PR is not appropriate?

> so in that case, it would be good if it goes in as a single PR

The only thing similar between this PR and #498 is that they both concern around `Status`.

This PR handles the case when a `Service` is not found.
#498 's concern is around `Application`.

I think the two PRs are modular enough and can go in separately.

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -155,6 +156,19 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		r.RestMapper,
 	)
 	if err != nil {
+		if err == ErrServiceNotFound {
+			conditionsv1.SetStatusCondition(&sbr.Status.Conditions, conditionsv1.Condition{
+				Type:    BindingReady,
+				Status:  corev1.ConditionFalse,
+				Reason:  ServiceNotFoundReason,
+				Message: ErrServiceNotFound.Error(),
```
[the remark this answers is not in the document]
DhritiShikhar: Yes we could. Let me check more on this and get back to you.

DhritiShikhar:
@pedjak 

This PR should not be on hold for #512 
The correct PR is #498

pedjak:
sorry @DhritiShikhar , my bad

/hold cancel

pkg/controller/servicebindingrequest/reconciler.go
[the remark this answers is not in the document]
DhritiShikhar: not every backing service has a reference number

Avni-Sharma:
@DhritiShikhar  what happens when we have multiple services and one of them is present and the other one isn't?

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -157,6 +158,19 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		r.restMapper,
 	)
 	if err != nil {
+		if err == errServiceNotFound {
+			conditionsv1.SetStatusCondition(&sbr.Status.Conditions, conditionsv1.Condition{
+				Type:    BindingReady,
+				Status:  corev1.ConditionFalse,
+				Reason:  ServiceNotFoundReason,
```
Avni-Sharma, line 165: We can have a slice of services for multiple backing services, wherein we know which service is found and which service is not found, and then flag the readiness as False

DhritiShikhar: Kind of like we have for applications?

```
status:
  applications:
  - group: apps
    kind: Deployment
    name: nodejs-rest-http-crud
    version: v1
```

pkg/controller/servicebindingrequest/reconciler_test.go
```diff
@@ -400,6 +400,45 @@ func TestEmptyApplicationSelector(t *testing.T) {
 	require.Equal(t, 0, len(sbrOutput.Status.Applications))
 }
 
+// TestEmptyApplicationSelector tests that Status is successfully updated when Service is not found
+func TestServiceNotFound(t *testing.T) {
```
isutton, line 404: Please adjust this to a test table with the following cases:

* Reconcile without a service, to validate whether the `BindingReady` condition is `False`; and
* Reconcile with a service, to validate whether `BindingReady` condition is `True`.

Please observe any other conditions that might be checked as well.

isutton:
> @DhritiShikhar  what happens when we have multiple services and one of them is present and the other one isn't? 

It depends whether we are taking a "best effort" or an "all or nothing" approach.

The best effort approach could be collect as many services as possible, and populate the SBR secret with the collected values; the way the user is informed by this could be yet another condition (IncompleteServices) with a proper reason (ServiceNotFoundReason) and a message containing the name (or names) of the missing services.

The all or nothing approach could be stop at the first missing service, state this in the `CollectionReady` condition with the proper reason (ServiceNotFoundReason) and not create the SBR's secret.

I believe the latter might be better suited for now.

pedjak:
@DhritiShikhar can we close this PR now, given the work in https://github.com/redhat-developer/service-binding-operator/pull/681 ?

DhritiShikhar:
Yes.

PR #514, opened 2020-06-16
Add unit test to verify binding two ServiceBindingRequests with a single application

pratikjagrut:
### Motivation

We support binding multiple backing services to a single application, this test is to vouch for that feature.

### Changes

Adding a new test `TestBindTwoSbrsWithSingleApplication` in reconciler_test.go.

### Testing

To run all the unit tests
`make test-unit`

To run only this test
`go test ./pkg/controller/servicebindingrequest/ -run=TestBindTwoSbrsWithSingleApplication -v`

### DIFF
--- a/pkg/controller/servicebindingrequest/reconciler_test.go
+++ b/pkg/controller/servicebindingrequest/reconciler_test.go
@@ -399,3 +399,114 @@ func TestEmptyApplicationSelector(t *testing.T) {
 	require.Equal(t, corev1.ConditionTrue, sbrOutput.Status.Conditions[0].Status)
 	require.Equal(t, 0, len(sbrOutput.Status.Applications))
 }
+
+func TestBindTwoSbrsWithSingleApplication(t *testing.T) {
+	applicationResourceRef := "applicationResourceRef"
+	backingServiceResourceRef1 := "backingServiceRef1"
+	backingServiceResourceRef2 := "backingServiceRef2"
+
+	sbrName1 := "binding-request1"
+	sbrName2 := "binding-request2"
+
+	f := mocks.NewFake(t, reconcilerNs)
+	f.AddMockedUnstructuredDeployment(applicationResourceRef, nil)
+	f.AddMockedUnstructuredServiceBindingRequest(sbrName1, backingServiceResourceRef1, applicationResourceRef, deploymentsGVR, nil)
+	f.AddMockedUnstructuredServiceBindingRequest(sbrName2, backingServiceResourceRef2, applicationResourceRef, deploymentsGVR, nil)
+	f.AddMockedUnstructuredDatabaseCR(backingServiceResourceRef1)
+	f.AddMockedUnstructuredDatabaseCR(backingServiceResourceRef2)
+
+	fakeDynClient := f.FakeDynClient()
+	mapper := testutils.BuildTestRESTMapper()
+	r := &reconciler{dynClient: fakeDynClient, restMapper: mapper, scheme: f.S}
+	r.resourceWatcher = newFakeResourceWatcher(mapper)
+
+	t.Run("test-bind-two-sbrs-with-single-application", func(t *testing.T) {
+		// Reconciling first sbr
+		// First sbr reconcile request
+		req := reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Namespace: reconcilerNs,
+				Name:      sbrName1,
+			},
+		}
+
+		// Reconcile first sbr
+		res, err := r.Reconcile(req)
+		require.NoError(t, err)
+		require.False(t, res.Requeue)
+
+		// Get sbr after reconciling
+		namespacedName := types.NamespacedName{Namespace: reconcilerNs, Name: sbrName1}
+		sbrOutput, err := r.getServiceBindingRequest(namespacedName)
+		require.NoError(t, err)
+
+		// expected sbr assertion
+		expectedStatus := v1alpha1.BoundApplication{
+			GroupVersionKind: v1.GroupVersionKind{
+				Group:   deploymentsGVR.Group,
+				Version: deploymentsGVR.Version,
+				Kind:    "Deployment",
+			},
+			LocalObjectReference: corev1.LocalObjectReference{
+				Name: applicationResourceRef,
+			},
+		}
+
+		require.Equal(t, BindingReady, sbrOutput.Status.Conditions[0].Type)
+		require.Equal(t, corev1.ConditionTrue, sbrOutput.Status.Conditions[0].Status)
+		require.Equal(t, sbrName1, sbrOutput.Status.Secret)
+		require.Len(t, sbrOutput.Status.Applications, 1)
+		require.True(t, reflect.DeepEqual(expectedStatus, sbrOutput.Status.Applications[0]))
+
+		// Reconciling second sbr
+		// Second sbr reconcile request
+		req = reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Namespace: reconcilerNs,
+				Name:      sbrName2,
+			},
+		}
+
+		// Reconcile second sbr
+		res, err = r.Reconcile(req)
+		require.NoError(t, err)
+		require.False(t, res.Requeue)
+
+		// Get sbr after reconciling
+		namespacedName = types.NamespacedName{Namespace: reconcilerNs, Name: sbrName2}
+		sbrOutput, err = r.getServiceBindingRequest(namespacedName)
+		require.NoError(t, err)
+
+		// expected sbr assertion
+		expectedStatus = v1alpha1.BoundApplication{
+			GroupVersionKind: v1.GroupVersionKind{
+				Group:   deploymentsGVR.Group,
+				Version: deploymentsGVR.Version,
+				Kind:    "Deployment",
+			},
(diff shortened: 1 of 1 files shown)


DhritiShikhar:
@pedjak 

We have two different terms - `application` and `service`.

The test you have added is for testing two services with single SBR.

So, the PR description should be "binding two **services** with single SBR" and thus the test name `TestBindTwoSbrsWithSingleApplication` should be updated to reflect the correct scenario i.e. **services** not applications.

DhritiShikhar:
@pratikjagrut This makes me wonder what should happen if user provides same service twice in the SBR.

Should we inform the user to remove the duplicate service?

pkg/controller/servicebindingrequest/reconciler_test.go
```diff
@@ -399,3 +399,121 @@ func TestEmptyApplicationSelector(t *testing.T) {
 	require.Equal(t, corev1.ConditionTrue, sbrOutput.Status.Conditions[0].Status)
 	require.Equal(t, 0, len(sbrOutput.Status.Applications))
 }
+
+func TestBindTwoSbrsWithSingleApplication(t *testing.T) {
+	applicationResourceRef := "applicationResourceRef"
+	matchLabels := map[string]string{
```
DhritiShikhar, line 405: No point in using both `applicationResourceRef` as well as `matchLabels`.

pkg/controller/servicebindingrequest/reconciler_test.go
```diff
@@ -399,3 +399,121 @@ func TestEmptyApplicationSelector(t *testing.T) {
...
+	sbrName2 := "binding-request2"
+
+	f := mocks.NewFake(t, reconcilerNs)
+	f.AddMockedUnstructuredDeployment(applicationResourceRef, matchLabels)
+	f.AddMockedUnstructuredServiceBindingRequest(sbrName1, backingServiceResourceRef1, applicationResourceRef, deploymentsGVR, matchLabels)
+	f.AddMockedUnstructuredServiceBindingRequest(sbrName2, backingServiceResourceRef2, applicationResourceRef, deploymentsGVR, matchLabels)
+	f.AddMockedUnstructuredDatabaseCRD()
+	f.AddMockedUnstructuredDatabaseCR(backingServiceResourceRef1)
+	f.AddMockedUnstructuredDatabaseCR(backingServiceResourceRef2)
+	f.AddMockedUnstructuredSecret("db-credentials")
```
DhritiShikhar, line 423: Shouldn't there be two different secrets for two different services?

pratikjagrut: Two different services can use the same secrete and in the end, we want to assert if sbr's name is injected in env var or not. So doesn't really matter if we are using the same secret in the unit test as long as they are consumed by two different sbrs.

pratikjagrut:
> @pratikjagrut This makes me wonder what should happen if user provides same service twice in the SBR.
> 
> Should we inform the user to remove the duplicate service?

@DhritiShikhar I think so, we should totally inform the user and skip the binding.

pratikjagrut:
> @pratikjagrut
> 
> We have two different terms - `application` and `service`.
> 
> The test you have added is for testing two services with single SBR.

@DhritiShikhar 

But for 2 services we are creating 2 different SBRs. Which then will bind these backing services to a single application.

pkg/controller/servicebindingrequest/reconciler_test.go
pratikjagrut: Yeah, I agree, I just kept it to be consistent with other tests.

DhritiShikhar:
> But for 2 services we are creating 2 different SBRs.

Thanks for the explaination Pratik

pkg/controller/servicebindingrequest/reconciler_test.go
DhritiShikhar: Based on the discussion in standup, one secret per SBR.
So, there should be two secrets in this case.

pratikjagrut: I can add two secrets but it won't make any difference in the unit test, it will be impactful in e2e and real scenario as they will consume them, here we are just mentioning them to form a mock SBR, it won't affect the assertion.

pkg/controller/servicebindingrequest/reconciler_test.go
pratikjagrut: MatchLabels removed.

pkg/controller/servicebindingrequest/reconciler_test.go
```diff
@@ -399,3 +399,116 @@ func TestEmptyApplicationSelector(t *testing.T) {
...
+		resourceClient := r.dynClient.Resource(deploymentsGVR).Namespace(namespacedName.Namespace)
+		u, err := resourceClient.Get(applicationResourceRef, metav1.GetOptions{})
+		require.NoError(t, err)
+		dep := &appsv1.Deployment{}
+		err = runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, dep)
+		require.NoError(t, err)
+
+		// Assert SBR name with secretRef object in containers env var
+		// Expected secretRef=binding-request1
+		require.Equal(t, sbrName1, dep.Spec.Template.Spec.Containers[0].EnvFrom[0].SecretRef.LocalObjectReference.Name)
```
DhritiShikhar, line 510: Shouldn't `dep.Spec.Template.Spec.Containers[0].EnvFrom[0].SecretRef.LocalObjectReference.Name`
contain the secret name which in this case is `db-credentials`  - https://github.com/redhat-developer/service-binding-operator/pull/514/files#diff-14a37dadc390fc3ec23459efc088de8eR418

pkg/controller/servicebindingrequest/reconciler_test.go
DhritiShikhar: @pratikjagrut 

It's not about adding 2 secrets just for the sake of it.

Two SBRs get created. These 2 SBRs create 2 secrets. We inject secret information (not SBR name) in the application.

So, there MUST be 2 secrets present and there MUST be a check that both secrets are injected in application.

Currently, you are checking if SBR name is present in the deployment (https://github.com/redhat-developer/service-binding-operator/pull/514/files#diff-14a37dadc390fc3ec23459efc088de8eR510)

That is not correct.
 The actual check should be against, the secret name. Even if both secret name and SBR name is same, the check should still be against the secret name.

isutton:
> > @pratikjagrut This makes me wonder what should happen if user provides same service twice in the SBR.
> > Should we inform the user to remove the duplicate service?
> 
> @DhritiShikhar I think so, we should totally inform the user and skip the binding.

I've encountered the same issue in #515, and have implemented [a function to return distinct values from declared services](https://github.com/redhat-developer/service-binding-operator/pull/515/files#diff-42835fd3b25b196b1bbc21495e0ac5b7R71-R85).

Duplicated elements could be removed automatically on behalf of the user when the admission web-hook gets implemented, but until that the safest is to use the collection of distinct services instead; it is safe because either the entry is duplicated or not, and in the case a service is duplicated, it is only processed once during  the reconciliation process.

The rationale is: if this is a situation the operator can recover from, there's no reason to stop the process and make the user remove an entry from the services declaration; it is worth to mention an informational condition could be created to inform the user there's a duplicated entry in the services field, but it shouldn't prevent the well functioning of the operator because of that fact.

Shall we continue with the best-effort approach of keeping the list as is and continue the process using the distinct values from the services declaration?

pkg/controller/servicebindingrequest/reconciler_test.go
pratikjagrut: As per my understanding, it should be the intermediate secret's name, [it is stated in the last  para of doc](https://github.com/redhat-developer/service-binding-operator/blob/master/examples/nodejs_postgresql/README.md#express-an-intent-to-bind-the-db-and-the-application)

https://github.com/redhat-developer/service-binding-operator/pull/514/files#diff-14a37dadc390fc3ec23459efc088de8eR418 this is an actual secret.

pkg/controller/servicebindingrequest/reconciler_test.go
pratikjagrut: We decided to remove the mock secret so I'm resolving this conversation.

DhritiShikhar:
> The rationale is: if this is a situation the operator can recover from, there's no reason to stop the process and make the user remove an entry from the services declaration; it is worth to mention an informational condition could be created to inform the user there's a duplicated entry in the services field, but it shouldn't prevent the well functioning of the operator because of that fact.

I agree the process should not stop. 
And yes, We can pass information about this situation through conditions to user.

> Shall we continue with the best-effort approach of keeping the list as is and continue the process using the distinct values from the services declaration?

Yes

pkg/controller/servicebindingrequest/reconciler_test.go
pratikjagrut: Resolving this conversation.

Avni-Sharma:
@isutton  regarding https://github.com/redhat-developer/service-binding-operator/pull/514#issuecomment-645809793 , an update in the status would only be happen effectively if the similar services are in the same SBR, however, with 2 different sbrs and a similar service in both the sbrs, won't be able to update the status of their respective sbrs, with the knowledge of having distinct services.

Issue #509, opened 2020-06-10
Controller reconciling indefinitely after applying sbr yaml


Avni-Sharma:
After #442 got merged the controller has started to watch on application workloads. When we add a SBR to connect the app and db the reconcile is on an indefinite loop, whereas it should be stimulus to changes in the sbr yaml. Now if we see the topology view(OpenShift 4.x +) the app pod dougnout indefinitely changes itself from lighter blue to darker blue circles. The activity has been captured here 
https://bluejeans.com/s/jli5w/
I am thinking it is because the secret is getting injected again and again, and because we are watching the workload the reconciler hits and it is like a chicken egg prob. The first time a secret is injected, the sbr controller reads it as a change and reconciles again, and again secret is injected and this goes on.

Avni-Sharma:
Another issue that I am facing is that the resourceVersion in the secret is changing even when data, annotations is not changing.
Old secret
```
avni@localhost service-binding-operator (reconcile-indefinte)=>>   k get secrets binding-request -o yaml
apiVersion: v1
data:
  DATABASE_CONFIGMAP_DB_HOST: MTcyLjMwLjYxLjI0Mw==
  DATABASE_CONFIGMAP_DB_NAME: ZGItZGVtbw==
  DATABASE_CONFIGMAP_DB_PASSWORD: cGFzc3dvcmQ=
  DATABASE_CONFIGMAP_DB_PORT: NTQzMg==
  DATABASE_DBCONNECTIONIP: MTcyLjMwLjYxLjI0Mw==
  DATABASE_DBCONNECTIONPORT: NTQzMg==
  DATABASE_DBNAME: ZGItZGVtbw==
  DATABASE_SECRET_PASSWORD: cGFzc3dvcmQ=
  DATABASE_SECRET_USER: cG9zdGdyZXM=
kind: Secret
metadata:
  annotations:
    service-binding-operator.apps.openshift.io/binding-name: binding-request
    service-binding-operator.apps.openshift.io/binding-namespace: default
  creationTimestamp: "2020-06-22T05:50:45Z"
  managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:data:
        .: {}
        f:DATABASE_CONFIGMAP_DB_HOST: {}
        f:DATABASE_CONFIGMAP_DB_NAME: {}
        f:DATABASE_CONFIGMAP_DB_PASSWORD: {}
[...]

PR #523, opened 2020-06-26
Fix infinite reconcile loop

Avni-Sharma:
### Motivation

To resolve issue #509 

### Changes

Initially the resourceVersion of the secret used to get updated even without any change in the data field. In this refactor we would only update the secret after fetching the already existing secret and comparing it's data with the payload(new data to be injected). If secret exists then we go further to check the contents with which the secret needs to be updated. If data is same we skip the update, if data is diff we go further with updating the secret resource.

### Testing

`make test-unit`

Import an app
create a db
create a sbr

Check the reosurceVersion of the secret and the in the envVar of the deployment. Both should be the same. 
Update the backing service, now the intermediate secret reourceVersion should change and further change envVar in the deployment

### DIFF
--- a/pkg/controller/servicebindingrequest/binder.go
+++ b/pkg/controller/servicebindingrequest/binder.go
@@ -454,8 +454,8 @@ func (b *binder) removeVolumeMounts(volumeMounts []corev1.VolumeMount) []corev1.
 	return cleanVolumeMounts
 }
 
-// nestedMapComparison compares a nested field from two objects.
-func nestedMapComparison(a, b *unstructured.Unstructured, fields ...string) (bool, error) {
+// nestedUnstructuredComparison compares a nested field from two objects.
+func nestedUnstructuredComparison(a, b *unstructured.Unstructured, fields ...string) (bool, error) {
 	var (
 		aMap map[string]interface{}
 		bMap map[string]interface{}
@@ -476,9 +476,14 @@ func nestedMapComparison(a, b *unstructured.Unstructured, fields ...string) (boo
 		return false, nil
 	}
 
-	result := cmp.DeepEqual(aMap, bMap)()
+	result := nestedMapComparison(aMap, bMap)
+	return result, nil
+}
 
-	return result.Success(), nil
+func nestedMapComparison(a, b map[string]interface{}) bool {
+	val := cmp.DeepEqual(a, b)()
+	return val.Success()
+
 }
 
 // update the list of objects informed as unstructured, looking for "containers" entry. This method
@@ -508,7 +513,7 @@ func (b *binder) update(objs *unstructured.UnstructuredList) ([]*unstructured.Un
 			}
 		}
 
-		if specsAreEqual, err := nestedMapComparison(&obj, updatedObj); err != nil {
+		if specsAreEqual, err := nestedUnstructuredComparison(&obj, updatedObj); err != nil {
 			log.Error(err, "")
 			continue
 		} else if specsAreEqual {
--- a/pkg/controller/servicebindingrequest/binder.go
+++ b/pkg/controller/servicebindingrequest/binder.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"strings"
-	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -386,10 +385,15 @@ func (b *binder) updateContainer(container interface{}) (map[string]interface{},
 	// effectively binding the application with intermediary secret
 	c.EnvFrom = b.appendEnvFrom(c.EnvFrom, b.sbr.GetName())
 
+	secretRes := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "secrets"}
+	existingSecret, err := b.dynClient.Resource(secretRes).Namespace(b.sbr.GetNamespace()).Get(b.sbr.GetName(), metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
 	// add a special environment variable that is only used to trigger a change in the declaration,
 	// attempting to force a side effect (in case of a Deployment, it would result in its Pods to be
 	// restarted)
-	c.Env = b.appendEnvVar(c.Env, changeTriggerEnv, time.Now().Format(time.RFC3339))
+	c.Env = b.appendEnvVar(c.Env, changeTriggerEnv, existingSecret.GetResourceVersion())
 
 	if len(b.volumeKeys) > 0 {
 		// and adding volume mount entries
--- a/pkg/controller/servicebindingrequest/binder_test.go
+++ b/pkg/controller/servicebindingrequest/binder_test.go
@@ -56,7 +56,7 @@ func TestBinderNew(t *testing.T) {
 		map[string]string{},
 	)
 
-	f.AddMockedUnstructuredSecret(name)
+	f.AddMockedUnstructuredSecretRV(name)
 	fakeDynClient := f.FakeDynClient()
 
 	t.Run("search-using-resourceref", func(t *testing.T) {
--- a/pkg/controller/servicebindingrequest/binder_test.go
+++ b/pkg/controller/servicebindingrequest/binder_test.go
@@ -172,7 +172,6 @@ func TestBinderNew(t *testing.T) {
 		envVar := getEnvVar(c.Env, changeTriggerEnv)
 		require.NotNil(t, envVar)
 		require.NotEmpty(t, envVar.Value)
-		require.NoError(t, err)
 
 		uSecret, err := fakeDynClient.Resource(secretsGVR).Get(name, metav1.GetOptions{})
 		require.NoError(t, err)
--- a/pkg/controller/servicebindingrequest/binder_test.go
+++ b/pkg/controller/servicebindingrequest/binder_test.go
@@ -4,13 +4,13 @@ import (
 	"context"
 	"fmt"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -55,6 +55,9 @@ func TestBinderNew(t *testing.T) {
 		deploymentsGVR,
 		map[string]string{},
 	)
+
+	f.AddMockedUnstructuredSecret(name)
+	fakeDynClient := f.FakeDynClient()
 
 	t.Run("search-using-resourceref", func(t *testing.T) {
 		binderForSBRWithResourceRef := newBinder(
@@ -169,10 +172,14 @@ func TestBinderNew(t *testing.T) {
 		envVar := getEnvVar(c.Env, changeTriggerEnv)
 		require.NotNil(t, envVar)
 		require.NotEmpty(t, envVar.Value)
-
-		parsedTime, err := time.Parse(time.RFC3339, envVar.Value)
 		require.NoError(t, err)
-		require.True(t, parsedTime.Before(time.Now()))
+
+		uSecret, err := fakeDynClient.Resource(secretsGVR).Get(name, metav1.GetOptions{})
+		require.NoError(t, err)
+		s := corev1.Secret{}
+		err = runtime.DefaultUnstructuredConverter.FromUnstructured(uSecret.Object, &s)
+		require.NoError(t, err)
+		require.Equal(t, s.ObjectMeta.ResourceVersion, envVar.Value)
 
 	})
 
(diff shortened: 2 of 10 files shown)


pkg/controller/servicebindingrequest/secret.go
```diff
@@ -47,14 +49,26 @@ func (s *secret) createOrUpdate(payload map[string][]byte) (*unstructured.Unstru
...
-	}
-
-	logger.Debug("Secret already exists, updating contents instead...")
-	_, err = resourceClient.Update(u, metav1.UpdateOptions{})
 	if err != nil {
-		return nil, err
+		if errors.IsAlreadyExists(err) {
+			existingSecret, err := resourceClient.Get(s.name, metav1.GetOptions{})
+			if err != nil {
+				logger.Debug("Could not fetch the existing secret")
```
DhritiShikhar, line 56: There should be a return statement here

Avni-Sharma: yeah, I saw later that I forget it

pkg/controller/servicebindingrequest/secret.go
```diff
@@ -47,14 +49,26 @@ func (s *secret) createOrUpdate(payload map[string][]byte) (*unstructured.Unstru
...
 	_, err = resourceClient.Create(u, metav1.CreateOptions{})
-	if err != nil && !errors.IsAlreadyExists(err) {
-		return nil, err
-	}
-
-	logger.Debug("Secret already exists, updating contents instead...")
-	_, err = resourceClient.Update(u, metav1.UpdateOptions{})
 	if err != nil {
-		return nil, err
+		if errors.IsAlreadyExists(err) {
```
isutton, line 53: I believe the right flow would be:

* Get the resource
* If not found, create
* If found and existing data is different from the given payload, update the secret.

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -377,10 +376,15 @@ func (b *binder) updateContainer(container interface{}) (map[string]interface{},
...
+	secretRes := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "secrets"}
+	existingSecret, err := b.dynClient.Resource(secretRes).Namespace(b.sbr.GetNamespace()).Get(b.sbr.GetName(), metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
 	// add a special environment variable that is only used to trigger a change in the declaration,
 	// attempting to force a side effect (in case of a Deployment, it would result in its Pods to be
 	// restarted)
-	c.Env = b.appendEnvVar(c.Env, changeTriggerEnv, time.Now().Format(time.RFC3339))
+	c.Env = b.appendEnvVar(c.Env, changeTriggerEnv, existingSecret.GetResourceVersion())
```
isutton, line 387: I like this change; this makes it possible to properly compare old and new Object's `spec` in the controller predicate's update function since the generated environment variable value will be stable as long as the secret has the same resource version.

pkg/controller/servicebindingrequest/secret.go
```diff
@@ -46,15 +48,27 @@ func (s *secret) createOrUpdate(payload map[string][]byte) (*unstructured.Unstru
...
-	}
-
-	logger.Debug("Secret already exists, updating contents instead...")
-	_, err = resourceClient.Update(u, metav1.UpdateOptions{})
+	existingSecret, err := resourceClient.Get(s.name, metav1.GetOptions{})
 	if err != nil {
-		return nil, err
+		if errors.IsAlreadyExists(err) {
+			existingSecretData, _, _ := unstructured.NestedMap(existingSecret.Object, "data")
+			eq := reflect.DeepEqual(existingSecretData, payload)
```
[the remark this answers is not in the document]
Avni-Sharma: nestedMapComparison requires unstructred.Unstructured.  Here I am trying to compare just the data fields

pkg/controller/servicebindingrequest/secret.go
```diff
@@ -46,16 +49,33 @@ func (s *secret) createOrUpdate(payload map[string][]byte) (*unstructured.Unstru
...
-	_, err = resourceClient.Update(u, metav1.UpdateOptions{})
+	existingSecret, err := resourceClient.Get(s.name, metav1.GetOptions{})
 	if err != nil {
+		if errors.IsNotFound(err) {
+			_, err := resourceClient.Create(u, metav1.CreateOptions{})
+			return u, err
+		}
 		return nil, err
 	}
+	existingSecretData, _, _ := unstructured.NestedMap(existingSecret.Object, "data")
```
isutton, line 60: `found` and `err` shouldn't be ignored in here since the client is not typed.

pkg/controller/servicebindingrequest/secret.go
```diff
@@ -46,16 +49,33 @@ func (s *secret) createOrUpdate(payload map[string][]byte) (*unstructured.Unstru
...
+	existingSecretDataStr := make(map[string]string)
+	for k, v := range existingSecretData {
+		existingSecretDataStr[k] = v.(string)
+	}
+	payloadStr := make(map[string]string)
+	for k, v := range payload {
+		payloadStr[k] = base64.StdEncoding.EncodeToString(v)
+	}
+	eq := reflect.DeepEqual(existingSecretDataStr, payloadStr)
```
isutton, line 69: This transformation can be extracted to its own function, and the comparison instead of using `reflect.DeepEqual` use `cmp.DeepEqual` as we're using in other cases.

Perhaps `nestedMapComparison` can be renamed to `nestedUnstructuredComparison`, and L454-L466 below can be extracted to `nestedMapComparison`:

https://github.com/redhat-developer/service-binding-operator/blob/a7aa2af3a3df9d7afef2f75d26f163155fbc4ff4/pkg/controller/servicebindingrequest/binder.go#L445-L469

pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -169,10 +172,14 @@ func TestBinderNew(t *testing.T) {
 		envVar := getEnvVar(c.Env, changeTriggerEnv)
 		require.NotNil(t, envVar)
 		require.NotEmpty(t, envVar.Value)
+		require.NoError(t, err)
```
DhritiShikhar, line 175: Why an additional `require.NoError(t, err)`?

pkg/controller/servicebindingrequest/secret.go
Avni-Sharma: @isutton here I am comparing between map[string]string , nestedComparison would want a map[string]interface{}.  Payload infact is coming as (payload map[string][]byte) in the createOrUpdate function

test/mocks/mocks.go
```diff
@@ -360,8 +360,9 @@ func SecretMock(ns, name string) *corev1.Secret {
 			APIVersion: "v1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace: ns,
-			Name:      name,
+			Namespace:       ns,
+			Name:            name,
+			ResourceVersion: "116076",
```
pratikjagrut, line 365: I think we can not set or update resourceVersion it is done by cluster, this mock will work for the unit test but on actual cluster setting the resourceVersion will throw an error. The `get, list and watch operations` support the resourceVersion parameter. You might need to create another mock secret with the resourceVersion field for the unit test.

pratikjagrut: Your e2e CI tests are failing for the same reason.
```
TestAddSchemesToFramework/end-to-end/scenario-sbr-db-app: servicebindingrequest_test.go:321: Creating Database credentials secret mock object...
    TestAddSchemesToFramework/end-to-end/scenario-sbr-db-app: servicebindingrequest_test.go:323: 
        	Error Trace:	servicebindingrequest_test.go:323
        	            				servicebindingrequest_test.go:540
        	            				servicebindingrequest_test.go:155
        	            				servicebindingrequest_test.go:103
        	Error:      	Received unexpected error:
        	            	resourceVersion should not be set on objects to be created
        	Test:       	TestAddSchemesToFramework/end-to-end/scenario-sbr-db-app
```

pratikjagrut:
looks like this experiment came to fruition.

pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -56,6 +56,9 @@ func TestBinderNew(t *testing.T) {
 		map[string]string{},
 	)
 
+	f.AddMockedUnstructuredSecretRV(name)
```
isutton, line 59: It might not be required another method for this resource, as `AddMOckedUnstructuredSecret` returns a reference for the secret that can be modified to add the resource version:

```go
s := f.AddMockedUnstructuredSecret(name)
s.SetResourceVersion("112200")
```

pkg/controller/servicebindingrequest/binder.go
```diff
@@ -377,10 +376,15 @@ func (b *binder) updateContainer(container interface{}) (map[string]interface{},
 	// effectively binding the application with intermediary secret
 	c.EnvFrom = b.appendEnvFrom(c.EnvFrom, b.sbr.GetName())
 
+	secretRes := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "secrets"}
```
isutton, line 379: Another option here is to modify `newBinder` to accept the `secret` created in `newServiceBinder`, and use `secret.get()`.

pkg/controller/servicebindingrequest/reconciler_test.go
```diff
@@ -282,14 +282,40 @@ func TestReconcilerGenericBinding(t *testing.T) {
...
+	f.AddMockedUnstructuredDatabaseCR(backingServiceResourceRef)
+	f.AddMockedUnstructuredSecret("db-credentials")
+	f.AddMockedUnstructuredDeployment(reconcilerName, matchLabels)
+
+	fakeDynClient := f.FakeDynClient()
+	mapper := testutils.BuildTestRESTMapper()
+	r := &reconciler{dynClient: fakeDynClient, restMapper: mapper, scheme: f.S}
+	r.resourceWatcher = newFakeResourceWatcher(mapper)
+
+	u, err := fakeDynClient.Resource(deploymentsGVR).Get(reconcilerName, metav1.GetOptions{})
```
isutton, line 313: `f.AddMockedUnstructuredDeployment` adds the deployment to the fake client and returns it for further configuration.

pkg/controller/servicebindingrequest/reconciler_test.go
```diff
@@ -282,14 +282,40 @@ func TestReconcilerGenericBinding(t *testing.T) {
...
+	require.NoError(t, err)
+
+	d := appsv1.Deployment{}
+	err = runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &d)
+	require.NoError(t, err)
 
 	u, err = fakeDynClient.Resource(secretsGVR).Get("db-credentials", metav1.GetOptions{})
```
isutton, line 320: `f.AddMockedUnstructuredSecret` adds the secret to the fake client and returns it for further configuration, such as setting the resource version as performed in L329:

```go
s.SetResourceVersion("112200")
```

pkg/controller/servicebindingrequest/secret.go
```diff
@@ -46,16 +49,33 @@ func (s *secret) createOrUpdate(payload map[string][]byte) (*unstructured.Unstru
...
 
 	logger.Debug("Attempt to create secret...")
-	_, err = resourceClient.Create(u, metav1.CreateOptions{})
-	if err != nil && !errors.IsAlreadyExists(err) {
-		return nil, err
-	}
-
-	logger.Debug("Secret already exists, updating contents instead...")
-	_, err = resourceClient.Update(u, metav1.UpdateOptions{})
+	existingSecret, err := resourceClient.Get(s.name, metav1.GetOptions{})
```
isutton, line 52: `s.get()` can be used instead.

pkg/controller/servicebindingrequest/binder_test.go
Avni-Sharma: I tried this but then s.metadata.resourceVersion is not working for me.

pkg/controller/servicebindingrequest/secret.go
Avni-Sharma: Done https://github.com/redhat-developer/service-binding-operator/pull/523/commits/123bc03e78ad4ffa0645ffeaac3854a620203ccf

pkg/controller/servicebindingrequest/reconciler_test.go
Avni-Sharma: I have added the above-said way of setting resourceVersion, but I guess that the e2e will fail because the same secret mock is being used in e2e as well and the resourceVersion then will change.  Hence I added a different mock with resourceVersion  which is not being used in e2e. The reason can be found here https://github.com/redhat-developer/service-binding-operator/pull/523#discussion_r452005133

Avni-Sharma:
retest

pkg/controller/servicebindingrequest/binder.go
Avni-Sharma: Yeah, It can be a possibility but adding that here seems to be an overkill IMO

pkg/controller/servicebindingrequest/binder_test.go
Avni-Sharma: Adding another secret due to https://github.com/redhat-developer/service-binding-operator/pull/523#discussion_r452005133

PR #529, opened 2020-06-29
Update docs with a missing example

Avni-Sharma:
- An example scenario got missed in the README.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,7 +148,6 @@ GOLANGCI_LINT_BIN=$(OUTPUT_DIR)/golangci-lint
 # -- Variables for uploading code coverage reports to Codecov.io --
 # This default path is set by the OpenShift CI
 CODECOV_TOKEN_PATH ?= "/usr/local/redhat-developer-service-binding-operator-codecov-token/token"
-CODECOV_TOKEN ?= @$(CODECOV_TOKEN_PATH)
 REPO_OWNER := $(shell echo $$CLONEREFS_OPTIONS | jq '.refs[0].org')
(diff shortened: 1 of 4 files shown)


docs/OperatorBestPractices.md
```diff
@@ -119,3 +119,7 @@ Each example makes use of one of the reference operators and includes
 instructions for deploying the reference operators to a cluster, either
 through the command line or client web console UI. The examples are
 available [here.](https://github.com/redhat-developer/service-binding-operator/blob/master/README.md#example-scenarios)
+
+## Service Binding Specification
+
+An effort has been taken to encapsulate the best practices of service binding in the form of a [specification](https://github.com/k8s-service-bindings). The goal of this specification is to create a Kubernetes-wide specification for communicating service secrets to applications in an automated way. It aims to create a mechanism that is widely applicable.
```
pedjak, line 125: IMHO, the last two sentence might be removed, without changing the message.

pedjak: At the same time, I would rather mention the spec in README.md, it is a more visible place.

Avni-Sharma: I don't have a strong opinion on this but I felt it would be good in Operator Best Practices as the spec is still being worked upon and it would take some time for us to incorporate the changes in SBO completely, like annotations part, annotations being extensions rather than in the core spec, etc. Till then maybe we can avoid it mentioning in the main README.md. ? cc @baijum @sbose78

sbose78:
@Avni-Sharma @pedjak I would hold adding a link to the specification right away till we scope out what exactly we'll adopt and when.

Avni-Sharma:
closing this as forced pushed by mistake

Issue #530, opened 2020-06-30
SBR can not be deleted after the bound application has been deleted


qibobo:
The issue was fixed in https://github.com/redhat-developer/service-binding-operator/commit/f5dfc45e069ef32d7d6fee04e1f575f068d04ae5,
but now it re-occurs again.

I will provide a PR for it.

PR #531, opened 2020-06-30
fix issue #530 SBR can not be deleted after the bound application has been deleted

qibobo:


Avni-Sharma:
Hi @qibobo Thanks for the PR. Can we have an additional test in the reconciler_test.go also to see the reconcile activity on a deleted application?

pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -480,3 +483,61 @@ func Test_extraFieldsModifier(t *testing.T) {
 	})
 
 }
+
+func TestBindAndUnbind(t *testing.T) {
+
+	t.Run("bind and unbind", func(t *testing.T) {
```
pedjak, line 489: can you please share more infos what we are testing here, given the scope of PR? And update the test name accordingly: what is the context and the expected outcome?

pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -480,3 +483,61 @@ func Test_extraFieldsModifier(t *testing.T) {
...
+		binder := newBinder(
+			context.TODO(),
+			f.FakeDynClient(),
+			sbr,
+			[]string{},
+			testutils.BuildTestRESTMapper(),
+		)
+		require.NotNil(t, binder)
+
+		_, err := binder.search()
```
pedjak, line 505: `binder.search()` is invoked in `binder.bind()`, so not sure if we really needed here?

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -108,7 +108,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	// fetch and validate namespaced ServiceBindingRequest instance
 	sbr, err := r.getServiceBindingRequest(request.NamespacedName)
 	if err != nil {
-		if errors.Is(err, applicationNotFound) {
+		if k8sapierrors.IsNotFound(err) {
```
pedjak, line 111: We need an e2e tests checking this.

pkg/controller/servicebindingrequest/binder_test.go
qibobo: Previously there is no UT for bind and unbind. I add them and have added more descriptions.

pkg/controller/servicebindingrequest/binder_test.go
qibobo: Add a comment for this. We need to ensure there is no error in search().

qibobo:
@Avni-Sharma @pedjak add the UT and description.

qibobo:
issue has been fixed.

PR #533, opened 2020-06-30
fixed CI unit jobs

pedjak:
it looks like that recently codecov bash script we pull from https://codecov.io/bash got updated,
breaking 'upload-codecov-report' make rule. The token we use is read from
`/usr/local/redhat-developer-service-binding-operator-codecov-token/token`
and that file is mounted from a read-only volume. The codecov script tries to write something on that volume and it fails.

In order to fix the issue, we read token uuid within Makefile and pass it to the codecov script


DhritiShikhar:
Is there a JIRA ticket for this @pedjak ?

pedjak:
> Is there a JIRA ticket for this @pedjak ?

Not yet @DhritiShikhar - was just trying to debug the issue here first.

pratikjagrut:
Looks like unit tests are working again. Still running once again.
/test 4.5-unit
/test unit

PR #550, opened 2020-07-10
WIP: Fix #549 Race condition between operator's watch and e2eutil.WaitForDeployment

### LINKED ISSUE #549: Race condition between operator's watch and e2eutil.WaitForDeployment.
I've noticed that if we create SBR before the application and then the `operator is watching for the application deployment to be ready` and at the same time `e2eutil.WaitForDeployment(t, f.KubeClient, ns, appName, 1, retryInterval, timeout) is also watching the same deployment`. So here these two watches are in deadlock or sort of deadlock condition if the operator watch detects the deployment is ready before e2eutil watch then it will redeploy it with the secrets hence the e2eutil watch might miss the first deployment readiness and will still wait for deployment this time deployment will take some more time to reach 1 replica ensuing e2eutil watch timeout. Hence failing the e2e tests.

pratikjagrut:
### Motivation

Fix #549 

### Changes

Add waitForDeployment inside require.Eventually().

### Testing

`make test-e2e`

### DIFF
--- a/test/e2e/servicebindingrequest_test.go
+++ b/test/e2e/servicebindingrequest_test.go
@@ -44,7 +44,7 @@ const (
 
 var (
 	retryInterval  = time.Second * 5
-	timeout        = time.Second * 180
+	timeout        = time.Second * 120
 	cleanupTimeout = time.Second * 5
 
 	// Intermediate secret should have following data
@@ -331,10 +331,14 @@ func CreateApp(
 
 	// waiting for application deployment to reach one replica
 	t.Log("Waiting for application deployment reach one replica...")
-	require.NoError(
-		t,
-		e2eutil.WaitForDeployment(t, f.KubeClient, ns, appName, 1, retryInterval, timeout),
-	)
+	require.Eventually(t, func() bool {
+		err := e2eutil.WaitForDeployment(t, f.KubeClient, ns, appName, 1, retryInterval, timeout)
+		if err != nil {
+			t.Log(err)
+			return false
+		}
+		return true
+	}, 120*time.Second, 2*time.Second)
 
 	// retrieveing deployment, to inspect its contents
 	t.Logf("Reading application deployment '%s'", appName)


test/e2e/servicebindingrequest_test.go
```diff
@@ -331,10 +331,14 @@ func CreateApp(
...
-		e2eutil.WaitForDeployment(t, f.KubeClient, ns, appName, 1, retryInterval, timeout),
-	)
+	require.Eventually(t, func() bool {
+		err := e2eutil.WaitForDeployment(t, f.KubeClient, ns, appName, 1, retryInterval, timeout)
+		if err != nil {
+			t.Log(err)
+			return false
+		}
+		return true
+	}, 120*time.Second, 2*time.Second)
```
isutton, line 341: In `WaitForDeployment`, it makes use of `wait.Poll` to achieve similar behavior[1]; `WaitForDeployment` will return after `timeout` seconds, which seems to be the same value as `require.Eventually` is using. This means both would return virtually at the same time.

https://github.com/redhat-developer/service-binding-operator/blob/a3e812baca0feefe31a5dae7f9da072894568b06/vendor/github.com/operator-framework/operator-sdk/pkg/test/e2eutil/wait_util.go#L48-L76

Further investigation in `assert.Eventually`[2] shows that every tick (in this case `2*time.Second`) the condition function will be executed *inside a gofunc*; in other words, this change can potentially create many go funcs that will be orphaned in the case the condition is never met in `WaitForDeployment`.

https://github.com/redhat-developer/service-binding-operator/blob/97172e464755bbf7e6220281a81bd4c168ff6aee/vendor/github.com/stretchr/testify/assert/assertions.go#L1635-L1662

Additionally, given that `WaitForDeployment`'s `retryInterval` (5s) is greater than the `Eventually` retry interval (2s), in the best scenario at least two `WaitForDeployment` will be racing.

pratikjagrut: > Further investigation in `assert.Eventually`[2] shows that every tick (in this case `2*time.Second`) the condition function will be executed _inside a gofunc_; in other words, this change can potentially create many go funcs that will be orphaned in the case the condition is never met in `WaitForDeployment`.

My theory was, as Eventually creates goroutines then waitForDeployment will be running concurrently in several goroutines created on each tick and this could increase the chance of detecting deployment readiness than detecting deployment by `single waitForDeployment`. There is a possibility that these goroutines could be orphaned but they will be returned or exited once the `waitForDeployment` timeouts.

> Additionally, given that `WaitForDeployment`'s `retryInterval` (5s) is greater than the `Eventually` retry interval (2s), in the best scenario at least two `WaitForDeployment` will be racing.

On each tick (per 2sec) new goroutine will be created in 120 sec there will be 60 goroutines(a bit high number) which will run for 120Sec each(timeout of `waitForDeployment`). Some goroutines could run after the Eventually is timed-out or met the condition, and probably if I'm not wrong all the goroutines will be exited by 240Sec(if Eventually orphans the gofunc) and we just need one goroutine to meet the condition.

This was my theory for trying out this nested polling sort of thing. Yeah, this seems bit costlier approach.

Or this racing condition could be solved by #523 and also we would need to make single write for patching and annotating deployment.

Any suggestion for other approaches?

pratikjagrut:
Closing this PR in favour of acceptance test.

Issue #542, opened 2020-07-07
Unable to delete a service binding request in spite of setting "ownerReferences"


dharmit:
### Steps to reproduce

1. Create an application based on [this repo](https://github.com/akashshinde/node-todo) using the "Topology" view of DevConsole.
2. Create an EtcdCluster from the Operator, again using "Topology"
3. Get the uid of the node application deployed in Step 1 - `oc get deployment/node-todo -o template="{{.metadata.uid}}"`.
4. Create an SBR with uid from previous step as `ownerReferences` like below:
    ```yaml
    apiVersion: apps.openshift.io/v1alpha1
    kind: ServiceBindingRequest
    metadata:
      name: binding-request
      namespace: myproject
      ownerReferences:
      - apiVersion: apps/v1
        kind: Deployment
        name: node-todo 
        uid: 9075567a-a948-419a-ac28-639f4ca8a2b0
    spec:
      applicationSelector:
        group: apps
        version: v1
        resource: deployments
        resourceRef: node-todo
      backingServiceSelector:
        group: etcd.database.coreos.com
        kind: EtcdCluster
        version: v1beta2
        resourceRef: example
      mountPathPrefix: ""
      customEnvVar: []
      detectBindingResources: true
    ```
5. Delete the `node-todo` deployment using `oc delete deployment/node-todo`.

### Actual behaviour
SBR created using above steps doesn't get deleted in spite of setting `ownerReferences`.

### Expected behaviour
SBR should get deleted

### Note

* If I delete the application using "Topology" view, SBR does get deleted.
* Being able to automatically delete an SBR is important for `odo` because as a part of `odo link`, we are creating an SBR with `ownerReferences` info added (this is not available in `odo`'s master branch; it's on a feature branch on my fork). Upon doing `odo delete`, `odo` deletes the deployment and all the k8s objects that have the deployment set in their `ownerReferences` but SBR doesn't get deleted.

DhritiShikhar:
When an application is deleted, ideally a Service Binding Request shouldn't get deleted.

But yes, if we set an ownerReference in SBR explicitly (as you have shown in above), then the SBR should get deleted.

DhritiShikhar:
@dharmit 

[...]

PR #551, opened 2020-07-10
Delete sbr if it's ownerRef resource is deleted

Avni-Sharma:
### Motivation

Fix issue #542 
dependent on #523 , #498 

### Testing

Please follow steps https://github.com/redhat-developer/service-binding-operator/issues/542#issue-652029225 manually
need to add an e2e test case since garbage collection is involved here. e2e can be taken up after the acceptance tests are in place. 

### DIFF
--- a/pkg/controller/servicebindingrequest/reconciler.go
+++ b/pkg/controller/servicebindingrequest/reconciler.go
@@ -205,7 +205,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		logger.Debug("Removing resource finalizers...")
 		sbr.SetFinalizers(removeStringSlice(sbr.GetFinalizers(), finalizer))
 		if _, err := updateServiceBindingRequest(r.dynClient, sbr); err != nil {
-			return noRequeue(err)
+			return requeueError(err)
 		}
 		return done()
 	}
--- a/pkg/controller/servicebindingrequest/reconciler.go
+++ b/pkg/controller/servicebindingrequest/reconciler.go
@@ -200,6 +200,15 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		logger.Error(err, "Error add watching application GVR")
 	}
 
+	if sbr.GetDeletionTimestamp() != nil && sbr.GetOwnerReferences() != nil {
+		logger := logger.WithName("Deleting SBR when it has ownerReference")
+		logger.Debug("Removing resource finalizers...")
+		sbr.SetFinalizers(removeStringSlice(sbr.GetFinalizers(), finalizer))
+		if _, err := updateServiceBindingRequest(r.dynClient, sbr); err != nil {
+			return noRequeue(err)
+		}
+		return done()
+	}
 	if sbr.GetDeletionTimestamp() != nil {
 		logger := logger.WithName("unbind")
 		logger.Info("Executing unbinding steps...")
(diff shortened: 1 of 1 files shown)


pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -200,6 +200,15 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		logger.Error(err, "Error add watching application GVR")
 	}
 
+	if sbr.GetDeletionTimestamp() != nil && sbr.GetOwnerReferences() != nil {
+		logger := logger.WithName("Deleting SBR when it has ownerReference")
+		logger.Debug("Removing resource finalizers...")
+		sbr.SetFinalizers(removeStringSlice(sbr.GetFinalizers(), finalizer))
```
isutton, line 206: #498 introduces a function to remove finalizers, perhaps cherry-picking the commit if possible will make life easier.

pkg/controller/servicebindingrequest/reconciler.go
```diff
@@ -200,6 +200,15 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+		if _, err := updateServiceBindingRequest(r.dynClient, sbr); err != nil {
+			return noRequeue(err)
```
isutton, line 208: In the case the update doesn't go through for any reason, the SBR should be requeued.

pkg/controller/servicebindingrequest/reconciler.go
Avni-Sharma: > Other than the review comments on the code, I'd like to understand more the rationale of this change; it seems that the unbind process is not executed before removing the finalizer and updating the SBR.
> 
> Perhaps the feature scenario here was related to an application owning the SBR, but this is not the only ownership scenario one might have as any other arbitrary resource can be used as resource reference and those should be taken in consideration.

@isutton , why unbind particularly failed here is because unbind() https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/controller/servicebindingrequest/binder.go#L578 was trying to find objects and the app is deleted previously and provides a deletionTimeStamp on SBR as well.  For any other application issue, it is tackled here https://github.com/redhat-developer/service-binding-operator/pull/498/files#diff-fae660884df4e35af60a04c5ddf930daR291 but in this case, this app is not owning the sbr. So I can mark this PR as dependent on #498 . But we do need a condition of sbr.GetDeletionTimestamp() != nil && sbr.GetOwnerReferences() when the app is owning the sbr

Avni-Sharma: As far as #542  is concerned, it isn't really a feature request. If we want a bigger refactor of unbind and tackling various scenarios, which is out of the scope of #542  then we should maybe have a Cabal discussion for the same, and have a separate refactor for unbind story. For now ODO folks are blocked on this.

sbose78: +1 . This feels like a bug because we are potentially messing with the usual `ownerReference` expectations otherwise ?

pkg/controller/servicebindingrequest/reconciler.go
Avni-Sharma: Done https://github.com/redhat-developer/service-binding-operator/pull/551/commits/aea0e8b950ab486e265f6d14deee8b8dd4859d9f

pkg/controller/servicebindingrequest/reconciler.go
Avni-Sharma: @isutton  I have reused the function added by #498

Issue #508, opened 2020-06-09
Controller doesn't reconcile on changes to BackingService CR


sbose78:
It's been reported that 

"We use ServiceBindingRequests to bind services like kafka/databases to microservices. In the kafka example, because a cluster can take up to a few minutes to come up and make its address available to the Service Binding operator, the secrets created by the SBO are empty. It is apparent that the SBO does not watch CRDs (like kafka here) and update/recreate the secrets when the expected data is available. "

"I think in this case, the status field of the Kafka CR is getting updated once the kafka cluster is up and running, but then SBO doesn’t get notified about the update. and therefore the binding secret doesn’t include the updated kafka address"

Taken out from https://github.com/redhat-developer/service-binding-operator/pull/442#issuecomment-637421911

sbose78:
* Initial implementation : https://github.com/redhat-developer/service-binding-operator/pull/162 used to watch all GVKs associated with CRDs owned by CSVs.
* In master, the controller watches all [GVKs](https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/controller/servicebindingrequest/sbrcontroller.go#L154 ) including those [owned by CSVs ](https://github.com/redhat-developer/service-binding-operator/blob/70cdc2caf9ac0d227353fd0f016f8b2b3e44b7b1/pkg/controller/servicebindingrequest/sbrcontroller.go#L92)

sbose78:
cc @pushkarnk @navidsh for more info.

DhritiShikhar:
### How to test

1. Follow the steps given in this [example](https://github.com/redhat-developer/service-binding-operator/tree/master/examples/nodejs_postgresql) to:
- install DB operator, Service Binding Operator
- Import an application
- Create a DB instance 
- Create a ServiceBindingRequest

Binding should be successful.

2. Delete DB instance 

```
oc delete deployment db-demo
```

3. Recreate the DB instance with the same name. Refer the [steps](https://github.com/redhat-developer/service-binding-operator/tree/master/examples/nodejs_postgresql#create-a-db-instance-for-the-application)

With the new backingservice, the service credential is changed. But the injected env in application containers is not updated.

Avni-Sharma:
> ### How to test
> 1. Follow the steps given in this [example](https://github.com/redhat-developer/service-binding-operator/tree/master/examples/nodejs_postgresql) to:
> 
> * install DB operator, Service Binding Operator
> * Import an application
> * Create a DB instance
[...]

PR #556, opened 2020-07-13
Temporarily disable e2e scenarios not supported by SBO.

pmacik:
### Motivation

As described in https://github.com/redhat-developer/service-binding-operator/issues/549#issuecomment-657500353 by @isutton:
>Currently when the application is created after the SBR, even if the GVK of the application is being observed (corev1.Deployment for example), there is no way to correlate the application resource and the SBR to be reconciled since there are no annotations in the application resource yet (the annotation is created after the first reconciliation).
In other words the scenarios where the Application is created after the SBR are currently not supported by the SBO and the tests that implement those passes or fails randomly by cheer chance.

### Changes

e2e scenarios where the application is created after SBR are disabled to un-block all the other PR that are blocked by those e2e scenarios failing. Those will be enabled again as soon as the issue is fixed.

### Testing

`make test-e2e`

### DIFF
--- a/test/e2e/servicebindingrequest_test.go
+++ b/test/e2e/servicebindingrequest_test.go
@@ -93,18 +93,22 @@ func TestAddSchemesToFramework(t *testing.T) {
 		t.Run("scenario-app-db-sbr", func(t *testing.T) {
 			ServiceBindingRequest(t, []Step{AppStep, DBStep, SBRStep})
 		})
+		/* Currently disabled as not supported by SBO
 		t.Run("scenario-db-sbr-app", func(t *testing.T) {
 			ServiceBindingRequest(t, []Step{DBStep, SBRStep, AppStep})
 		})
+		*/
 		t.Run("scenario-app-sbr-db", func(t *testing.T) {
 			ServiceBindingRequest(t, []Step{AppStep, SBRStep, DBStep})
 		})
+		/* Currently disabled as not supported by SBO
 		t.Run("scenario-sbr-db-app", func(t *testing.T) {
 			ServiceBindingRequest(t, []Step{SBRStep, DBStep, AppStep})
 		})
 		t.Run("scenario-sbr-app-db", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


isutton:
I've made similar change in c552b9048d0a55fe1bc4e96a109343c316abbcbe with the benefit the tests will still be compiled but not executed at runtime.

pmacik:
@isutton Cool! I like your solution even better! let's use yours.
Looking forward to a PR ;)

sbose78:
> the scenarios where the Application is created after the SBR 

Does this imply,
* The SBR has the "Application" specified ..
* but the workload is absent at the time of creation of the SBR
* When the workloads shows up, the SBR still doesn't succeed in making the binding ?

That looks like a bug to me.

In such a scenario, if the SBR has the application specified && when the workload shows up on the cluster later, binding **should** succeed because the reconcile look shouldn't end till that happens.

cc @DhritiShikhar

Avni-Sharma:
@sbose78 yup, that is what is happening here as well https://github.com/redhat-developer/service-binding-operator/issues/508.

sbose78:
#508 has to be watch or event-driven reconciliation after the first time successful reconciliation.

In this case, if the specified application wasn't found, then the reconcile loop needs to keep looking for the specific resource till it is found. This hasn't got anything to do with watching of workloads.

isutton:
> #508 has to be watch or event-driven reconciliation after the first time successful reconciliation.
> 
> 
> 
> In this case, if the specified application wasn't found, then the reconcile loop needs to keep looking for the specific resource till it is found. This hasn't got anything to do with watching of workloads.

The fact that a resource doesn't exist now doesn't mean it will exist in the future. 

Polling is a decent workaround on the short time, but since the platform itself gives means for users to  spare cycles by waiting for relevant change notifications there's no reason to leave iron out the event handling workflow for later. 

The objective is to reduce as much as possible the writes in related resources to avoid excessive usage of API server resources, where getting rid of the relationship annotations has a play, reducing from 1+N writes (without considering changes made to make the binding effective, such as including the envFrom value in a Deployment) to 1 write only.

PR #567, opened 2020-07-17
Acceptance tests-quarkus_posgresql_as_knative_service

shvasude:
### Motivation

https://issues.redhat.com/browse/APPSVC-560

### Changes

quarkus_postgres scenario implemented as knative service:

> Scenario: Bind an imported quarkus app which is deployed as knative service to PostgreSQL database
>         Given Openshift Serverless Operator is running
>         * Quarkus s2i builder image is present
>         * Knative serving is running
>         * DB "db-demo-knative" is running

### Testing

make test-acceptance

### DIFF
--- a/test/acceptance/features/bindKnativeService.feature
+++ b/test/acceptance/features/bindKnativeService.feature
@@ -1,0 +1,45 @@
+Feature: Bind knative service to a service
+
+  As a user of Service Binding Operator
+  I want to bind knative service to services it depends on
+
+  Background:
+    Given Namespace [TEST_NAMESPACE] is used
+    * Service Binding Operator is running
+    * PostgreSQL DB operator is installed
+
+  Scenario: Bind an imported quarkus app which is deployed as knative service to PostgreSQL database
+    Given Openshift Serverless Operator is running
+    * Knative serving is running
+    * DB "db-demo-knative" is running
+    * Quarkus application "knative-app" is imported as Knative service
+    When Service Binding Request is applied to connect the database and the application
+      """
+            apiVersion: apps.openshift.io/v1alpha1
+            kind: ServiceBindingRequest
+            metadata:
+              name: binding-request-knative
+            spec:
+              applicationSelector:
+                group: serving.knative.dev
+                version: v1beta1
+                resource: services
+                resourceRef: knative-app
+              backingServiceSelector:
+                group: postgresql.baiju.dev
+                version: v1alpha1
+                kind: Database
+                resourceRef: db-demo-knative
+                id: knav
+              customEnvVar:
+                - name: JDBC_URL
+                  value: jdbc:postgresql://{{ .knav.status.dbConnectionIP }}:{{ .knav.status.dbConnectionPort }}/{{ .knav.status.dbName }}
+                - name: DB_USER
+                  value: "{{ .knav.status.dbCredentials.user }}"
+                - name: DB_PASSWORD
+                  value: "{{ .knav.status.dbCredentials.password }}"
+      """
+    Then deployment must contain intermediate secret "binding-request-knative"
+    Then application should be connected to the DB "db-demo-knative"
+    And jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding Request "binding-request-knative" should be changed to "True"
+    And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding Request "binding-request-knative" should be changed to "True"
--- a/test/acceptance/features/steps/knative_serving.py
+++ b/test/acceptance/features/steps/knative_serving.py
@@ -1,0 +1,39 @@
+import re
+from command import Command
+from openshift import Openshift
+
+
+class KnativeServing():
+
+    openshift = Openshift()
+    cmd = Command()
+
+    name = ""
+    namespace = ""
+
+    def __init__(self, name="knative-serving", namespace="knative-serving"):
+        self.name = name
+        self.namespace = namespace
+        self.knative_serving_yaml_template = '''
+apiVersion: operator.knative.dev/v1alpha1
+kind: KnativeServing
+metadata:
+  name: {name}
+  namespace: {namespace}
+'''
+
+    def is_present(self):
+        cmd = f'oc get ks {self.name} -n {self.namespace}'
+        output, exit_code = self.cmd.run(cmd)
+        if exit_code != 0:
+            print(f"cmd-{cmd} result for getting available knative serving is {output} with the exit code {exit_code}")
+            return False
+        return True
+
+    def create(self):
+        serving_output = self.openshift.oc_apply(self.knative_serving_yaml_template.format(name=self.name, namespace=self.namespace))
+        pattern = f'knativeserving.(serving|operator).knative.dev/{self.namespace}\\screated'
+        if re.search(pattern, serving_output):
+            return True
+        print(f"Pattern {pattern} did not match as creating knative serving yaml output is {serving_output}")
+        return False
--- a/test/acceptance/features/steps/namespace.py
+++ b/test/acceptance/features/steps/namespace.py
@@ -9,25 +9,26 @@ class Namespace():
         self.cmd = Command()
 
     def create(self):
-        create_namespace_output, exit_code = self.cmd.run("oc new-project {}".format(self.name))
-        if re.search(r'Now using project \"%s\"\son\sserver' % self.name, create_namespace_output) or \
-                re.search(r'.*Already\son\sproject\s\"%s\"\son\sserver.*' % self.name, create_namespace_output):
+        create_namespace_output, exit_code = self.cmd.run(f"oc new-project {self.name}")
+        if re.search(r'Now using project \"%s\"\son\sserver' % self.name, create_namespace_output) is not None or \
+                re.search(r'.*Already\son\sproject\s\"%s\"\son\sserver.*' % self.name, create_namespace_output) is not None:
             return True
-        elif re.search(r'.*project.project.openshift.io\s\"%s\"\salready exists' % self.name, create_namespace_output):
+        elif re.search(r'.*project.project.openshift.io\s\"%s\"\salready exists' % self.name, create_namespace_output) is not None:
             return self.switch_to()
         else:
-            print("Returned a different value {}".format(create_namespace_output))
+            print(f"Unexpected value returned '{create_namespace_output}'")
         return False
 
     def is_present(self):
-        output, exit_code = self.cmd.run('oc get ns {}'.format(self.name))
+        output, exit_code = self.cmd.run(f'oc get ns {self.name}')
         return exit_code == 0
 
     def switch_to(self):
-        create_namespace_output, exit_code = self.cmd.run('oc project {}'.format(self.name))
-        if re.search(r'Now using project \"%s\"\son\sserver' % self.name, create_namespace_output):
+        create_namespace_output, exit_code = self.cmd.run(f'oc project {self.name}')
+        if re.search(r'Now using project \"%s\"\son\sserver' % self.name, create_namespace_output) is not None:
             return True
-        elif re.search(r'.*Already\son\sproject\s\"%s\"\son\sserver.*' % self.name, create_namespace_output):
+        elif re.search(r'.*Already\son\sproject\s\"%s\"\son\sserver.*' % self.name, create_namespace_output) is not None:
             return True
         else:
-            return False
+            print(f"Unexpected namespace creating output: '{create_namespace_output}'")
+        return False
--- a/test/acceptance/features/steps/openshift.py
+++ b/test/acceptance/features/steps/openshift.py
@@ -3,8 +3,6 @@ import time
 from pyshould import should
 from command import Command
 
-nodejs_app = "https://github.com/pmacik/nodejs-rest-http-crud"
-
 
 class Openshift(object):
     def __init__(self):
@@ -38,13 +36,61 @@ spec:
     sourceNamespace: openshift-marketplace
     startingCSV: '{csv_version}'
 '''
+        self.image_stream_template = '''
+---
+apiVersion: image.openshift.io/v1
+kind: ImageStream
+metadata:
+    name: {name}
+    namespace: {namespace}
+'''
+        self.build_config_template = '''
+---
+apiVersion: build.openshift.io/v1
+kind: BuildConfig
+metadata:
+  name: {name}
+  namespace: {namespace}
+spec:
+  source:
+    git:
+      ref: master
+      uri: {application_source}
+    type: Git
+  strategy:
+    sourceStrategy:
+      from:
+        kind: ImageStreamTag
+        name: {image_name_with_tag}
+        namespace: openshift
+    type: Source
+  output:
+    to:
+      kind: ImageStreamTag
+      name: {name}:latest
+  triggers:
+    - type: ConfigChange
+'''
+        self.service_template = '''
+---
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
+  name: {name}
+  namespace: {namespace}
+spec:
+  template:
+    spec:
+      containers:
+        - image: {image_repository}
+'''
 
     def get_pod_lst(self, namespace):
         return self.get_resource_lst("pods", namespace)
 
     def get_resource_lst(self, resource_plural, namespace):
-        (output, exit_code) = self.cmd.run(f'oc get {resource_plural} -n {namespace} -o "jsonpath={{.items[*].metadata.name}}"')
-        exit_code | should.be_equal_to(0)
+        output, exit_code = self.cmd.run(f'oc get {resource_plural} -n {namespace} -o "jsonpath={{.items[*].metadata.name}}"')
+        assert exit_code == 0, f"Getting resource list failed as the exit code is not 0 with output - {output}"
         return output
 
     def search_item_in_lst(self, lst, search_pattern):
@@ -148,26 +194,35 @@ spec:
         return re.search(r'.*%s\sexposed' % service_name, output)
 
     def get_route_host(self, name, namespace):
-        (output, exit_code) = self.cmd.run(f'oc get route {name} -n {namespace} -o "jsonpath={{.status.ingress[0].host}}"')
-        exit_code | should.be_equal_to(0)
+        output, exit_code = self.cmd.run(f'oc get route {name} -n {namespace} -o "jsonpath={{.status.ingress[0].host}}"')
+        assert exit_code == 0, f"Getting route host failed as the exit code is not 0 with output - {output}"
         return output
 
-    def check_for_deployment_status(self, deployment_name, namespace, wait_for_status="True"):
-        deployment_status_cmd = f'oc get deployment {deployment_name} -n {namespace} -o "jsonpath={{.status.conditions[*].status}}"'
-        deployment_status, exit_code = self.cmd.run_wait_for_status(deployment_status_cmd, wait_for_status, 5, 300)
-        exit_code | should.be_equal_to(0)
-        return deployment_status
+    def get_deployment_status(self, deployment_name, namespace, wait_for_status=None):
+        deployment_status_cmd = f'oc get deployment {deployment_name} -n {namespace} -o json' \
+            + ' | jq -rc \'.status.conditions[] | select(.type=="Available").status\''
+        output = None
+        exit_code = -1
+        if wait_for_status is not None:
+            status_found, output, exit_code = self.cmd.run_wait_for_status(deployment_status_cmd, wait_for_status, 5, 400)
+            if exit_code == 0:
+                assert status_found is True, f"Deployment {deployment_name} result after waiting for status is {status_found}"
+        else:
+            output, exit_code = self.cmd.run(deployment_status_cmd)
+        assert exit_code == 0, "Getting deployment status failed as the exit code is not 0"
+
+        return output
 
     def get_deployment_env_info(self, name, namespace):
         env_cmd = f'oc get deploy {name} -n {namespace} -o "jsonpath={{.spec.template.spec.containers[0].env}}"'
         env, exit_code = self.cmd.run(env_cmd)
-        exit_code | should.be_equal_to(0)
+        assert exit_code == 0, f"Non-zero exit code ({exit_code}) returned while getting deployment's env: {env}"
         return env
 
     def get_deployment_envFrom_info(self, name, namespace):
         env_from_cmd = f'oc get deploy {name} -n {namespace} -o "jsonpath={{.spec.template.spec.containers[0].envFrom}}"'
         env_from, exit_code = self.cmd.run(env_from_cmd)
-        exit_code | should.be_equal_to(0)
+        assert exit_code == 0, f"Non-zero exit code ({exit_code}) returned while getting deployment's envFrom: {env_from}"
         return env_from
 
     def get_resource_info_by_jsonpath(self, resource_type, name, namespace, json_path, wait=False, interval=5, timeout=120):
@@ -193,3 +248,88 @@ spec:
                     time.sleep(interval)
         exit_code | should.be_equal_to(0).desc(f'Exit code should be 0:\n OUTPUT:\n{output}')
         return output.rstrip("\n")
+
+    def create_image_stream(self, name, registry_namespace):
+        image_stream = self.image_stream_template.format(name=name, namespace=registry_namespace)
+        return self.oc_apply(image_stream)
+
+    def get_docker_image_repository(self, name, namespace):
+        cmd = f'oc get is {name} -n {namespace} -o "jsonpath={{.status.dockerImageRepository}}"'
+        (output, exit_code) = self.cmd.run(cmd)
+        assert exit_code == 0, f"cmd-{cmd} result for getting docker image repository is {output} with exit code-{exit_code} not equal to 0"
+        return output
+
+    def create_build_config(self, name, namespace, application_source, image_name_with_tag):
+        build_config_yaml = self.build_config_template.format(
+            name=name, namespace=namespace, application_source=application_source, image_name_with_tag=image_name_with_tag)
+        return self.oc_apply(build_config_yaml)
+
+    def create_knative_service(self, name, namespace, image):
+        knative_service_yaml = self.service_template.format(name=name, namespace=namespace, image_repository=image)
+        return self.oc_apply(knative_service_yaml)
+
+    def wait_for_build_pod_status(self, build_pod_name, namespace, wait_for_status="Succeeded", timeout=780):
+        cmd = f'oc get pod {build_pod_name} -n {namespace} -o "jsonpath={{.status.phase}}"'
+        status_found, output, exit_status = self.cmd.run_wait_for_status(cmd, wait_for_status, timeout=timeout)
+        return status_found, output
+
+    def get_deployment_name_in_namespace(self, deployment_name_pattern, namespace, wait=False, interval=5, timeout=120):
+        if wait:
+            start = 0
+            while ((start + interval) <= timeout):
+                deployment = self.search_resource_in_namespace("deployment", deployment_name_pattern, namespace)
+                if deployment is not None:
+                    return deployment
+                time.sleep(interval)
+                start += interval
+            return None
+        else:
+            return self.search_resource_in_namespace("deployment", deployment_name_pattern, namespace)
+
+    def get_knative_route_host(self, name, namespace):
+        cmd = f'oc get rt {name} -n {namespace} -o "jsonpath={{.status.url}}"'
+        output, exit_code = self.cmd.run(cmd)
+        assert exit_code == 0, f"cmd-{cmd} result for getting knative route is {output} with exit code not equal to 0"
+        return output
+
+    def get_revisions(self, namespace):
+        return self.get_resource_lst("rev", namespace)
+
+    def get_last_revision_status(self, revision, namespace):
+        cmd = f'oc get rev {revision} -n {namespace} -o "jsonpath={{.status.conditions[*].status}}"'
+        (output, exit_code) = self.cmd.run(cmd)
+        assert exit_code == 0, f"cmd-{cmd} for getting last revision status is {output} with exit code not equal to 0"
+        last_revision_status = output.split(" ")[-1]
+        return last_revision_status
+
+    def get_resource_list_in_namespace(self, resource_plural, name_pattern, namespace):
+        print(f"Searching for {resource_plural} that matches {name_pattern} in {namespace} namespace")
+        lst = self.get_resource_lst(resource_plural, namespace)
+        if len(lst) != 0:
+            print("Resource list is {}".format(lst))
+            return self.get_all_matched_pattern_from_lst(lst, name_pattern)
+        else:
+            print('Resource list is empty under namespace - {}'.format(namespace))
+            return None
+
+    def get_all_matched_pattern_from_lst(self, lst, search_pattern):
+        lst_arr = lst.split(" ")
+        output_arr = []
+        for item in lst_arr:
+            if re.fullmatch(search_pattern, item) is not None:
+                print(f"item matched {item}")
+                output_arr.append(item)
+        if not output_arr:
+            print("Given item not matched from the list of pods")
+            return None
+        else:
+            return output_arr
+
+    def search_resource_lst_in_namespace(self, resource_plural, name_pattern, namespace):
+        print(f"Searching for {resource_plural} that matches {name_pattern} in {namespace} namespace")
+        lst = self.get_resource_list_in_namespace(resource_plural, name_pattern, namespace)
+        if len(lst) != 0:
+            print("Resource list is {}".format(lst))
+            return lst
+        print('Resource list is empty under namespace - {}'.format(namespace))
+        return None
--- a/test/acceptance/features/steps/quarkus_application.py
+++ b/test/acceptance/features/steps/quarkus_application.py
@@ -1,0 +1,140 @@
+from command import Command
+import re
+import time
+import requests
+from openshift import Openshift
+
+
+class QuarkusApplication(object):
+
+    cmd = Command()
+
+    image_name_with_tag = "quay.io/pmacik/using-spring-data-jqa-quarkus:latest"
+    api_end_point = '{route_url}/api/status/dbNameCM'
+    openshift = Openshift()
+
+    name = ""
+    namespace = ""
+    deployment_name_pattern = "{name}-\\w+-deployment"
+
+    def __init__(self, name, namespace):
+        self.name = name
+        self.namespace = namespace
+
+    def install(self):
+        knative_service_output = self.openshift.create_knative_service(self.name, self.namespace, self.image_name_with_tag)
+        output = re.search(r'.*service.serving.knative.dev/%s\s(created|configured|unchanged)' % self.name, knative_service_output)
+        assert output is not None, f"Knative serving is not created as the result is {knative_service_output}"
+        return True
+
+    def get_pod_name_running(self, pod_name_pattern, wait=False):
+        if wait:
+            pod_name = self.openshift.wait_for_pod(self.format_pattern(pod_name_pattern), self.namespace, timeout=500)
+        else:
+            pod_name = self.openshift.search_pod_in_namespace(self.format_pattern(pod_name_pattern), self.namespace)
+        return pod_name
+
+    def is_imported(self):
+        deployment_name = self.openshift.get_deployment_name_in_namespace(
+            self.format_pattern(self.deployment_name_pattern), self.namespace, wait=True, timeout=400)
+        if deployment_name is None:
+            return False
+        else:
+            deployment_status = self.openshift.get_deployment_status(deployment_name, self.namespace, wait_for_status="True")
+            print(f"The deployment {deployment_name} status is {deployment_status}")
+            output_match = re.search(r'True', deployment_status)
+            assert output_match is not None, "Matched deployment status is not True"
+            return True
+
+    def get_db_name_from_api(self, wait=False, interval=5, timeout=300):
+        route_url = self.openshift.get_knative_route_host(self.name, self.namespace)
+        if route_url is None:
+            return None
+        if wait:
+            start = 0
+            while ((start + interval) <= timeout):
+                url = self.api_end_point.format(route_url=route_url)
+                db_name = requests.get(url)
+                if db_name.status_code == 200:
+                    return db_name.text
+                time.sleep(interval)
+                start += interval
+        else:
+            url = self.api_end_point.format(route_url=route_url)
+            db_name = requests.get(url)
+            if db_name.status_code == 200:
+                return db_name.text
+        return None
+
+    def get_observed_generation(self):
+        deployment_name = self.openshift.get_deployment_name_in_namespace(self.format_pattern(self.deployment_name_pattern), self.namespace)
+        return self.openshift.get_resource_info_by_jsonpath("deployment", deployment_name, self.namespace, "{.status.observedGeneration}")
+
+    def format_pattern(self, pattern):
+        return pattern.format(name=self.name)
+
+    def get_redeployed_rev_name(self, old_rev_name, interval=5, timeout=300):
+        start = 0
+        while ((start + interval) <= timeout):
+            revisions = self.openshift.get_revisions(self.namespace)
+            for rev in revisions.split(" "):
+                if rev != old_rev_name and re.match(self.name, rev) is not None:
+                    new_revision = self.openshift.get_last_revision_status(rev, self.namespace)
+                    if new_revision == 'True':
+                        return rev
+            time.sleep(interval)
+            start += interval
+        return None
+
+    def get_rev_name_redeployed_by_generation(self, old_generation, interval=5, timeout=300):
+        start = 0
+        while ((start + interval) <= timeout):
+            current_generation = self.get_generation()
+            revisions = self.openshift.get_revisions(self.namespace)
+            for rev in revisions.split(" "):
+                if (current_generation > old_generation) and (re.match(self.name, rev) is not None):
+                    new_revision = self.openshift.get_last_revision_status(rev, self.namespace)
+                    if new_revision == 'True':
+                        return rev
+            time.sleep(interval)
+            start += interval
+        return None
+
+    def get_generation(self):
+        deployment_name = self.openshift.get_deployment_name_in_namespace(self.format_pattern(self.deployment_name_pattern), self.namespace)
+        return self.openshift.get_resource_info_by_jsonpath("deployment", deployment_name, self.namespace, "{.metadata.generation}")
+
+    def get_deployment_names(self):
+        return self.openshift.search_resource_lst_in_namespace("deployment", self.format_pattern(self.deployment_name_pattern), self.namespace)
+
+    def get_deployment_with_intermediate_secret(self, intermediate_secret_name, wait=False, interval=5, timeout=300):
+        expected_secretRef_1 = f'[{{"secretRef":{{"name":"{intermediate_secret_name}"}}}}]'
+        expected_secretRef_2 = f'[map[secretRef:map[name:{intermediate_secret_name}]]]'
+        deployment_name_pattern = self.format_pattern(self.deployment_name_pattern)
+        if wait:
+            start = 0
+            while ((start + interval) <= timeout):
+                deployment_list = self.get_deployment_names()
+                if deployment_list is not None:
+                    for deployment in deployment_list:
+                        result = self.openshift.get_deployment_envFrom_info(deployment, self.namespace)
+                        if result == expected_secretRef_1 or result == expected_secretRef_2:
+                            return deployment
+                        else:
+                            print(f"\nUnexpected deployment's envFrom info: \nExpected: {expected_secretRef_1} or {expected_secretRef_2} \nbut was: {result}\n")
+                else:
+                    print(f"No deployment that matches {deployment_name_pattern} found.\n")
+                time.sleep(interval)
+                start += interval
+        else:
+            deployment_list = self.get_deployment_names()
+            if deployment_list is not None:
+                for deployment in deployment_list:
+                    result = self.openshift.get_deployment_envFrom_info(deployment, self.namespace)
+                    if result == expected_secretRef_1 or result == expected_secretRef_2:
+                        return deployment
+                    else:
+                        print(f"\nUnexpected deployment's envFrom info: \nExpected: {expected_secretRef_1} or {expected_secretRef_2} \nbut was: {result}\n")
+            else:
+                print(f"No deployment that matches {deployment_name_pattern} found.\n")
+        return None
--- a/test/acceptance/features/steps/service_binding_request.py
+++ b/test/acceptance/features/steps/service_binding_request.py
@@ -4,7 +4,6 @@ from openshift import Openshift
 
 
 class ServiceBindingRequest(object):
-
     openshift = Openshift()
 
     def create(self, yaml):
--- a/test/acceptance/features/steps/steps.py
+++ b/test/acceptance/features/steps/steps.py
@@ -15,6 +15,10 @@ from postgres_db import PostgresDB
 from namespace import Namespace
 from nodejs_application import NodeJSApp
 from service_binding_request import ServiceBindingRequest
+from serverless_operator import ServerlessOperator
+from quarkus_application import QuarkusApplication
+from quarkus_s2i_builder_image import QuarkusS2IBuilderImage
+from knative_serving import KnativeServing
 import time
 
 
@@ -104,9 +108,8 @@ def imported_nodejs_app_is_running(context, application_name):
         application.is_running(wait=True) | should.be_truthy.desc("Application is running")
     print("Nodejs application is running!!!")
     application.get_db_name_from_api() | should_not.be_none
-    context.nodejs_app_original_generation = application.get_observed_generation()
-    context.nodejs_app_original_pod_name = application.get_running_pod_name()
-    context.nodejs_app = application
+    context.application = application
+    context.application_type = "nodejs"
 
 
 # STEP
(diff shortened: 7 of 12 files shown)


pratikjagrut:
Cluster side error.
/retest

test/acceptance/features/bindToService.feature
```diff
@@ -216,3 +216,40 @@ Feature: Bind an application to a service
...
+
+    Scenario: Bind an imported quarkus app which is deployed as knative service to PostgreSQL database
+        Given Openshift Serverless Operator is running
+        * Quarkus s2i builder image is present
+        * Knative serving is running
+        * DB "db-demo-knative" is running
+        * Quarkus application "knative-app" is imported as Knative service
+        When Service Binding Request is applied to connect the database and the application
+            """
+            ---
```
pedjak, line 228: this line could be remove, it separates multiple resources in single yaml file, but here we have only one resource.

test/acceptance/features/steps/dboperator.py
```diff
@@ -50,7 +50,7 @@ def install_operator_subscription(self):
         return True
 
     def get_package_manifest(self):
-        cmd = 'oc get packagemanifest %s -o "jsonpath={.metadata.name}"' % self.pkgManifest
+        cmd = 'oc get packagemanifest %s -o jsonpath="{.metadata.name}"' % self.pkgManifest
```
pedjak, line 53: what is the impact of this change? it looks like no-op to me?

test/acceptance/features/steps/knative_serving.py
```diff
@@ -0,0 +1,39 @@
+import re
+from command import Command
+from openshift import Openshift
+from pyshould import should_not
+
+
+class Knative_Serving():
```
pedjak, line 7: Can we name the class as `KnativeServing`, so that we follow the same pattern as with other classes introduced in this PR?

test/acceptance/features/steps/knative_serving.py
```diff
@@ -0,0 +1,39 @@
...
+        if exit_status != 0:
+            return False
+        return True
+
+    def create(self):
+        serving_output = self.openshift.oc_apply(self.knative_serving_yaml_template.format(name=self.name, namespace=self.namespace))
+        pattern = f'knativeserving.(serving|operator).knative.dev/{self.namespace}\\screated'
+        output_match = re.search(pattern, serving_output)
+        output_match | should_not.be_none.desc("Matched knative serving creation output")
+        return True
```
pedjak, line 39: is it possible to return `False` from this function?

test/acceptance/features/steps/namespace.py
```diff
@@ -9,25 +10,26 @@ def __init__(self, name):
...
-        create_namespace_output, exit_code = self.cmd.run('oc project {}'.format(self.name))
-        if re.search(r'Now using project \"%s\"\son\sserver' % self.name, create_namespace_output):
+        create_namespace_output, exit_code = self.cmd.run(f'oc project {self.name}')
+        if re.search(r'Now using project \"%s\"\son\sserver' % self.name, create_namespace_output) is not None:
             return True
-        elif re.search(r'.*Already\son\sproject\s\"%s\"\son\sserver.*' % self.name, create_namespace_output):
+        elif re.search(r'.*Already\son\sproject\s\"%s\"\son\sserver.*' % self.name, create_namespace_output) is not None:
             return True
         else:
+            False | should.be_truthy.desc(f"Unexpected namespace creating output: {create_namespace_output}")
```
pedjak, line 34: could you avoid using `should.be_truthy.` here and use approach as in line 20?

test/acceptance/features/steps/service_binding_request.py
```diff
@@ -4,9 +4,12 @@
 
 
 class ServiceBindingRequest(object):
-
+    namespace = ""
```
pedjak, line 7: is it used somewhere in this class later on?

test/acceptance/features/steps/steps.py
```diff
@@ -144,27 +149,39 @@ def db_instance_is_running(context, db_name):
...
 @then(u'application should be connected to the DB "{db_name}"')
 def then_app_is_connected_to_db(context, db_name):
-    application = context.nodejs_app
-    app_db_name = application.get_db_name_from_api()
+    application = context.application
+    if context.application_type.__contains__("knative"):
+        app_db_name = application.get_db_name_from_knative_api(timeout=180)
+    else:
+        app_db_name = application.get_db_name_from_api()
     app_db_name | should.be_equal_to(db_name)
```
pedjak, line 185: Instead of doing if/then, how about that both `NodeJSApp` and `QuarkusApplication` implement the same method, e.g. `get_db_name`? In that case, the code would become:

```python
def then_app_is_connected_to_db(context, db_name):
  app_db_name = context.application.get_db_name()
  app_db_name | should.be_equal_to(db_name)
```

pratikjagrut: @pedjak Do you think introducing an interface (say application) with common methods would help in such situations in future test scenarios?

test/acceptance/features/steps/builder_image.py
```diff
@@ -0,0 +1,52 @@
...
+        if exit_code != 0:
+            return False
+        if re.search(f'.*{self.image}\\simported', import_output) is None:
+            return False
+        spec = '{"spec": {"tags": [{"name" : \"%s\", "annotations": {"tags": "builder"}}]}}' % self.tag
+        cmd_patch = f'oc patch is {self.image} -n {self.namespace} -p \'{spec}\''
+        print(f'===> Patch CMD: {cmd_patch}')
+        (patch_output, exit_code) = self.cmd.run(cmd_patch)
+        if exit_code != 0:
+            return False
```
pratikjagrut, line 49: Will it be useful to print `patch_output` in case of failure, this might help to troubleshoot.

test/acceptance/features/steps/knative_serving.py
```diff
@@ -0,0 +1,39 @@
...
+    cmd = Command()
+
+    name = ""
+    namespace = ""
+
+    def __init__(self, name="knative-serving", namespace="knative-serving"):
+        self.name = name
+        self.namespace = namespace
+        self.knative_serving_yaml_template = '''
+---
```
pratikjagrut, line 19: I guess this separator `---` is not necessary.

test/acceptance/features/steps/knative_serving.py
```diff
@@ -0,0 +1,39 @@
...
+kind: KnativeServing
+metadata:
+  name: {name}
+  namespace: {namespace}
+'''
+
+    def is_present(self):
+        cmd = f'oc get ks {self.name} -n {self.namespace}'
+        output, exit_status = self.cmd.run(cmd)
+        if exit_status != 0:
```
pratikjagrut, line 30: Will printing output in case of failure helps?

test/acceptance/features/steps/knative_serving.py
```diff
@@ -0,0 +1,39 @@
...
+            return False
+        return True
+
+    def create(self):
+        serving_output = self.openshift.oc_apply(self.knative_serving_yaml_template.format(name=self.name, namespace=self.namespace))
+        pattern = f'knativeserving.(serving|operator).knative.dev/{self.namespace}\\screated'
+        output_match = re.search(pattern, serving_output)
+        output_match | should_not.be_none.desc("Matched knative serving creation output")
```
pratikjagrut, line 38: Will this check be sufficient for `None` and `No match` cases returned by `re.search()`?

test/acceptance/features/steps/steps.py
pedjak: not sure if an interface is really required - I mean more about Duck Typing (https://realpython.com/lessons/duck-typing/)

test/acceptance/features/bindToService.feature
```diff
@@ -216,3 +216,40 @@ Feature: Bind an application to a service
         Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding Request "binding-request-empty-app" should be changed to "True"
         And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding Request "binding-request-empty-app" should be changed to "False"
         And jq ".status.conditions[] | select(.type=="InjectionReady").reason" of Service Binding Request "binding-request-empty-app" should be changed to "EmptyApplicationSelector"
+
+    Scenario: Bind an imported quarkus app which is deployed as knative service to PostgreSQL database
```
pedjak, line 220: For an easier user/human navigation, I would extract this scenario into a separate feature file, called `bindKnativeService.feature`

test/acceptance/features/steps/steps.py
pratikjagrut: I was thinking about future test cases where we might need to add more applications and similar functions might be needed.

shvasude:
> squash commits as well.

sure

test/acceptance/features/bindToService.feature
shvasude: Agreed, but please make a note that we have to add steps under 'Background' to this feature file as well.

test/acceptance/features/bindToService.feature
shvasude: Done, Thanks

test/acceptance/features/steps/dboperator.py
shvasude: Frankly not sure i remember the reason of this change. Looks the same. I will retain the old one.

test/acceptance/features/steps/knative_serving.py
shvasude: Sure

test/acceptance/features/steps/knative_serving.py
shvasude: Can return False like in case of nodejs_application.py(install()), However, with the intent of failing fast, introduced should if pattern fails. Test has to Fail early and made not to continue if creation of knative serving is failed. Hence has to FAIL early in case of failure and hence used Should package.

test/acceptance/features/steps/namespace.py
shvasude: Any issue you see using should.be_truthy?

test/acceptance/features/steps/dboperator.py
pedjak: there are plenty of similar changes in this PR, could you also double check them?

test/acceptance/features/bindToService.feature
pedjak: yes, that is fine.

test/acceptance/features/steps/knative_serving.py
pedjak, line 38: Instead of `Matched knative serving creation output`, it would be better to display the output, for potential debugging.

test/acceptance/features/steps/knative_serving.py
pedjak: the caller of this method could check the obtained result and fail if the received value is `False`.

test/acceptance/features/steps/knative_serving.py
pratikjagrut: @shvasude There are so many other places too where failure output could be printed, could you take a look there as well.

test/acceptance/features/steps/steps.py
```diff
@@ -145,27 +151,39 @@ def db_instance_is_running(context, db_name):
...
 # STEP
+
+
 @then(u'application should be re-deployed')
 def then_application_redeployed(context):
-    application = context.nodejs_app
-    application.get_redeployed_pod_name(context.nodejs_app_original_pod_name) | should_not.be_none.desc(
-        "There is a running pod of the application different from the original one before redeployment.")
+    application = context.application
+    if context.application_type.__contains__("knative"):
```
pratikjagrut, line 171: Do you think we can make use of switch statement here? so in future, if any new application comes instead of extending the if-else ladder we can add a new case.

test/acceptance/features/steps/steps.py
```diff
@@ -145,27 +151,39 @@ def db_instance_is_running(context, db_name):
...
+            "There is a running pod of the application different from the original one before redeployment.")
 
 
 # STEP
 @then(u'application should be connected to the DB "{db_name}"')
 def then_app_is_connected_to_db(context, db_name):
-    application = context.nodejs_app
-    app_db_name = application.get_db_name_from_api()
+    application = context.application
+    if context.application_type.__contains__("knative"):
```
pratikjagrut, line 183: Same switch case use here as well.

test/acceptance/features/steps/knative_serving.py
pedjak: would error trace will be visible in log? If not, it would be better if we display `serving_output` in the log, so that one can understand more what happened, i.e. why the operation was unsuccessful.

test/acceptance/features/steps/knative_serving.py
shvasude: Done, Thanks

test/acceptance/features/bindToService.feature
```diff
@@ -217,6 +217,42 @@ Feature: Bind an application to a service
...
+                - name: JDBC_URL
+                  value: jdbc:postgresql://{{ .knav.status.dbConnectionIP }}:{{ .knav.status.dbConnectionPort }}/{{ .knav.status.dbName }}
+                - name: DB_USER
+                  value: "{{ .knav.status.dbCredentials.user }}"
+                - name: DB_PASSWORD
+                  value: "{{ .knav.status.dbCredentials.password }}"
+            """
+        Then application should be connected to the DB "db-demo-knative"
+        And jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding Request "binding-request-knative" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding Request "binding-request-knative" should be changed to "True"
```
pratikjagrut, line 255: Could it be useful to verify injected `secretRef` in `knative service`? Just to add an extra assertion.

test/acceptance/features/steps/knative_serving.py
shvasude: Returns None if there is no match. However added the output variable

test/acceptance/features/steps/knative_serving.py
shvasude: `        if exit_status != 0:
            print(f"cmd-{cmd} issued result is {output} with the exit status {exit_status}")
            return False
        return True`

shvasude: added this, let me know if this is fine

test/acceptance/features/steps/namespace.py
pedjak: you do not have consistent API.. if the method is supposed to return `false` is creation fails, then we should do that.. here we throw assertion error, and line 35 is never reached.

test/acceptance/features/steps/openshift.py
```diff
@@ -193,3 +248,44 @@ def get_resource_info_by_jq(self, resource_type, name, namespace, jq_expression,
...
         exit_code | should.be_equal_to(0).desc(f'Exit code should be 0:\n OUTPUT:\n{output}')
         return output.rstrip("\n")
+
+    def create_image_stream(self, name, registry_namespace):
+        image_stream = self.image_stream_template.format(name=name, namespace=registry_namespace)
+        return self.oc_apply(image_stream)
+
+    def get_docker_image_repository(self, name, namespace):
+        (output, exit_code) = self.cmd.run(f'oc get is {name} -n {namespace} -o jsonpath="{{.status.dockerImageRepository}}"')
+        exit_code | should.be_equal_to(0)
```
pedjak, line 258: add `.desc(output)` so that we can see what happened.

test/acceptance/features/steps/builder_image.py
shvasude: `        if exit_code != 0:
            print(f"cmd-{cmd_patch} issued result is {patch_output} with the exit status {exit_code}")
            return False`

Let me know if this fine

shvasude: Thanks

test/acceptance/features/steps/knative_serving.py
shvasude: I think yes, as `cmd.run` does other checks.

test/acceptance/features/bindToService.feature
shvasude: Need to know more on this. First thing, I believe validating injected `secretRef` is applicable for other scenarios for binding as well. like nodejs. if applicable, then i would like to do that on another PR, else if it is only applicable for binding knative scenario, let me know. Thanks

test/acceptance/features/steps/openshift.py
shvasude: sure, Thanks

shvasude: `exit_code | should.be_equal_to(0).desc(f"cmd-{cmd} result for getting docket image repository is {output} with exit code-{exit_code}")`
Let me know if this is fine

test/acceptance/features/steps/dboperator.py
shvasude: sure, Thanks

test/acceptance/features/steps/knative_serving.py
shvasude: `if re.search(pattern, serving_output):
            return True
        print(f"Pattern {pattern} did not match as creating knative serving yaml output is {serving_output}")
        return False`

Let me know if this is fine, Thanks

test/acceptance/features/steps/builder_image.py
pratikjagrut: You're already printing `{cmd_patch}` at L46 so it is not necessary here. Just output would be sufficient.

test/acceptance/features/bindToService.feature
pratikjagrut: It is applicable for all the scenarios. You can refer that 2 sbr's binding scenario for the same. Just you need to find JSON path for the knative-service scenario.

test/acceptance/features/steps/builder_image.py
shvasude: done, thanks`print(f"Result for patching image stream is {patch_output} with the exit code {exit_code}")`

test/acceptance/features/steps/service_binding_request.py
shvasude: not used. removed for now. Thanks

test/acceptance/features/steps/namespace.py
shvasude: Thanks, followed the same pattern
`return True
        else:
            print(f"Unexpected namespace creating output: '{create_namespace_output}'")
        return False`
let me know if this is fine

test/acceptance/features/steps/knative_serving.py
pratikjagrut: It looks good.

test/acceptance/features/bindToService.feature
shvasude: I would handle this in another PR as we need to have a common function for handling number of binding request across the tests (including nodejs) maybe like this: 
`Then "nodejs-app" deployment must contain SBR name "binding-request-1" and "binding-request-2" Examples: | binding_request_name | | binding-request-1 | | binding-request-2 |`

test/acceptance/features/steps/steps.py
shvasude: made to this for now. 
`application = context.application
    app_db_name = application.get_db_name_from_api()
    app_db_name | should.be_equal_to(db_name)` and renamed the function as get_db_name_from_api under QuarkusApplication
Let me know if this fine.

test/acceptance/features/steps/steps.py
shvasude: there is no switch in python, have to implement a dict function to handle that.

test/acceptance/features/steps/steps.py
shvasude: there is no switch in python, have to implement a dict function to handle that.

test/acceptance/features/bindKnativeService.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Bind an application to a knative service
+
+  As a user of Service Binding Operator
+  I want to bind applications to knative services it depends on
```
Avni-Sharma, line 4: single app or multiple apps? also if it is single app then we might want to rename the file to AppBindKnativeService.feature. Please refer https://github.com/redhat-developer/service-binding-operator/pull/598#issuecomment-669760403

test/acceptance/features/steps/builder_image.py
pratikjagrut: I don't think we need any statement to point out the output, I think it will be good if we can trim down the sentence to something like, `print(f"Exit Code: {exit_code}\nResult: {patch_output}")`.

test/acceptance/features/steps/knative_serving.py
pratikjagrut: You can resolve this conversation.

test/acceptance/features/steps/knative_serving.py
pratikjagrut: I don't think we need any statement to point out the output, I think it will be good if we can trim down the sentence to something like, `print(f"cmd: {cmd}\nExit Code: {exit_code}\nResult: {patch_output}")`.

test/acceptance/features/bindToService.feature
pratikjagrut: We already have a function to fetch `secretRef` but it is for deployment https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/steps/openshift.py#L167. 

For Knative service, you'll need to write another one and it will not make sense to raise another PR for it, this function will be very specific for knative-service only so you won't be able to use it for other scenarios.

test/acceptance/features/steps/builder_image.py
shvasude: I think, it is okay to have some statement as well. we can quickly get into this while debugging.

test/acceptance/features/steps/knative_serving.py
shvasude: I think, it is okay to have some statement as well. we can quickly get into this while debugging.
also {patch_output} is part of https://github.com/redhat-developer/service-binding-operator/pull/567#discussion_r464808372

test/acceptance/features/bindKnativeService.feature
shvasude: For not it is to bind single app, renamed to bindAppToKnativeService.feature
Let me know if this is fine.

test/acceptance/features/bindToService.feature
shvasude: I will look into it for secretRef for knative. meanwhile what do you suggest for nodejs example?

test/acceptance/features/steps/builder_image.py
pratikjagrut: Okay keep it but it should not unnecessarily prolix the log.

test/acceptance/features/steps/knative_serving.py
pratikjagrut: Okay keep it but it should not unnecessarily prolix the log.

test/acceptance/features/steps/steps.py
pratikjagrut: See if it makes sense make the changes or will keep as it is for now.

shvasude: will keep as is now

test/acceptance/features/steps/knative_serving.py
shvasude: anyways introducing logging will handle as this might go as part of debug messages

test/acceptance/features/steps/builder_image.py
shvasude: anyways introducing logging will handle as this might go as part of debug messages

test/acceptance/features/bindKnativeService.feature
pratikjagrut: @shvasude The name `bindAppToKnativeService.feature` suggest we're binding an app to knative service, but we're not doing that. We're binding an app to backend service where the app is regulated or controlled by knative-service. In this example, we're not directly making any changes to the application deployment but we're updating the knative-service which further takes care of application. 
So @Avni-Sharma IMHO `bindKnativeService.feature` is good enough even if we do bind multiple knative services they will be added under the same feature with a different scenario name.

test/acceptance/features/bindToService.feature
pratikjagrut: We can use this https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/steps/openshift.py#L167 for nodejs example.

test/acceptance/features/steps/builder_image.py
pratikjagrut: Okay, you can resolve this conversation.

test/acceptance/features/steps/knative_serving.py
pratikjagrut: Okay, you can resolve this conversation.

test/acceptance/features/steps/knative_serving.py
pratikjagrut: Okay, you can resolve this conversation.

test/acceptance/features/steps/steps.py
pratikjagrut: Okay, you can resolve this conversation.

test/acceptance/features/steps/steps.py
pratikjagrut: Okay, you can resolve this conversation.

test/acceptance/features/bindKnativeService.feature
pedjak: agree with @pratikjagrut here. but then the feature title should be updated:

```
Feature: Bind Knative service to a service

As a user of Service Binding Operator
I want to bind Knative service to services it depends on
```

test/acceptance/features/bindToService.feature
```diff
@@ -217,6 +217,7 @@ Feature: Bind an application to a service
         And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding Request "binding-request-empty-app" should be changed to "False"
         And jq ".status.conditions[] | select(.type=="InjectionReady").reason" of Service Binding Request "binding-request-empty-app" should be changed to "EmptyApplicationSelector"
 
+
```
pedjak, line 220: nit, revert whitespace change.

test/acceptance/features/steps/quarkus_application.py
```diff
@@ -0,0 +1,110 @@
...
+        build_pod_name | should_not.be_none.desc("Running Build pod name output")
+
+        status, output = self.openshift.wait_for_build_pod_status(build_pod_name, self.namespace, wait_for_status="Succeeded")
+        output | should.be_equal_to("Succeeded").desc("Build pod status should be Succeeded")
+
+        knative_service_output = self.openshift.create_knative_service(self.name, self.namespace)
+        output_match = re.search(r'.*service.serving.knative.dev/%s\s(created|configured|unchanged)' % self.name, knative_service_output)
+        output_match | should_not.be_none.desc(f"Knative serving creation output mismatch: {knative_service_output}")
+
+        return True
```
pedjak, line 46: what is the point of  return always true? we can drop that then.

test/acceptance/features/steps/service_binding_request.py
```diff
@@ -4,9 +4,11 @@
 
 
 class ServiceBindingRequest(object):
-
     openshift = Openshift()
 
+    def __init__(self, namespace):
+        self.namespace = namespace
```
pedjak, line 10: where do use/read this namespace?

test/acceptance/features/steps/steps.py
```diff
@@ -15,10 +15,16 @@
...
+from serverless_operator import ServerlessOperator
+from quarkus_application import QuarkusApplication
+from builder_image import BuilderImage
+from knative_serving import KnativeServing
 import time
 
 
 # STEP
+
+
```
pedjak, line 27: nit: revert whitespace changes.

test/acceptance/features/steps/steps.py
```diff
@@ -104,9 +110,10 @@ def imported_nodejs_app_is_running(context, application_name):
         application.is_running(wait=True) | should.be_truthy.desc("Application is running")
     print("Nodejs application is running!!!")
     application.get_db_name_from_api() | should_not.be_none
-    context.nodejs_app_original_generation = application.get_observed_generation()
-    context.nodejs_app_original_pod_name = application.get_running_pod_name()
-    context.nodejs_app = application
+    context.application_original_generation = application.get_observed_generation()
+    context.application_original_pod_name = application.get_running_pod_name()
+    context.application = application
+    context.application_type = "nodejs"
```
pedjak, line 116: why do we need to know the application type? Cannot not we make that NodeJs adn Quarkus implement the same API? I would to the later, because the number of applications will increase over the time.

test/acceptance/features/steps/steps.py
```diff
@@ -145,26 +152,35 @@ def db_instance_is_running(context, db_name):
 @when(sbr_is_applied_step)
 def sbr_is_applied(context):
     sbr_yaml = context.text
-    sbr = ServiceBindingRequest()
-    if context.__contains__("nodejs_app"):
-        application = context.nodejs_app
-        context.nodejs_app_original_generation = application.get_observed_generation()
-        context.nodejs_app_original_pod_name = application.get_running_pod_name()
+    namespace = context.namespace
+    sbr = ServiceBindingRequest(namespace.name)
```
pedjak, line 156: cannot we have simply 

```
 sbr = ServiceBindingRequest(context.namespace)

```

test/acceptance/features/steps/steps.py
```diff
@@ -145,26 +152,35 @@ def db_instance_is_running(context, db_name):
...
+    if context.__contains__("application"):
```
pedjak, line 157: how do we handle a case when this resoves to `false`? Can it happen at all?

test/acceptance/features/steps/steps.py
```diff
@@ -145,26 +152,35 @@ def db_instance_is_running(context, db_name):
...
+        application = context.application
+        context.application_original_generation = application.get_observed_generation()
+        if context.application_type.__contains__("knative"):
+            context.application_original_rev_name = application.get_rev_name()
+        else:
+            context.application_original_pod_name = application.get_running_pod_name()
```
pedjak, line 163: do we really need these here? what do we achieve with it since we overwrite the values for example in `imported_nodejs_app_is_running(context, application_name` method?

test/acceptance/features/steps/steps.py
```diff
@@ -145,26 +152,35 @@ def db_instance_is_running(context, db_name):
...
     sbr.create(sbr_yaml) | should.be_truthy.desc("Service Binding Request Created")
 
-
 # STEP
+
+
```
pedjak, line 168: nit: remove whitespace changes.

test/acceptance/features/steps/steps.py
```diff
@@ -145,26 +152,35 @@ def db_instance_is_running(context, db_name):
...
-    application = context.nodejs_app
-    application.get_redeployed_pod_name(context.nodejs_app_original_pod_name) | should_not.be_none.desc(
-        "There is a running pod of the application different from the original one before redeployment.")
+    application = context.application
+    if context.application_type.__contains__("knative"):
+        application.get_redeployed_rev_name(context.application_original_rev_name) | should_not.be_none.desc(
+            "There is a running revision of the application different from the original one before redeployment")
+    else:
+        application.get_redeployed_pod_name(context.application_original_pod_name) | should_not.be_none.desc(
+            "There is a running pod of the application different from the original one before redeployment.")
```
pedjak, line 177: we should really avoid this if/then/switch statements... instead create a common api... if we want to know if an application  could redeployed, then we could have a method e.g. `is_redeployed()` which would return `true/false`. `context.application_original_rev_name` or `context.application_original_pod_name` already known internal application properties, so they do not need to preserved explicitly in context.

test/acceptance/features/steps/steps.py
```diff
@@ -190,6 +206,78 @@ def then_sbo_jq_is(context, jq_expression, sbr_name, json_value_regex):
 
 
 # STEP
+is_serverless_operator_running_step = u'Openshift Serverless Operator is running'
```
pedjak, line 209: do we really need this extra var, or we could apply the approach similar to others steps:

```

@given(u'Openshift Serverless Operator is running')
def given_serverless_operator_is_running(context)
```

test/acceptance/features/steps/steps.py
```diff
@@ -190,6 +206,78 @@ def then_sbo_jq_is(context, jq_expression, sbr_name, json_value_regex):
...
+        print("Serverless operator is not installed, installing...")
+        serverless_operator.install_operator_subscription() | should.be_truthy.desc(
+            "serverless operator subscription installed")
+        serverless_operator.is_running(wait=True) | should.be_truthy.desc("serverless operator installed")
+    print("Serverless operator is running!!!")
+    context.serverless_operator = serverless_operator
+
+
+# STEP
+is_quarkus_s2i_image_present_step = u'Quarkus s2i builder image is present'
```
pedjak, line 228: see the comment above.

test/acceptance/features/steps/steps.py
```diff
@@ -190,6 +206,78 @@ def then_sbo_jq_is(context, jq_expression, sbr_name, json_value_regex):
...
+
+
+@given(is_quarkus_s2i_image_present_step)
+def given_quarkus_builder_image_is_present(context):
+    """
+    Checks if quarkus s2i builder image is present
+    """
+    builder_image = BuilderImage()
+    if not builder_image.is_present():
+        builder_image.import_and_patch() | should.be_truthy.desc("Quarkus image is imported from image registry and patched to builder")
```
pedjak, line 238: it looks to me that the assertion error messages are not correct, for example here: the error gets triggered if `import_and_patch` returns false, but then the message should be "Quarkus image is iNOT mported from image registry and patched to builder"

Same goes for the next message. Please double check other messages in this PR.

test/acceptance/features/steps/steps.py
```diff
@@ -190,6 +206,78 @@ def then_sbo_jq_is(context, jq_expression, sbr_name, json_value_regex):
...
+        builder_image.is_present() | should.be_truthy.desc("Quarkus image is present")
+    print("Quarkus s2i builder image is present!!!")
```
pedjak, line 240: that is expected, right? In that case, why do we print this message and why with so many exclamation marks? Should this be alarming for somebody reading the logs?

test/acceptance/features/steps/steps.py
```diff
@@ -190,6 +206,78 @@ def then_sbo_jq_is(context, jq_expression, sbr_name, json_value_regex):
...
+
+
+# STEP
+is_knative_serving_running_step = u'Knative serving is running'
+
+
+@given(is_knative_serving_running_step)
```
pedjak, line 247: can we define the annotation as in another steps:

```

@given(u'Knative serving is running')
```

test/acceptance/features/steps/steps.py
```diff
@@ -190,6 +206,78 @@ def then_sbo_jq_is(context, jq_expression, sbr_name, json_value_regex):
...
+def given_knative_serving_is_running(context):
+    """
+    creates a KnativeServing object to install Knative Serving using the OpenShift Serverless Operator.
+    """
+    knative_namespace = Namespace("knative-serving")
+    knative_namespace.create() | should.be_truthy.desc("knative serving namespace is created")
```
pedjak, line 253: again I think that the assertion message is wrong, it should be the negative form.

test/acceptance/features/steps/steps.py
```diff
@@ -190,6 +206,78 @@ def then_sbo_jq_is(context, jq_expression, sbr_name, json_value_regex):
...
+    Namespace(context.namespace.name).switch_to() | should.be_truthy.desc("Switch back to the context namespace")
+    knative_serving = KnativeServing(namespace=knative_namespace.name)
+    if not knative_serving.is_present():
+        print("knative serving is not present, create knative serving")
+        knative_serving.create() | should.be_truthy.desc("knative serving is created")
+        knative_serving.is_present() | should.be_truthy.desc("knative serving is present")
+    print("knative serving is present!!!")
```
pedjak, line 260: see above comments for this.

test/acceptance/features/steps/steps.py
```diff
@@ -190,6 +206,78 @@ def then_sbo_jq_is(context, jq_expression, sbr_name, json_value_regex):
...
+    namespace = context.namespace
+    application = QuarkusApplication(application_name, namespace.name)
+    if not application.is_imported():
+        print("application is not imported, trying to import it")
+        application.install() | should.be_truthy.desc("Quarkus application is installed")
+        application.is_imported() | should.be_truthy.desc("Quarkus application is imported")
+    context.application_original_generation = application.get_observed_generation()
+    context.application = application
+    context.application_type = "knative"
+
```
pedjak, line 279: see above comments for this, they apply here as well.

test/acceptance/features/bindKnativeService.feature
shvasude: Done, feature file name: bindKnativeService.feature
and 
Feature: Bind knative service to a service
  As a user of Service Binding Operator
  I want to bind knative service to services it depends on

shvasude: or should we say ?
As a user of Service Binding Operator
I want to bind app running as knative service to services it depends on

test/acceptance/features/bindToService.feature
shvasude: fixed, thanks

test/acceptance/features/steps/quarkus_application.py
shvasude: As `quarkus_app_is_imported_as_knative_service` function from steps.py expects a value for assertion. We have followed this in nodejs as well.

test/acceptance/features/steps/steps.py
shvasude: Fixed, thanks

test/acceptance/features/steps/steps.py
shvasude: I feel it is easy handle with type, moreover there are rev as a resource that we check for knative, otherwise for others it checks pods.

test/acceptance/features/steps/steps.py
shvasude: ` sbr = ServiceBindingRequest(context.namespace)` will not fetch the name of the namespace. i think it gets namespace object. maybe we should do ` sbr = ServiceBindingRequest(context.namespace.name)`. segregated for better readability

test/acceptance/features/steps/steps.py
shvasude: Done, thanks

test/acceptance/features/steps/steps.py
shvasude: We could apply the same for other steps as well. As this var that holds the description of the step can be in `given` or any other keywords like `then`

test/acceptance/features/steps/steps.py
shvasude: We could apply the same for other steps as well. As this var that holds the description of the step can be in given or any other keywords like then

test/acceptance/features/bindKnativeService.feature
pratikjagrut: @pedjak Shouldn't the feature title be like:
```
Feature: Bind Knative service to a backend service


Feature: Bind Knative service to a database
```
Shouldn't we be articulate about which service is binding to which service?

test/acceptance/features/bindToService.feature
shvasude: Sure, added this block 
`@ then(u'"{app_name}" deployment must contain SBR name "{sbr_name}"')
def then_envFrom_contains_SBR_name(context, app_name, sbr_name):
    time.sleep(60)
    openshift = Openshift()
    result = openshift.get_deployment_envFrom_info(app_name, context.namespace.name)
    assert re.search(r'.*[map[secretRef:map[name:%s]]' % app_name, result) is not None, f'{app_name} deployment does not contain secretRef: {sbr_name}'
`

test/acceptance/features/bindKnativeService.feature
shvasude: As far as i remember from an initial discussions is that the service can be any service within a feature file that will be part of different scenario. for now, i am sticking to this
https://github.com/redhat-developer/service-binding-operator/pull/567#discussion_r467027706

test/acceptance/features/steps/service_binding_request.py
shvasude: removed this, thanks

test/acceptance/features/steps/steps.py
shvasude: removed the application type, and handling this by capturing get_generation and the same will be validated in is_redeployed()
thanks

test/acceptance/features/steps/steps.py
shvasude: this is a condition where if application is involved then only captures the generation number and for all other yaml to be applied it skips tihs `if` condition and proceeds.

test/acceptance/features/steps/steps.py
shvasude: I feel it should be here for capturing generation/observed generation, reason being as we check if application is redeployed, we would need to check these generation right before sbr is applied.

test/acceptance/features/steps/steps.py
shvasude: sure, this has changed to this, please confirm. Now there are no checks for application type as knative or nodejs
`@then(u'application should be re-deployed')
def then_application_redeployed(context):
    application = context.application
    assert application.is_redeployed(context.application_original_generation) is not None, "application is not redeployed"`

test/acceptance/features/bindToService.feature
pratikjagrut: Could you please check the JSON path for injected secretRef in knative service. If the JSON path is same is deployment then `openshift.get_deployment_envFrom_info()` is useful or you can use a robust method to retrieve elements using JSON path: https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/steps/openshift.py#L173

test/acceptance/features/steps/steps.py
shvasude: I have replaced should with assert with corresponding error message.

test/acceptance/features/steps/steps.py
shvasude: removed.thanks

test/acceptance/features/steps/steps.py
shvasude: if it has only a specific step to run then applied in the annotation itself, else if the same step for given and then, in that case added variable.

test/acceptance/features/steps/steps.py
shvasude: I have replaced should with assert with corresponding error message.

test/acceptance/features/steps/steps.py
shvasude: I have replaced should with assert with corresponding error message.

test/acceptance/features/steps/steps.py
shvasude: I have replaced should with assert with corresponding error message.

test/acceptance/features/bindToService.feature
shvasude: it is same for knative as well,
`[shobith@localhost util]$ oc get deploy knative-app-44k86-deployment -n test-namespace-c5501dbe -o "jsonpath={.spec.template.spec.containers[0].envFrom}"
[map[secretRef:map[name:binding-request-knative]]]`
Hence using the same jsonpath path for secretRef

test/acceptance/features/steps/openshift.py
```diff
@@ -148,15 +194,24 @@ def expose_service_route(self, service_name, namespace):
         return re.search(r'.*%s\sexposed' % service_name, output)
 
     def get_route_host(self, name, namespace):
-        (output, exit_code) = self.cmd.run(f'oc get route {name} -n {namespace} -o "jsonpath={{.status.ingress[0].host}}"')
-        exit_code | should.be_equal_to(0)
+        output, exit_code = self.cmd.run(f'oc get route {name} -n {namespace} -o "jsonpath={{.status.ingress[0].host}}"')
+        exit_code | should.be_equal_to(0).desc(f"Getting route host exit code should be 0\nERROR: {output}")
```
pedjak, line 198: the discussion out come was to replace `should` calls with `assert`

test/acceptance/features/steps/openshift.py
```diff
@@ -148,15 +194,24 @@ def expose_service_route(self, service_name, namespace):
...
-        return deployment_status
+    def get_deployment_status(self, deployment_name, namespace, wait_for_status=None):
+        deployment_status_cmd = f'oc get deployment {deployment_name} -n {namespace} -o json' \
+            + ' | jq -rc \'.status.conditions[] | select(.type=="Available").status\''
+        output = None
+        exit_code = -1
+        if wait_for_status is not None:
+            status_found, output, exit_code = self.cmd.run_wait_for_status(deployment_status_cmd, wait_for_status, 5, 300)
+            if exit_code == 0:
+                status_found | should.be_truthy.desc(f"Deployment {deployment_name} status should be {wait_for_status},\n ERROR: {output}")
```
pedjak, line 209: see above

test/acceptance/features/steps/openshift.py
```diff
@@ -148,15 +194,24 @@ def expose_service_route(self, service_name, namespace):
...
+        else:
+            output, exit_code = self.cmd.run(deployment_status_cmd)
+        exit_code | should.be_equal_to(0).desc(f"Getting deployment status exit code should be 0\nERROR: {output}")
```
pedjak, line 212: see above

test/acceptance/features/steps/service_binding_request.py
```diff
@@ -4,7 +4,6 @@
 
 
 class ServiceBindingRequest(object):
-
```
pedjak, line 7: please revert changes on files containing only whitespace changes.

test/acceptance/features/steps/steps.py
```diff
@@ -118,7 +120,7 @@ def imported_nodejs_app_is_running(context, application_name):
 def imported_nodejs_app_is_not_running(context, application_name):
     namespace = context.namespace
     application = NodeJSApp(application_name, namespace.name)
-    application.is_running() | should.be_falsy.desc("Aplication not running")
+    application.is_running() | should.be_falsy.desc("Application not running")
```
pedjak, line 123: replace with assert

test/acceptance/features/steps/openshift.py
shvasude: I have replaced should with assert with corresponding error message.

test/acceptance/features/steps/openshift.py
shvasude: I have replaced should with assert with corresponding error message.

test/acceptance/features/steps/openshift.py
shvasude: I have replaced should with assert with corresponding error message.

test/acceptance/features/steps/service_binding_request.py
shvasude: done

test/acceptance/features/steps/steps.py
shvasude: I have replaced should with assert with corresponding error message.

test/acceptance/features/steps/steps.py
```diff
@@ -189,11 +188,72 @@ def then_sbo_jq_is(context, jq_expression, sbr_name, json_value_regex):
...
 @then(u'"{app_name}" deployment must contain SBR name "{sbr_name1}" and "{sbr_name2}"')
 def then_envFrom_contains(context, app_name, sbr_name1, sbr_name2):
     time.sleep(60)
     openshift = Openshift()
     result = openshift.get_deployment_envFrom_info(app_name, context.namespace.name)
     result | should.be_equal_to("[map[secretRef:map[name:binding-request-1]] map[secretRef:map[name:binding-request-2]]]")\
         .desc(f'{app_name} deployment should contain secretRef: {sbr_name1} and {sbr_name2}')
+
+
+@then(u'deployment must contain SBR name "{sbr_name}"')
```
DhritiShikhar, line 256: Well, actually it's intermediate secret name and not SBR name. 
The intermediate secret name created by SBO is same as SBR name.

But this should be corrected. :)

test/acceptance/features/bindKnativeService.feature
```diff
@@ -0,0 +1,46 @@
...
+                  value: jdbc:postgresql://{{ .knav.status.dbConnectionIP }}:{{ .knav.status.dbConnectionPort }}/{{ .knav.status.dbName }}
+                - name: DB_USER
+                  value: "{{ .knav.status.dbCredentials.user }}"
+                - name: DB_PASSWORD
+                  value: "{{ .knav.status.dbCredentials.password }}"
+      """
+    Then deployment must contain SBR name "binding-request-knative"
+    Then application should be connected to the DB "db-demo-knative"
+    And jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding Request "binding-request-knative" should be changed to "True"
+    And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding Request "binding-request-knative" should be changed to "True"
```
DhritiShikhar, line 46: There should be a check that the intermediate secret contains all these customEnvVar?

test/acceptance/features/steps/steps.py
shvasude: done

test/acceptance/features/bindKnativeService.feature
```diff
@@ -0,0 +1,46 @@
...
+            kind: ServiceBindingRequest
+            metadata:
+              name: binding-request-knative
+            spec:
+              applicationSelector:
+                group: serving.knative.dev
+                version: v1beta1
+                resource: services
+                resourceRef: knative-app
+              backingServiceSelector:
```
DhritiShikhar, line 29: backingServiceSelector is going to be deprecated. 

We should use backingServiceSelectors

Avni-Sharma: I think the API will change collectively from backingserviceSelector/selectors to services. I don't think we would be deprecating backingServiceSelector separately.

shvasude: I tried with backingServiceSelectors and sbr fails

PR #568, opened 2020-07-17
Add roles for crds and routes

Avni-Sharma:
Update deploy/role.yaml for CRDs and routes.


sbose78:
Do we want to follow-up with your ci tests and ensure this path is validated
? @pmacik @Avni-Sharma

Issue #569, opened 2020-07-18
Incorrect instruction steps in nodejs_etcd_operator example


noelo:
Missing labelling of deployment
Incorrect ServiceBindingRequest, missing  labelSelector:
Incorrect etcd cluster name in SBR

Avni-Sharma:
Examples have been updated. Closing this issue

PR #571, opened 2020-07-18
Fix for #569, Incorrect instruction steps in nodejs_etcd_operator example

noelo:
### Motivation

Unable to run the nodejs_etcd_operator example as instructions were incomplete

### Changes

Fixed up README.md with correct instructions

### Testing

Tested on OCP 4.5

### DIFF
--- a/examples/nodejs_etcd_operator/README.md
+++ b/examples/nodejs_etcd_operator/README.md
@@ -1,6 +1,6 @@
 ## Binding an Imported app to an In-cluster Operator Managed Etcd Database.
 
-1. Install Etcd operator using operator hub, 
+1. Install Etcd operator using operator hub,
    follow https://operatorhub.io/operator/etcd
 2. Create an Etcd cluster.
  ```yaml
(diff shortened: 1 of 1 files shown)


examples/nodejs_etcd_operator/README.md
```diff
@@ -16,31 +16,46 @@
...
-  name: binding-request
+  name: node-binding-request
 spec:
   applicationSelector:
-    matchLabels:
-      connects-to: etcd
-      environment: demo
-    group: apps.openshift.io
+    labelSelector:
+      matchLabels:
```
Avni-Sharma, line 42: Hi @noelo . Thanks for the PR. I believe removing the matchlabels from the ServiceBindingRequest,SBR, completely would be better since we are adding it in step 4.

examples/nodejs_etcd_operator/README.md
```diff
@@ -16,31 +16,46 @@
...
+oc label deployment node-todo-git 'connects-to=etcd' 'environment=demo'
+```
+
+5. Create SBR.
 ```yaml
 kind: ServiceBindingRequest
 metadata:
-  name: binding-request
+  name: node-binding-request
```
Avni-Sharma, line 38: IMO, This can stay as `binding-request`

examples/nodejs_etcd_operator/README.md
noelo: Step 4 is adding the labels to the deployment. I just tried the SBR without the matchLabels and it does work. Do you want that removed ?

DhritiShikhar: @Avni-Sharma On step 4, we are adding labels to the deployment.
Here, the labelSelector is mentioned in SBR.
Both are needed.

Avni-Sharma: Yeah, either or works but I don't think both are needed- labels and resourceRef. It is more convenient to just have a resourceRef. We can have other examples where we are actually using multiple apps via labels. But having both is indeed redundant

Avni-Sharma:
Taking this forward in https://github.com/redhat-developer/service-binding-operator/pull/599 for a speedy merge :rocket:

PR #573, opened 2020-07-20
[acceptance-tests] Replace deprecated OperatorSource by CatalogSource.

pmacik:
### Motivation

`OperatorSource` is [deprecated](https://docs.openshift.com/container-platform/4.4/release_notes/ocp-4-4-release-notes.html#ocp-4-4-marketplace-apis-deprecated) for OpenShift 4.4+

Fixes [APPSVC-592](https://issues.redhat.com/browse/APPSVC-592) for the acceptance tests.

### Changes

Use of `OperatorSource` to install PostgreSQL DB operator is replaced by the use of plain `CatalogSource`.

### Testing

`make test-acceptance`

### DIFF
--- a/test/acceptance/features/steps/dboperator.py
+++ b/test/acceptance/features/steps/dboperator.py
@@ -14,8 +14,8 @@ class DbOperator():
 
     name = ""
     namespace = ""
-    operator_source_name = "db-operators"
-    operator_registry_namespace = "pmacik"
+    operator_source_name = "pmacik-olm-registry"
+    operator_registry_image = "quay.io/pmacik/olm:v1"
     operator_registry_channel = "stable"
     package_name = "db-operators"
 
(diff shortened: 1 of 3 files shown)


pmacik:
Depends on https://github.com/redhat-developer/service-binding-operator/pull/574

test/acceptance/features/steps/dboperator.py
```diff
@@ -14,8 +14,8 @@ class DbOperator():
 
     name = ""
     namespace = ""
-    operator_source_name = "db-operators"
-    operator_registry_namespace = "pmacik"
+    operator_source_name = "pmacik-olm-registry"
+    operator_registry_image = "quay.io/pmacik/olm:v1"
```
Avni-Sharma, line 18: Should we consider having the image at `quay.io/redhat-developer/sbo-olm:v1` ?

test/acceptance/features/steps/openshift.py
```diff
@@ -9,17 +9,19 @@
...
+kind: CatalogSource
 metadata:
     name: {name}
     namespace: openshift-marketplace
 spec:
-    type: appregistry
-    endpoint: https://quay.io/cnr
-    registryNamespace: {registry_namespace}
+    sourceType: grpc
+    image: {registry_image}
```
Avni-Sharma, line 20: Just curious, Now that operator source is deprecated, this registry is not the quay app registry ? It can be simply an image of the bundle?

test/acceptance/features/steps/dboperator.py
pmacik: > Should we consider having the image at `quay.io/redhat-developer/sbo-olm:v1` ?

There are 2 app registries actually in quay (the `redhat-developer` where sbo is pushed, and `pmacik` where sample db operators are pushed) at this moment.

So `quay.io/redhat-developer/sbo-olm:v1` would be better for SBO itself and so the DB operators registry image would need a different name since it comes from different app registry.

So, what about creating the two image registries, one for SBO and other for the sample DB operators (`quay.io/redhat-developer/sbo-olm:v1` and `quay.io/redhat-developer/sample-db-operators-olm:v1`) and push the db operators registry image to the `sample-db-operators-olm` instead?

test/acceptance/features/steps/openshift.py
pmacik: Actually it is a "operator catalog" rather then "app registry". According to the [official docs](https://docs.openshift.com/container-platform/4.4/release_notes/ocp-4-4-release-notes.html#ocp-4-4-marketplace-apis-deprecated) we need to "build the operator catalog image from the quay app registry". So, I see that the name of the variable `{registry_image}` is a bit misleading - so I'll rename it to the `catalog_image`.

It can't be simply an image of the bundle, it is supposed to be an operator catalog to be consumed by the openshift marketplace

test/acceptance/features/steps/dboperator.py
Avni-Sharma: yeah +1 for sample-db-operators-olm

Avni-Sharma:
Hi @pmacik Should we consider OperatorSource changes in examples/ as well?
In accordance with Acceptance criteria in https://issues.redhat.com/browse/APPSVC-592

pmacik:
> Hi @pmacik Should we consider OperatorSource changes in examples/ as well?
> In accordance with Acceptance criteria in https://issues.redhat.com/browse/APPSVC-592

@Avni-Sharma this PR is meant just  for the acceptnace tests, that is why I didn't include it.

PR #583, opened 2020-07-23
Reconcile should log error if a resource is not found in detect Bindable resource

Avni-Sharma:
Fix for #584 
Do not error out when resource not found in the binding resources, Instead log it as stated below
```
2020-07-23T21:04:33.309+0530	DEBUG	reconciler.buildServiceContexts	Resource not found in Bindable Resources	{"Request.Namespace": "try", "Request.Name": "binding-request-etcd", "ServiceBindingRequest.Name": "binding-request-etcd", "Error": "the server could not find the requested resource"}
```

### DIFF
--- a/pkg/controller/servicebindingrequest/service.go
+++ b/pkg/controller/servicebindingrequest/service.go
@@ -173,13 +173,12 @@ func getOwnedResources(
 	error,
 ) {
 	var resources []*unstructured.Unstructured
-BINDING_RESOURCES:
 	for _, br := range bindableResources {
 		lst, err := client.Resource(br.gvr).Namespace(ns).List(metav1.ListOptions{})
 		if err != nil {
 			if k8serrors.IsNotFound(err) {
 				logger.Debug("Resource not found in Bindable Resources", "Error", err)
-				continue BINDING_RESOURCES
+				continue
 			}
 			return resources, err
 		}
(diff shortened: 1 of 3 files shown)


pkg/controller/servicebindingrequest/service.go
```diff
@@ -170,9 +173,14 @@ func getOwnedResources(
 	error,
 ) {
 	var resources []*unstructured.Unstructured
+BINDING_RESOURCES:
```
sbose78, line 176: This is stray text 🤔

Avni-Sharma: this is a label used in L182 with continue for better clarity

dharmit:
Works for me on minikube. I'm able to create a deployment (`odo push`) and link it with a running service (`odo link`).

pkg/controller/servicebindingrequest/service.go
baijum: I think `continue` without the label is not confusing.

DhritiShikhar: @baijum Looks okay to me. Are you okay with this change?

baijum: @DhritiShikhar I think the label can be removed.

Avni-Sharma: was trying to stay consistent with https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/controller/servicebindingrequest/servicecontext.go#L62

PR #588, opened 2020-07-28
[acceptance-tests] Enable test artifacts gathering in CI

pmacik:
### Motivation

Currently the only archived data is the test console output archived in CI. To be able to investigate any acceptance tests failures it is useful and necessary to gather as much data as possible and archive it along the prow job that executed it.

### Changes

This PR:
* Moves the local SBO log under `out/acceptance-tests` directory
* Adds `test-acceptance-artifacts` target (so it can be called in the CI as soon as the tests are finished) to copy the whole `out/acceptnace-tests` directory to the `/tmp/artifacts` (which is the directory from CI ultimately gathers and archives the artifacts)

This is a dependency for https://github.com/openshift/release/pull/10489

### Testing

`make test-acceptance-artifacts`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,6 +159,8 @@ PULL_NUMBER := $(shell echo $$CLONEREFS_OPTIONS | jq '.refs[0].pulls[0].number')
 
 # -- Variables for acceptance tests
 TEST_ACCEPTANCE_START_SBO ?= local
+TEST_ACCEPTANCE_OUTPUT_DIR ?= $(OUTPUT_DIR)/acceptance-tests
+TEST_ACCEPTANCE_ARTIFACTS ?= /tmp/artifacts
 
 ## -- Static code analysis (lint) targets --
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -159,6 +159,8 @@ PULL_NUMBER := $(shell echo $$CLONEREFS_OPTIONS | jq '.refs[0].pulls[0].number')
 
 # -- Variables for acceptance tests
 TEST_ACCEPTANCE_START_SBO ?= local
+TEST_ACCEPTANCE_OUTPUT_DIR ?= $(OUTPUT_DIR)/acceptance-tests
+TEST_ACCEPTANCE_ARTIFACTS ?= /tmp/artifacts
```
baijum, line 163: Is global `/tmp` a good idea? Why not under `OUTPUT_DIR`?

pmacik: `/tmp/artifacts` is the target where the `test-acceptance-artifacts` are copied/stored. It is used by the OpenShift CI jobs to collect and archive the artifacts from.

But you are right, it might be better to default it to the output dir and set it to the `/tmp/artifacts` for the CI job via parameter.

baijum: If this is the current configured location, please go ahead.

PR #596, opened 2020-08-03
Deprecate relationship annotations

### LINKED ISSUE #458: Only the last SBR is recorded in backingservice annotation.
Currently,  we have a requirement to  support the multiple application binding testcase: 
 binding application `A` -> backingService CR `Y`  with SBR CR named as A-Y
 binding application `B` -> backingService CR  `Y` as well  with SBR CR named as B-Y

After the binding operation, I would like to do some further queries to understand how many application is bound to the service.   
But when checking the Y's annotation,  I only see: 
```
service-binding-operator.apps.openshift.io/binding-name:  B-Y
```
Also, once I deleted SBR `B-Y`, there is no annotation at all on the backingService CR `Y` any more ! 

I know there may be some discussion to define multiple applications inside a single SBR,  but if the end-user intends to define multiple SBRs over the same backingService (as my example above),  I think the annotation should reflect a correct information. 

So, I hope to  need to change this field from json entry to json array as:
```
service-binding-operator.apps.openshift.io/binding-name:  A-Y,B-Y
```
or 
```
     service-binding-operator.apps.openshift.io/binding-name: |
      [
       {"binding-name":"sbr1"},
       {"binding-name":"sbr2"},
      ]
```

Any comment? Any other suggestion?

isutton:
This PR deprecates the usage of annotations to relate Service Binding Requests and its services and applications in favor of a programmatic lookup.

Closes #458 

### Motivation

The main reason for this change is to reduce the amount of writes in the API server during the SBR reconciliation: those writes can trigger other reconciliation processes leading to unexpected race conditions; additionally, those annotations also seem to not be necessary to find a direct relationship between a Service Binding Request and a related resource, being it a service, application or the intermediate secret as changes in this PR can show.

### Changes

This PR does not stamp the SBR information to related resources using annotations during the reconciliation process, and performs a lookup up to identify the SBR associated with an incoming change, if any.

### Testing

Unit tests have been modified to a test table, and no e2e tests have been changed.

### DIFF
--- a/pkg/controller/servicebindingrequest/mapper.go
+++ b/pkg/controller/servicebindingrequest/mapper.go
@@ -1,9 +1,18 @@
 package servicebindingrequest
 
 import (
-	"github.com/redhat-developer/service-binding-operator/pkg/log"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/redhat-developer/service-binding-operator/pkg/apis/apps/v1alpha1"
+	"github.com/redhat-developer/service-binding-operator/pkg/log"
 )
 
 var (
@@ -17,6 +26,93 @@ type sbrRequestMapper struct {
 	restMapper meta.RESTMapper
 }
 
+var serviceBindingRequestGVK = v1alpha1.SchemeGroupVersion.WithKind("ServiceBindingRequest")
+var secretGVK = corev1.SchemeGroupVersion.WithKind("Secret")
+
+// isServiceBindingRequest checks whether the given obj is a Service Binding Request through GVK
+// comparison.
+func isServiceBindingRequest(obj runtime.Object) bool {
+	return obj.GetObjectKind().GroupVersionKind() == serviceBindingRequestGVK
+}
+
+// isSecret checks whether the given obj is a Secret through GVK comparison.
+func isSecret(obj runtime.Object) bool {
+	return obj.GetObjectKind().GroupVersionKind() == secretGVK
+}
+
+// isSBRService checks whether the given obj is a service in given sbr.
+func isSBRService(sbr *v1alpha1.ServiceBindingRequest, obj runtime.Object) bool {
+	services := extractServiceSelectors(sbr)
+	for _, svc := range services {
+		svcGVK := schema.GroupVersionKind{Group: svc.Group, Version: svc.Version, Kind: svc.Kind}
+		if obj.GetObjectKind().GroupVersionKind() == svcGVK {
+			return true
+		}
+	}
+	return false
+}
+
+// isApplication checks whether the given obj is an application in given sbr.
+func isApplication(
+	restMapper meta.RESTMapper,
+	sbr *v1alpha1.ServiceBindingRequest,
+	obj runtime.Object,
+) (bool, error) {
+	appGVR := schema.GroupVersionResource{
+		Group:    sbr.Spec.ApplicationSelector.Group,
+		Version:  sbr.Spec.ApplicationSelector.Version,
+		Resource: sbr.Spec.ApplicationSelector.Resource,
+	}
+	appGVK, err := restMapper.KindFor(appGVR)
+	if err != nil {
+		return false, err
+	}
+
+	isEqual := obj.GetObjectKind().GroupVersionKind() == appGVK
+
+	return isEqual, nil
+}
+
+// isSecretOwnedBySBR checks whether the given obj is a secret owned by the given sbr.
+func isSecretOwnedBySBR(obj metav1.Object, sbr *v1alpha1.ServiceBindingRequest) bool {
+	return sbr.GetNamespace() == obj.GetNamespace() && sbr.Status.Secret == obj.GetName()
+}
+
+// convertToSBR attempts to convert the given obj into a Service Binding Request.
+func convertToSBR(obj map[string]interface{}) (*v1alpha1.ServiceBindingRequest, error) {
+	sbr := &v1alpha1.ServiceBindingRequest{}
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj, sbr)
+	return sbr, err
+}
+
+// convertToNamespacedName returns a NamespacedName with information extracted from given obj.
+func convertToNamespacedName(obj metav1.Object) types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	}
+}
+
+// namespacedNameSet is a set of NamespacedNames.
+type namespacedNameSet map[types.NamespacedName]bool
+
+// add adds the given namespaced name n into the set.
+func (t namespacedNameSet) add(n types.NamespacedName) {
+	t[n] = true
+}
+
+func convertToRequests(t namespacedNameSet) []reconcile.Request {
+	toReconcile := make([]reconcile.Request, 0)
+	for n := range t {
+		toReconcile = append(
+			toReconcile,
+			reconcile.Request{NamespacedName: n},
+		)
+	}
+	return toReconcile
+
+}
+
 // Map execute the mapping of a resource with the requests it would produce. Here we inspect the
 // given object trying to identify if this object is part of a SBR, or a actual SBR resource.
 func (m *sbrRequestMapper) Map(obj handler.MapObject) []reconcile.Request {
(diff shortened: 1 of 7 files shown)


pkg/controller/servicebindingrequest/mapper.go
```diff
@@ -21,17 +120,60 @@ func (m *sbrRequestMapper) Map(obj handler.MapObject) []reconcile.Request {
...
+
+	sbrList, err := m.client.Resource(groupVersion).List(metav1.ListOptions{})
 	if err != nil {
-		log.Error(err, "on inspecting object for annotations for SBR object")
-		return toReconcile
+		log.Error(err, "listing SBRs")
+		// NOTE: nothing to do anymore if obj isn't a SBR and there's an error listing SBRs; I'm not
+		// sure whether retrying in the same reconcile loop is the proper approach so we'll rely on
+		// the resync period to re-process all watched objects.
+		return []reconcile.Request{}
```
Avni-Sharma, line 140: Perhaps we can use this function https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/controller/servicebindingrequest/common.go#L45

isutton: I believe you're referring to the following function:

https://github.com/redhat-developer/service-binding-operator/blob/ffbf35ff6f419d09795389a820fee61954abfdc6/pkg/controller/servicebindingrequest/common.go#L45-L50

The mapper might return reconciliation requests which will be queued to be processed by a worker; in othe words, the `Reconcile()` method is only executed after the mapping happens and produces some result.

Now that I'm thinking, that comment is misleading since it is fine if a resource is examined and the mapper results in an empty slice. 

Take the following use cases:

* Add a Service Binding Request in a namespace without declaring the `applicationSelector` field.
* Add a Deployment (the mapper will not find a relationship between the given resource and a SBR, so no SBRs will be further enqueued).
* Modify the Service Binding Request and modify the `applicationSelector` to include the newly added Deployment (this will trigger the SBR's reconciliation process).

The other use case would be create a Service Binding Request declaring `applicationSelector` with a missing application:

* Add a Service Binding Request in a namespace declaring the `applicationSelector` field, but no resource as of yet.
* Add a Deployment (the relationship will be identified by the mapper and the related SBR is enqueued).

Given this I think the behavior is fine, what do you think if I just amend the comment?

isutton: Inspecting further the [`List` method][1], there are a couple of error types being returned, although from a developer's perspective I believe this section should be resilient mostly to `*errors.StatusError`, at least initially. 

Some sub-errors could be retried in the mapping procedure, such as `TooManyRequests` and `Timeout`, where others that are not recoverable such as `Forbidden` or `NotFound`, so it is likely this block should take this in account.



[1]: https://github.com/isutton/service-binding-operator/blob/ba129432e43bf26455edeebf88bcaf92184586c3/vendor/k8s.io/client-go/dynamic/simple.go#L260-L282

isutton: I've learned the client is [programmed to retry][1] in specific situations the API server instructs it to; this makes me believe all errors at this point are not recoverable. I'll modify the comment to make that clear.

[1]: https://github.com/isutton/service-binding-operator/blob/e17445570bd3889bcf7499142350a3b81463c6be/vendor/k8s.io/client-go/rest/request.go#L723-L812

pkg/controller/servicebindingrequest/mapper.go
```diff
@@ -12,7 +21,97 @@ var (
...
+var secretGVK = corev1.SchemeGroupVersion.WithKind("Secret")
+
+// isServiceBindingRequest checks whether the given obj is a Service Binding Request through GVK
+// comparison.
+func isServiceBindingRequest(obj runtime.Object) bool {
+	return obj.GetObjectKind().GroupVersionKind() == serviceBindingRequestGVK
+}
+
+// isSecret checks whether the given obj is a Secret through GVK comparison.
+func isSecret(obj runtime.Object) bool {
```
DhritiShikhar, line 39: We should also add a check to see if an object is `configMap`

pkg/controller/servicebindingrequest/mapper.go
```diff
@@ -12,7 +21,97 @@ var (
...
+		svcGVK := schema.GroupVersionKind{Group: svc.Group, Version: svc.Version, Kind: svc.Kind}
+		if obj.GetObjectKind().GroupVersionKind() == svcGVK {
+			return true
+		}
+	}
+	return false
+}
+
+// isApplication checks whether the given obj is an application in given sbr.
+func isApplication(
```
DhritiShikhar, line 56: We are using terminology like is**SBRService** which is nice. Then we should rename `isApplication` to is**SBRApplication** so that we know this is not just any application. This is a check for SBR's application.

pkg/controller/servicebindingrequest/mapper.go
```diff
@@ -12,7 +21,97 @@ var (
...
+		return false, err
+	}
+
+	isEqual := obj.GetObjectKind().GroupVersionKind() == appGVK
+
+	return isEqual, nil
+}
+
+// isSecretOwnedBySBR checks whether the given obj is a secret owned by the given sbr.
+func isSecretOwnedBySBR(obj metav1.Object, sbr *v1alpha1.ServiceBindingRequest) bool {
```
DhritiShikhar, line 77: The correct way to check if a secret is owned by an SBR would be through owner reference.

pkg/controller/servicebindingrequest/mapper.go
```diff
@@ -21,17 +120,61 @@ func (m *sbrRequestMapper) Map(obj handler.MapObject) []reconcile.Request {
...
+	for _, item := range sbrList.Items {
+		namespacedName := convertToNamespacedName(&item)
+
+		sbr, err := convertToSBR(item.Object)
+		if err != nil {
+			log.Error(err, "converting unstructured to SBR")
+			continue ITEMS
+		}
+
+		if isSecret(obj.Object) && isSecretOwnedBySBR(obj.Meta, sbr) {
```
DhritiShikhar, line 157: Just asking, is there any logical reason why secret check is before Service and Application check?

pkg/controller/servicebindingrequest/mapper.go
```diff
@@ -12,26 +21,170 @@ var (
...
+
+		if isSBRService(sbr, obj.Object) {
+			log.Debug("resource identified as service in SBR")
+			namespacedNamesToReconcile.add(namespacedName)
+		} else {
+			log.Debug("resource is not a service declared by the SBR")
+		}
+
+		if ok, err := isApplication(m.restMapper, sbr, obj.Object); err != nil {
+			log.Error(err, "identifying resource resource as SBR application")
```
DhritiShikhar, line 172: Typo -> identifying resource resource

pkg/controller/servicebindingrequest/mapper_test.go
```diff
@@ -4,44 +4,142 @@ import (
...
+		Spec: v1alpha1.ServiceBindingRequestSpec{
+			ApplicationSelector: v1alpha1.ApplicationSelector{
+				GroupVersionResource: metav1.GroupVersionResource{
+					Group:    "apps",
+					Version:  "v1",
+					Resource: "deployments",
+				},
+				ResourceRef: "mapper-unit-deployment",
+			},
+			BackingServiceSelector: &v1alpha1.BackingServiceSelector{
```
DhritiShikhar, line 38: since, we plan to deprecate `v1alpha1.BackingServiceSelector`, can we use `v1alpha1.BackingServiceSelectors` over here?

pkg/controller/servicebindingrequest/mapper.go
isutton: In this particular context only `corev1.Secret` resources should be checked since it might appear in the `.status.secret` field of a SBR; I can't see whether the comparison you're mentioning fits, could you clarify?

pkg/controller/servicebindingrequest/mapper.go
isutton: "Owning" here is a loose term;  the relationship between a `corev1.Secret` and a `ServiceBindingRequest` can happen in two distinct ways: one is as intermediate/intermediary secret through SBR's `.status.secret` and another is through services relationship.

In other words, a SBR owns the intermediate secret when the secret lives in the same namespace as the SBR and its name is present in `.status.secret`.

pkg/controller/servicebindingrequest/mapper.go
isutton: That's because we also need to check whether the secret is being "owned" by the SBR through its `.status.secret` field, which is not part of service and application checks.

isutton: I believe this block could also continue to the next `ITEMS` element at that point, since it is not important where the relationship has been identified but only that it has been identified.

pkg/controller/servicebindingrequest/mapper_test.go
isutton: Solved in bc678c3e3dcce1851d3ec5a006e9bd6d170f5c43.

pkg/controller/servicebindingrequest/mapper.go
isutton: Solved in ba54c6237f5d22cbaba2bb10e293aae6d518b253.

pkg/controller/servicebindingrequest/mapper.go
isutton: Solved in 34d695f6f7fa42e218ba58a7421436c77f175ae3.

DhritiShikhar:
> additionally, those annotations also seem to not be necessary to find a direct relationship between a Service Binding Request and a related resource, being it a service, application or the intermediate secret as changes in this PR can show.

Hmm.. To conclude, there are no side effects of not stamping annotations to SBR's related resources

pkg/controller/servicebindingrequest/mapper.go
DhritiShikhar: Oh Okay!

pkg/controller/servicebindingrequest/mapper.go
DhritiShikhar: @isuttonl I don't understand this comment :slightly_smiling_face: 

Isn't it a hack to determine the relationship between an intermediate secret and an SBR based on intermediate secret's name when we can do that with ownerReference?

Just trying to understand,
This intermediate secret was created by the Service Binding Operator.

It makes sense to check relationship between SBR - application based on name present in spec.Application.ResourceRef because SBR doesn't OWN it.

But in the case of intermediate secret, There is a more concrete way of checking the relationship with SBR.

Let me know if my understanding is not right. :) 

You mentioned "owning" here is a loose term. Why is that so?

pkg/controller/servicebindingrequest/mapper.go
DhritiShikhar: Okay Igor!

DhritiShikhar:
The changes seem good to be merged.

Waiting for Igor to address this -> https://github.com/redhat-developer/service-binding-operator/pull/596#discussion_r466986562

pkg/controller/servicebindingrequest/mapper.go
isutton: There's a couple of ownership concepts here:

1. Ownership from Kubernetes' garbage collector point-of-view, where the SBR is present in the Secret's `OwnerReferences` field; since it is a slice and lives in the resource's metadata and can be modified by the user, I believe it should be used only to coordinate resource deletion (blocking or cascading, for example).
2. Ownership from a SBO business logic point-of-view, represented by the `.status.secret` field in SBR. In other words the, intermediary/intermediate secret name is stored there and does not interfere with the garbage collector. 

I meant "loose" because the function is responsible for identifying 2) not 1) as you've previously thought.

isutton: Another point is that mapper is responsible for identifying the relationship of an arbitrary resource to a slice `ServiceBindingRequest` namespace/name pairs, and currently there are 3 distinct places this is encoded in the `ServiceBindingRequest`:

1. Applications
2. Services
3. `.status.secret`

DhritiShikhar: Okay. Thanks for the explaination.

isutton:
> > additionally, those annotations also seem to not be necessary to find a direct relationship between a Service Binding Request and a related resource, being it a service, application or the intermediate secret as changes in this PR can show.
> 
> Hmm.. To conclude, there are no side effects of not stamping annotations to SBR's related resource.

Not that I'm aware of... Perhaps the only one I can think of with is related to "nested" bindings, where for example a SBR might include a CR storing relevant data in a Secret: the current implementation won't match this Secret with a SBR but it can definitely be expanded to match resources related to the SBR at hand, in addition of checking whether the resource might appear as an application, service and intermediate/intermediary secret.

> There is only unidirectional tracing now. We can trace relationship from SBR to application but not from application to SBR.

Yes, and to be more explicit: you do like you would with a database: filter the SBRs that have declared an arbitrary resource as either application, service and/or intermediate/intermediary secret.

PR #598, opened 2020-08-05
Sunder the present feature file

pratikjagrut:
### Motivation

Currently, we have a single feature file this PR is to split that file into multiple feature files based on the feature we're testing.

### Testing

`make test-acceptance`


test/acceptance/features/bindToMultipleServices.feature
```diff
@@ -0,0 +1,55 @@
...
+                    group: apps
+                    version: v1
+                    resource: deployments
+                backingServiceSelector:
+                    group: postgresql.baiju.dev
+                    version: v1alpha1
+                    kind: Database
+                    resourceRef: db-demo-2
+            """
+        Then "nodejs-app" deployment must contain SBR name "binding-request-1" and "binding-request-2"
```
shvasude, line 51: Can we do something like this so that any number of binding-request is handled with the same step implementation: 
`        Then "nodejs-app" deployment must contain SBR name "binding-request-1" and "binding-request-2"
        Examples:
            | binding_request_name |
            | binding-request-1    |
            | binding-request-2    |`

Avni-Sharma:
Should we do it for all scenarios? What are the criteria for deciding a particular feature file vs having a scenario in the same file?

pratikjagrut:
> Should we do it for all scenarios? What are the criteria for deciding a particular feature file vs having a scenario in the same file?

Just test needs to be testing different feature. Like, in this case, binding multiple services is another feature.

baijum:
Splitting the feature file at this granularity looks good to me.

/lgtm

Avni-Sharma:
+1 for the feature file. Thanks for the insight @pratikjagrut . In that case I believe we can rename the feature to singleApptoMultipleServices. In the future we can have a multiple app to multiple services using matchLabels

pedjak:
> +1 for the feature file. Thanks for the insight @pratikjagrut . In that case I believe we can rename the feature to singleApptoMultipleServices. In the future we can have a multiple app to multiple services feature file when there are multiple apps using matchLabels. (multipleAppstoMultipleServices)

Agree with @Avni-Sharma here.

pratikjagrut:
Force push for squashing commits.

pratikjagrut:
@pedjak @Avni-Sharma made changes as you suggested please review and merge it.

pedjak:
@pratikjagrut the chosen names contain a bit of redundancy. For example `bindSingleAppToSiingleService` is semantically the same as `bindAppToService`, but shorter. I prefer the later. See https://github.com/redhat-developer/service-binding-operator/pull/598#pullrequestreview-462380661

pratikjagrut:
> @pratikjagrut the chosen names contain a bit of redundancy. For example `bindSingleAppToSiingleService` is semantically the same as `bindAppToService`, but shorter. I prefer the later. See [#598 (review)](https://github.com/redhat-developer/service-binding-operator/pull/598#pullrequestreview-462380661)

Okay, I'll revert it.

pratikjagrut:
Acceptance test failed due to some connection error: ` The connection to the server api.ci-op-kwqxgj35-f9842.origin-ci-int-aws.dev.rhcloud.com:6443 was refused - did you specify the right host or port?`


/test 4.5-acceptance

pratikjagrut:
There seems to be an issue with cluster provisioning or something from cluster side/
test-e2e failure: `{"component":"entrypoint","file":"prow/entrypoint/run.go:169","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Entrypoint received interrupt: terminated","severity":"error","time":"2020-08-07T16:07:39Z"}
time="2020-08-07T16:07:39Z" level=info msg="Received signal." signal=interrupt`

4.5-acceptance failure: `Pod pending timeout.`

/retest

PR #606, opened 2020-08-12
Improve logging in sbrRequestMapper

isutton:


pkg/controller/servicebindingrequest/mapper.go
```diff
@@ -132,6 +133,8 @@ func (m *sbrRequestMapper) Map(obj handler.MapObject) []reconcile.Request {
 		}
 		log.Debug("current resource is a SBR", "Requests", requests)
 		return requests
+	} else {
```
Avni-Sharma, line 136: else is redundant here

isutton: You're right in principle, but in order to make it easier to follow the debug logs more information should be added.

pedjak:
obsolete after #919 got merged.

Issue #587, opened 2020-07-27
Support the spec  ServiceBinding CRD


arthurdm:
The [Service Binding Specification](https://github.com/k8s-service-bindings/spec) has created a [new CRD](https://github.com/k8s-service-bindings/spec/blob/master/service.binding_servicebindings.yaml) called `ServiceBinding`.   It was inspired by SBO's `ServiceBindingRequest`, but has certain restrictions and small tuning from the spec.  

This issue will track SBO's adoption of this CRD, and move SBO closer to spec compliance (along with #403 and #548).

For reference, [this repo](https://github.com/vmware-labs/service-bindings/releases/tag/v0.1.0) has already implemented support for this CRD.

sbose78:
Ack.
We shall triage and evaluate which parts are reasonable to be adopted in the recent future.

sbose78:
Arthur, 
Given that we are under feature freeze ( other than existing PRs ) till the September 15th release, could you help me with the following :

1) Feature comparison between things in Service Binding operator & the spec. 

2) If you were to have new contributions from your team, how would the development plan look like? Could you or someone from your team help with that?

I'll let the newer contributors make detailed proposals and drive the discussion.

Note, the September 15th release addreses

1) Our exploration with strimzi team and how the annotations/descriptors were shaped since then.

2) Multiple backing services and its impact on all features.

3) General stability.

sbose78:
Here's a process to enable the enhancements to flow in
https://github.com/redhat-developer/service-binding-operator/pull/610

arthurdm:
thanks @sbose78 

@arturdzm - can you please work on the proposal outlined above?

sbose78:
Even better, the proposal process #610 has been merged, @arthurdm and @arthurdzm ( cc @scothis ), please feel free!

baijum:
Are we going to call this Tech Preview in GA release?

PR #513, opened 2020-06-16
Support custom containers path and secret path

baijum:
Many Application workloads do not support podSpec, and might need
association of the binding secret in an arbitrary location.
Since we are really concerned about `containers` and `volumes`, users
should be able to provide a custom path to them as well.
(Support for volumes will be introduced in another PR)

Example, in the below workload, the binding secret has to be associated
in
`spec.eventSourceSecret` instead of `spec.template.spec.containers`

```
apiVersion: sources.eventing.knative.dev/v1alpha1
 kind: KafkaSource
 metadata:
  name: my-kafka-source
 spec:
  consumerGroup: knative-group
  eventSourceSecret: eventsource-binding-foo-bar-binding-secret <---
Association made by service binding operator
  sink:
    ref:
      apiVersion: serving.knative.dev/v1alpha1
      kind: Service
      name: event-display
```

To accomplish the same, the `ServiceBindingRequest`'s
`applicationSelector` could look like the following:

```
apiVersion: apps.openshift.io/v1alpha1
kind: ServiceBindingRequest

metadata:
  name: eventsource-binding-foo-bar
  namespace: service-binding-demo

spec:
  applications:
    - bindingPath:
         secretPath: spec.eventSourceSecret # <--- In  KafkaSource
      group: sources.eventing.knative.dev
      resource: kafkasources
      resourceRef: my-kafka-source
      version: v1alpha1

  services:
    - group: kafka.strimzi.io
      version: v1beta1
      kind: KafkaTopic
      resourceRef: my-kafka-topic
    - group: kafka.strimzi.io
      version: v1beta1
      kind: Kafka
      resourceRef: my-kafka-cluster
```

New non-mandatory API field `bindingPath` introduced.

```
applications:
  - bindingPath:
       containers: spec.template.spec.containers
       customSecretPath: spec.custom.path
     group: apps.openshift.io
     resource: deploymentconfigs
     resourceRef: nodejs-rest-http-crud
     version: v1
```

**Default behaviour**
If `bindingPath` is not set, default containers/volumes path is used as
per podSpec.

Ensured that binding is successful with/without podSpec path set.

Based on Shoubhik's work in PR #353

PR #495, opened 2020-05-27
Change API Group and Kind, and Spec

sbose78:
#### This is how the new CR looks like

```
---
apiVersion: operators.coreos.com/v1alpha1
kind: ServiceBinding
metadata:
  name: example-servicebinding
  namespace: sbose
spec:
  services:
    - name: demo-database
      group: postgresql.baiju.dev
      kind: Database
      version: v1alpha1

  application:
    name: nodejs-rest-http-crud
    group: apps
    version: v1
    resource: deployments
```

NOTE: I've used the `operators.coreos.com` API Group for the time-being since we are in discussions with the OLM team on moving this project to github.com/operator-framework . However, folks are open to using a different API Group too. While I don't want to deviate too much, I am considering using an OperatorFramework-ish group & kind. Suggestions welcome! :)

#### What should you review?
You don't really need to start looking at all 80+ files changed.  
This file is what you need to be look at, to begin with:  [servicebinding_types.go](https://github.com/redhat-developer/service-binding-operator/blob/48b322fa709b3a2a21bf570f919d0edd4637c7f4/pkg/apis/apps/v1alpha1/servicebinding_types.go)

#### Primarily

* Align the ServiceBinding API with the [Proposal](https://github.com/operator-framework/enhancements/pull/12) which also takes into account the new specification.
* Open up the API for review by the OLM team. ( There's no point doing an API review for the API which is in master since it was going to change )

#### Also .. 

* Standardize use of Golang Kubernetes APIs. Example, used `corev1.LocalObjectReference` instead of `name` wherever application.

### Changes

- Set a Group and Kind which is Kubernetes / OperatorFramework friendly. 
```
apiVersion: operators.coreos.com/v1alpha1
kind: ServiceBinding
```

- [x]  Removed `spec.backingServiceSelector`

- [x]  Replaced `spec.applicationSelector` with `spec.application`, 
- [x]  Using `Application *Application` instead of `Application Application`  ( feedback by OLM team )
- [x]  Using `DetectBindingResources *bool` instead of `DetectBindingResources bool` ( feedback by OLM team )
- [x]  Replaced `spec.backingServiceSelectors` with `spec.services` 
- [x]  Replaced `spec.customEnvVar` with `spec.dataMapping` <--- we don't have agreement on this, we may keep this as is. 
- [x]  Replaced `resourceRef` in `application` and `services` to `name` `corev1.LocalObjectReference` 
[...]

PR #369, opened 2020-03-09
Avoid appearance of empty fields

sbose78:
### Motivation

If the following SBR is created with a populated `BackingServiceSelectors`... 

```
apiVersion: apps.openshift.io/v1alpha1
kind: ServiceBindingRequest
metadata:
  name: nodejs-rest-http-crud-d-demo-database-postgresql-d
  namespace: sbose
spec:
  applicationSelector:
    group: apps
    resource: deployments
    resourceRef: nodejs-rest-http-crud
    version: v1
  backingServiceSelectors:
    - group: postgresql.baiju.dev
      kind: Database
      resourceRef: demo-database
      version: v1alpha1
```
Post reconciliation, magically, the empty `BackingServiceSelector` also shows up:
```
apiVersion: apps.openshift.io/v1alpha1
kind: ServiceBindingRequest
metadata:
  creationTimestamp: '2020-03-09T04:44:50Z'
  finalizers:
    - finalizer.servicebindingrequest.openshift.io
  generation: 3
  name: nodejs-rest-http-crud-d-demo-database-postgresql-d
  namespace: sbose
  resourceVersion: '47774'
  selfLink: >-
    /apis/apps.openshift.io/v1alpha1/namespaces/sbose/servicebindingrequests/nodejs-rest-http-crud-d-demo-database-postgresql-d
  uid: cb7a2db1-3d88-4308-9224-893058a1cad4
spec:
  applicationSelector:
    group: apps
    resource: deployments
    resourceRef: nodejs-rest-http-crud
    version: v1
  backingServiceSelector:    # <--- This should not show up
    group: ''
    kind: ''
    resourceRef: ''
    version: ''
  backingServiceSelectors:
    - group: postgresql.baiju.dev
      kind: Database
      resourceRef: demo-database
      version: v1alpha1
  detectBindingResources: false
status:
  bindingStatus: BindingSuccess
  conditions:
    - lastHeartbeatTime: '2020-03-09T04:49:41Z'
      lastTransitionTime: '2020-03-09T04:44:51Z'
      status: 'True'
      type: Ready
  secret: nodejs-rest-http-crud-d-demo-database-postgresql-d

```

### Changes

- This PR changes both the `BackingServiceSelector` and `BackingServiceSelectors` as nil-able fields to avoid this from happening. This changes fixes the issue.

[...]

PR #608, opened 2020-08-13
Breaking api changes - Refactor ServiceBindingRequest api

akashshinde:
### Motivation

Ref: https://issues.redhat.com/browse/APPSVC-574

Primarily

- Align the ServiceBinding API with the [Proposal](https://github.com/operator-framework/enhancements/pull/12) which also takes into account the new specification.
- Open up the API for review by the OLM team. ( There's no point doing an API review for the API which is in master since it was going to change )

Also ..

- Standardize use of Golang Kubernetes APIs. Example, used corev1.LocalObjectReference instead of name wherever application.

### Changes
This is how ServiceBinding CR would look like
```
---
kind: ServiceBinding
metadata:
  name: example-servicebinding
  namespace: sbose
spec:
  services:
    - name: demo-database
      group: postgresql.baiju.dev
      kind: Database
      version: v1alpha1

  application:
    name: nodejs-rest-http-crud
    group: apps
    version: v1
    resource: deployments
```

Set a Group and Kind which is Kubernetes / OperatorFramework friendly.
```
kind: ServiceBinding
```

Following changes have been made in this PR

- [x] Updated `apps.openshift.io` group to `operators.coreos.com`

- [x]  Update `ServiceBindingRequest` kind to `ServiceBinding`

- [x]  Removed `spec.backingServiceSelector`

- [x]  Replaced `spec.applicationSelector` with `spec.application`

- [x]  Using `Application *Application` instead of `Application Application` ( feedback by OLM team )

- [x]  Using `DetectBindingResources *bool` instead of `DetectBindingResources bool` ( feedback by OLM team )

- [x]  Replaced `spec.backingServiceSelectors` with `spec.services`

- [x]  Replaced `resourceRef` in application and services to `name`

### Testing

- No extra tests added. Ensured existing ones work!
- Validated manually by creating a binding between a nodejs app & Database. (connecting single service)
- Validated manually by creating a binding between a nodejs app & Database and etcd (connecting multiple services)


deploy/olm-catalog/service-binding-operator/0.1.1/service-binding-operator.v0.1.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,171 @@
...
+    type: MultiNamespace
+  - supported: true
+    type: AllNamespaces
+  keywords:
+  - ""
+  maintainers:
+  - {}
+  maturity: alpha
+  provider: {}
+  replaces: service-binding-operator.v0.0.0
```
Avni-Sharma, line 170: shouldn't this be 0.1.1-352? The current operatorhub release.

deploy/olm-catalog/service-binding-operator/0.1.1/service-binding-operator.v0.1.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,171 @@
...
+          },
+          "spec": {
+            "application": {
+              "group": "apps",
+              "name": "nodejs-rest-http-crud",
+              "resource": "deployments",
+              "version": "v1"
+            },
+            "services": [{
+              "group": "postgresql.example.dev",
```
Avni-Sharma, line 21: this should be a list

deploy/olm-catalog/service-binding-operator/manifests/service-binding-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,176 @@
...
+            "name": "example-servicebinding"
+          },
+          "spec": {
+            "application": {
+              "group": "apps",
+              "resource": "deployments",
+              "name": "nodejs-rest-http-crud",
+              "version": "v1"
+            },
+            "services": [{
```
Avni-Sharma, line 20: services should be a list here

deploy/olm-catalog/service-binding-operator/0.1.1/service-binding-operator.v0.1.1.clusterserviceversion.yaml
pedjak: I think this file should not be part of git - it generated on the fly when we want to release the operator.

deploy/role.yaml
```diff
@@ -66,6 +66,21 @@ rules:
       - "list"
       - "watch"
       - "update"
+  - apiGroups:
```
pedjak, line 69: you should revisit the existing roles from above and remove those pointing to the old apigroups.

pedjak (changes requested):

deploy/olm-catalog/service-binding-operator/manifests/service-binding-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,176 @@
...
+    type: MultiNamespace
+  - supported: true
+    type: AllNamespaces
+  keywords:
+  - ""
+  maintainers:
+  - {}
+  maturity: alpha
+  provider: {}
+  version: 0.1.1
```
Avni-Sharma, line 176: Should we be on 0.1.1 or 0.2.0 as described here https://issues.redhat.com/browse/APPSVC-663

pedjak: I would leave it as it is and focus on version increase in another PR.

deploy/olm-catalog/service-binding-operator/manifests/service-binding-operator.clusterserviceversion.yaml
akashshinde: @Avni-Sharma it's already a list

Avni-Sharma: :+1:

deploy/olm-catalog/service-binding-operator/0.1.1/service-binding-operator.v0.1.1.clusterserviceversion.yaml
akashshinde: handled here https://github.com/redhat-developer/service-binding-operator/pull/608/commits/12dd1096680c741f78bdcf8aa8e5e93b08b5d07a

deploy/olm-catalog/service-binding-operator/manifests/service-binding-operator.clusterserviceversion.yaml
akashshinde: Resolving this

pedjak (changes requested):

pedjak:
could we rename `customEnvVar` to `env`, to match the specification better?

cdlliuy:
@sbose78 ,  @DhritiShikhar ,   please make sure PR #513 is merged before this changes.  Thanks.  I would like to pickup #513  first before involving this breaking changes.

pkg/apis/apps/v1alpha1/servicebinding_types.go
```diff
@@ -22,33 +22,26 @@ type ServiceBindingRequestSpec struct {
...
 	// backing service operator.
 	// +optional
-	ApplicationSelector ApplicationSelector `json:"applicationSelector"`
+	Application *Application `json:"application"`
 
 	// DetectBindingResources is flag used to bind all non-bindable variables from
 	// different subresources owned by backing operator CR.
 	// +optional
-	DetectBindingResources bool `json:"detectBindingResources,omitempty"`
+	DetectBindingResources *bool `json:"detectBindingResources,omitempty"`
```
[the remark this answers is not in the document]
akashshinde: @pedjak I was referring changes made on https://github.com/redhat-developer/service-binding-operator/pull/495, @sbose78 has mentioned this are the changes are the result of feedback he's received from OLM team. Anything to add to it @sbose78 ?

Makefile
```diff
@@ -409,13 +409,13 @@ deploy-rbac:
...
 deploy-crds:
-	$(Q)kubectl apply -f deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
+	$(Q)kubectl apply -f deploy/crds/operators.coreos.com_servicebindings_crd.yaml
 
 .PHONY: deploy-clean
 ## Deploy-Clean: Removing CRDs and CRs
 deploy-clean:
-	$(Q)-kubectl delete -f deploy/crds/apps_v1alpha1_servicebindingrequest_cr.yaml
-	$(Q)-kubectl delete -f deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
+	$(Q)-kubectl delete -f deploy/crds/apps_v1alpha1_servicebinding_cr.yaml
```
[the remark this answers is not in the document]
akashshinde: we can keep it as `servicebinding_cr.yaml`, wdyt @DhritiShikhar @isutton ?

pkg/controller/servicebinding/annotations/bindinginfo.go
```diff
@@ -40,7 +40,7 @@ func NewBindingInfo(name string, value string) (*bindingInfo, error) {
 	}
 	parts := strings.SplitN(cleanName, "-", 2)
 
-	resourceReferencePath := parts[0]
+	namerefPath := parts[0]
```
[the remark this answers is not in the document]
akashshinde: Not important but got changed because `resourceRef` to `name`, I will revert it.

pkg/apis/apps/v1alpha1/servicebinding_types.go
```diff
@@ -75,45 +69,45 @@ type BoundApplication struct {
 	corev1.LocalObjectReference `json:",inline"`
 }
 
-// ApplicationSelector defines the selector based on labels and GVR
-type ApplicationSelector struct {
+// Application defines the selector based on labels and GVR
+type Application struct {
+	corev1.LocalObjectReference `json:",inline"`
```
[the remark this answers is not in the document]
akashshinde: > could we rename `customEnvVar` to `env`, to match the specification better?

Don't we want to rename `customEnvVar` to `dataMapping` ? I see ticket raised for this https://issues.redhat.com/browse/APPSVC-538

akashshinde:
> could we rename `customEnvVar` to `env`, to match the specification better?

Don't we want to rename `customEnvVar` to `dataMapping` ? I see ticket raised for this https://issues.redhat.com/browse/APPSVC-538

Avni-Sharma:
> could we rename `customEnvVar` to `env`, to match the specification better?

According to spec `customEnvVar` should be `dataMapping` and not `env`. According to spec by default the values are injected as volume mounts. Env API then helps to make thmn env variables which is different from `customEnvVar`

sbose78:
A more env/vol agnostic name for the attribute would eventually be nice. If we don't like `dataMapping` as a name, we could use something else :)

pkg/apis/apps/v1alpha1/servicebinding_types.go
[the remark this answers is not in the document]
sbose78: Helps us differentiate between false and unspecified :)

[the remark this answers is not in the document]
pedjak: any particular usecase in mind? When unspecified, we should assume some default value, and that is implicit to user, and that could be confusing. If we do not have any concrete usecase, I would avoid this for now.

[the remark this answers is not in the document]
sbose78: I would do it to [avoid appearance of empty fields](https://github.com/redhat-developer/service-binding-operator/pull/369) and to be consistent with other changes in the current PR.

In the current PR, we are doing the same for `Application` too and hence feels natural to me that we would do the same for other attributes as well.

I would rather argue, 
* Do we have a concrete use case to show users fields they never cared about to begin with? If not, we should do the needful to avoid them showing up. 
* I'm OK with the opposite as well - as long as we are consistent.


Here was the feedback from the OLM team:

```
Hello,

Did a quick pass through just some things to consider:

The Application is optional but I don’t think that I would be able to set nil here. Making it a pointer to an Application I think is what we should do.

The DetectBindingResources is optional, but has a default of false. Is there ever going to potentially be a time when a non-set value is going to be different then a false value?

The Application type has a label selector and a LocalObjectReference do I need to set a name and then use a label selector? I think that these should be a union type

Thanks,

Shawn Hurley
```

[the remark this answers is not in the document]
pedjak: @sbose78 so the decision depends on the answer to "The DetectBindingResources is optional, but has a default of false. Is there ever going to potentially be a time when a non-set value is going to be different then a false value?" 

Currently, our code assumes that the value is false, if omitted. IMHO, I find it less surprising if the default value is `false`, it is something users are used to in all mainstream languages.

https://stackoverflow.com/questions/14157180/why-does-the-default-value-of-a-boolean-variable-tend-to-be-false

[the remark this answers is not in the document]
sbose78: > so the decision depends on the answer

Not necessarily, as in, not how I reasoned this change.

I proposed it to be a pointer for two reasons across all spec attributes.

* So that everything optional is nil-able.
* If the user doesn't set it, it shouldn't show up post creation.

No strong opinion here 🤷‍♂️  We could go either way.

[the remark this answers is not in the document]
pedjak: > We could go either way.

IThanks for providing the details about the reasoning. Still not sure if this particular field should be declared as optional, but instead of splitting hair further, let's see how this is going to work in future. We can always change that later.

deploy/crds/operators.coreos.com_servicebindings_crd.yaml
```diff
@@ -275,9 +230,37 @@ spec:
             mountPathPrefix:
               description: MountPathPrefix is the prefix for volume mount
               type: string
+            services:
```
baijum, line 233: Looks like this is a major gap between the spec and our implementation. The spec only supports only one `service` per `ServiceBinding` CRD. https://k8s-service-bindings.github.io/spec/#resource-type-schema-1
@sbose78 What would you suggest here?

sbose78: We would lose critical use cases with a single `service` and hence we'll stick to `services`. The CRD/API evolution will happen within the project and not **necessarily** as per the spec. 

After we reach a reasonable state of stability ( next upstream ) release, we'll use the enhancement proposal process to make well socialized and thought our changes from our existing state.

baijum: I would suggest not to use the word `ServiceBinding` to avoid confusion with the spec. This is not going to help our users. Let's stick with `ServiceBindingRequest` and continue with rest of the changes.

If the spec is going to co-exist in the Kubernetes community, it will be really confusing to our end-users. Sometimes they might read the spec and expect the same API here as well.

baijum: I can think of a couple more practical reason to not use `ServiceBinding`:

1. Potential future compatibility: If we choose later to adapt the spec, it would be easy as it will be new CRD
2. Community members can create some kind of bridge between the objects of spec CRD and our CRD.

sbose78: Would our API group being different make things any better, in your opinion ?

Note, we are never going to move into a new set of CRDs at one go, we'll continue to make incremental changes.
https://github.com/redhat-developer/service-binding-operator/issues/587#issuecomment-671682260

> If the spec is going to co-exist in the Kubernetes community, it will be really confusing to our end-users. Sometimes they might read the spec and expect the same API here as well.

I did consider this, however, since these APIs are not part of core kubernetes, there's no way a user would read the spec and want to try things out on their cluster. 

Users have to make a conscious choice to install a specific controller+CRD / operator. That should 'tell' them what to do next and which CRs to create next. The "user documentation" should really come from implementation and not the spec.

ron1: The spec suggests using multiple ServiceBinding resources to bind multiple services to a single application. What value does the the single SBRs support for multiple services offer over simply using multiple SBs each of which references the same application?

sbose78: Ron, good question!

Here's why we had gone down the path of multiple services:

1. In a lot of cases, to achieve anything reasonable, multiple services need to be used ( CR, secret, configmap ), unless the backing service authors had made code changes to include everything interesting in the same CR. This project aims to allow backing service authors to get away with absolutely bare minimum changes.

2. Having multiple objects ( SBs ) when you can have one feels like an overkill to me especially when you have a controller reconciling on those and modifying a workload as part of that.

baijum: 1. Everyone agrees that multiple services exist in the real world.    But that's not affecting the creators of backing services.
2. When I asked about the second point, this is the response I got from @scothis
   "multiple services would be much more complex to support. The `type`, `provisioner`, `mappings` and `env` are all specific to a single service" (Ref: https://kubernetes.slack.com/archives/C012F2GPMTQ/p1598966672013000?thread_ts=1598960730.010900&cid=C012F2GPMTQ )

scothis: There are a couple different approaches to support this behavior that are perfectly acceptable to the spec:
1. decompose a higher order resource into multiple `ServiceBinding` resources
2. create an alternative equivalent to the `ServiceBinding` resource that consumes the [Provisioned Service](https://github.com/k8s-service-bindings/spec#provisioned-service) and [Application Projection](https://github.com/k8s-service-bindings/spec#application-projection) pillars of the spec independently.

The primary intent of the spec is to provide a baseline set of behavior that is portable across implementations. Each implementation is free to go beyond the spec and provide high level or more capable behavior, at the cost of no longer being portable. You can choose the right experience to offer to your users and customers.

sbose78: > Everyone agrees that multiple services exist in the real world.

@baijum , 
Note, I was talking about one instance of a database service which gets represented as multiple "service binding" spec.services here. 

Example, a single PostgreSQL database service might be represented as 
* one CR, 
* one secret, 
* one ConfigMap  .. which aren't necessarily referenced in the CR

I wouldn't want to have the user create multiple SBs with each of the above objects listed as a `spec.service`.

> But that's not affecting the creators of backing services.

It does impact the backing service author if we ask the backing service author to ensure everything associated with one instance of a database is present in one CR.

I'm not making an argument against the specification's proposal, rather I'm arguing why we considered multiple backing services in the first place in a single `ServiceBinding[Request]`

scothis: > Note, I was talking about one instance of a database service which gets represented as multiple "service binding" spec.services here.

This concept was in RC1, but was removed for RC2 after a good bit of discussion. Forcing a user to be aware of multiple discrete parts that compose to represent a single service is not a great experience. In an ideal world, it's better to push that problem to the service provider to provide all the metadata a user needs to consume a service within a single package.

Knowing that we don't live in an ideal world, for the case where supporting existing resources is necessary, the ServiceBinding `mappings` are available to inject additional values or the [Binding Secrete Generation Strategies](https://github.com/k8s-service-bindings/spec#binding-secret-generation-strategies) spec extension. For even more advanced needs a new custom resource could aggregate complex state and expose a uniform [Provisioned Service(https://github.com/k8s-service-bindings/spec#provisioned-service).

sbose78: > Forcing a user to be aware of multiple discrete parts that compose to represent a single service is not a great experience

Not saying you said so :) But, "Service Binding" isn't the one really forcing that. The provisioned service forces that.

If a single package is available, spec.services[0] is all we need. As you mentioned, in an ideal world, provisioned services should be the ones making things easily consumable as a single package. We don't control the world of provisioned services, but we can make it a little easy to use existing provisioned services in service binding scenarios.

>  the ServiceBinding mappings are available to inject additional values or the Binding Secrete Generation Strategies spec extension

Unfortunately, these wouldn't work unless the discrete parts are referenced in the backing service CR's spec or status.

> For even more advanced needs a new custom resource could aggregate complex state and expose a uniform [Provisioned Service(https://github.com/k8s-service-bindings/spec#provisioned-service).

We all agree that there is value in aggregation. There are different ways to get there. Remember this project didn't have multiple services to begin with :) 

@baijum @scothis 

I'm not necessarily giving an opinion on the `ProvisionedService` , if you feel it is valuable, please create an enhancement proposal ( which can be as simple as a Github issue ) to begin with. We do need to have a path to go from existing implementation to the suggested one and therefore a proposal with details would be awesome.

pedjak:
@akashshinde `deploy/olm-catalog/service-binding-operator/0.0.23/service-binding-operator.v0.0.23.clusterserviceversion.yaml` needs to updated as well, in particular the embedded example.

deploy/crds/operators.coreos.com_servicebindings_crd.yaml
arthurdm: just to add my 2 cents :)  I used to also really like specifying multiple services on the same `ServiceBinding`, but two points that made me agree with a single service per `ServiceBinding` were:
* it's easier for a tool to add and remove services by simply adding or removing independent CRs, instead of having to deal with editing (and potentially corrupting, if concurrent updates are not well synchronized) an array inside a CR. It simply becomes "kubectl apply .." and "kubectl delete .."
* cleaner association of attributes with the service.  For example,  when we identify a set of extra mappings or environment variables in a `ServiceBinding`, which service does that get associated with if we have multiple services?  Or if we want to change the service name or type that gets mounted, we again run into the issue that they all require a way to identify the service you want to apply those attributes against.

pkg/apis/apps/v1alpha1/servicebinding_types.go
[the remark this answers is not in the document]
akashshinde: @pedjak `Application` uses `GVR` where as `BoundApplication` uses `GVK`. If we want to share application reference there would quite a few changes we need to make in order to support it. So not sure if that should be in the scope of this PR.

pedjak (approved):

PR #612, opened 2020-08-15
docs: Add a tabular roadmap

sbose78:
.. highlighting the 
* feature set
* maturity of the feature
* status of implementation

## What's next

Add a paragraph or two for each feature and move this section to a different file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,3 +107,32 @@ Meeting Agenda is maintained [here](https://github.com/redhat-developer/service-
 Please file bug reports on [Github](https://github.com/redhat-developer/service-binding-operator/issues/new). For any other questions, reach out on `service-binding-support@redhat.com`.
 
 
+## Roadmap
+
+
+**Key**
+
+⚪️ Concept validated
+
+:white_check_mark: Works, has tests and at least one documented scenario
+
+
+
+### Maturity matrix
+
+| Feature                                        | Alpha                | Beta | GA |
+| -----------------------------------------------| -------------------- | ---- | -- |
+| Binding with Kubernetes resources              |  :white_check_mark:  |      |    |
+| Binding with CRD-backed services               |  :white_check_mark:  |      |    |
+| Binding with OLM CRD-backed services           |  :white_check_mark:  |      |    |
+| Binding with secrets/configmaps                |  :white_check_mark:  |      |    |
+| Binding of PodSpec-based workloads             |  :white_check_mark:  |      |    |
+| Binding of non-PodSpec-based workloads         |  ⚪️    |      |    |
+| Binding with multiple-backing services         |  :white_check_mark:  |      |    |
+| Custom binding variables                       |  :white_check_mark:  |      |    |
+| Auto-detect binding resources                  |  :white_check_mark:  |      |    |
+| Security: Validating admission webhook         |   ⚪️             |      |    |
+| Support on Kubernetes                         |   :white_check_mark: |      |    |
+| Support on OpenShift                          |   :white_check_mark: |      |    |
+| Adoption by third-party backing services  |    ⚪️   |      |    |
+
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-# Connecting Applications with Services
+# The Service Binding Operator
+## Connecting Applications with Services in Kubernetes
 
 <p align="center">
     <a alt="GoReport" href="https://goreportcard.com/report/github.com/redhat-developer/service-binding-operator">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -107,3 +107,32 @@ Meeting Agenda is maintained [here](https://github.com/redhat-developer/service-
...
 
 
+## Roadmap
+
+
+**Key**
+
+⚪️ Concept validated
+
+:white_check_mark: Works, has tests and at least one documented scenario
```
pedjak, line 117: I would put the under the table.

Could you please clarify what is the difference between "concept validated" and "Works, has tests and at least one documented scenario". Should we just have a single check mark, and by positioning it in the right column we signal its maturity? Is "Beta" equal to "Technical Preview"?

sbose78: > "concept validated" 
This means that there's enough prior work done to understand that building this is possible. 


> "Works, has tests and at least one documented scenario".
Code that powers this is in master.

sbose78: > I would put the under the table.

Sure! Will do.

pedjak: Instead of introducing multiple checkmark symbols, how about that we add additional column for "conceptually validated", i.e. to name it "Planned" or "Ready for implementation"? And perhaps even link to the corresponding ehancement/github issue?

sbose78: > how about that we add additional column for "conceptually validated", i.e. to name it "Planned" or "Ready for implementation"? And perhaps even link to the corresponding enhancement/github issue?

I'm expecting that we will have two states per maturity level in many scenarios, because of newer production readiness clauses we find out. Hence, adding one new column wouldn't quite help I guess.

As an example, for "security", while an additional validating webhook is good enough in "alpha", there might be additional RBAC needs that would have to be taken are of, which will make it incorrect to be a ✅ in Beta even though it was a ✅ in Alpha :)

sbose78: > And perhaps even link to the corresponding enhancements/github issue?

( check PR description )
I was planning to have a longer doc for each feature with linked issues/PRs/proposals. What do you think?

README.md
```diff
@@ -107,3 +107,32 @@ Meeting Agenda is maintained [here](https://github.com/redhat-developer/service-
...
+| Binding with Kubernetes resources              |  :white_check_mark:  |      |    |
+| Binding with CRD-backed services               |  :white_check_mark:  |      |    |
+| Binding with OLM CRD-backed services           |  :white_check_mark:  |      |    |
+| Binding with secrets/configmaps                |  :white_check_mark:  |      |    |
+| Binding of PodSpec-based workloads             |  :white_check_mark:  |      |    |
+| Binding of non-PodSpec-based workloads         |  ⚪️    |      |    |
+| Binding with multiple-backing services         |  :white_check_mark:  |      |    |
+| Custom binding variables                       |  :white_check_mark:  |      |    |
+| Auto-detect binding resources                  |  :white_check_mark:  |      |    |
+| Security: Validating admission webhook         |   ⚪️             |      |    |
```
pedjak, line 134: not sure if this could be called "feature" - IMHO it sounds as the implementation approach.

IMHO, this table should contain features - functionalities visible to users. Also, an entry should not be added to this table before it lands in some format in the code base.

README.md
```diff
@@ -107,3 +107,32 @@ Meeting Agenda is maintained [here](https://github.com/redhat-developer/service-
...
+| Support on Kubernetes                         |   :white_check_mark: |      |    |
+| Support on OpenShift                          |   :white_check_mark: |      |    |
+| Adoption by third-party backing services  |    ⚪️   |      |    |
```
pedjak, line 137: IMHO, this is not a feature, or can you explain what is meant by this?

README.md
```diff
@@ -107,3 +107,32 @@ Meeting Agenda is maintained [here](https://github.com/redhat-developer/service-
...
+
+:white_check_mark: Works, has tests and at least one documented scenario
+
+
+
+### Maturity matrix
+
+| Feature                                        | Alpha                | Beta | GA |
+| -----------------------------------------------| -------------------- | ---- | -- |
+| Binding with Kubernetes resources              |  :white_check_mark:  |      |    |
```
pedjak, line 125: can we be more specific here? what resources exactly? Does this include support for secrets, configmap, CRs? If yes, do we need to list explicitly in the table?

README.md
```diff
@@ -107,3 +107,32 @@ Meeting Agenda is maintained [here](https://github.com/redhat-developer/service-
...
+| Binding with CRD-backed services               |  :white_check_mark:  |      |    |
+| Binding with OLM CRD-backed services           |  :white_check_mark:  |      |    |
+| Binding with secrets/configmaps                |  :white_check_mark:  |      |    |
+| Binding of PodSpec-based workloads             |  :white_check_mark:  |      |    |
+| Binding of non-PodSpec-based workloads         |  ⚪️    |      |    |
```
pedjak, line 130: most of the time we use the term `applications`, now here we have `workloads`. Let's be consistent.

README.md
sbose78: This implies "any kubernetes resource" which isn't necessarily backed by an operator.

sbose78:
I could stop using the term 'feature' in feature matrix if needed, but we need to represent the information that we are actively considering the following in our radar : 

* adoption by third party backing services.
* security / admission webhooks. 

I wish to avoid a different section altogether just for a couple of items.

README.md
sbose78: Reason I used `workloads` is because `podSpec`-based applications isn't a commonly used term. Usually, `podSpec`-based workloads is the term that's friendlier to kubernetes users.

pedjak: do we really need to emphasize here `podSpec` at all? Could we simply not list all k8s resources containing podSpec?

README.md
pedjak: Do we then treat support of resource X as a separate future? Should we remove this line and list explicitelly each supported resource when gets supported?

README.md
sbose78: We do, to reflect that CRs with a podSpec are perfectly legal in the eyes of Service Binding.

pedjak:
> I wish to avoid a different section altogether just for a couple of items.

A number of items could /will grow over time, so let's not try to squeeze this forever, nothing is set in stone - it is a live document at the end. I would prefer narow/focused scope now, and develop it from there over time.

README.md
sbose78: Unless the resource X deserves a special mention, I wouldn't list it. At the moment, we don't really restrict any resource as long as RBAC allows it.

sbose78:
Sorry :) Didn't get you.

> adoption by third party backing services.
security / admission webhooks.

This is a roadmap doc, and not mentioning the above makes it critically incomplete.

README.md
pedjak: Binding with Kubernetes resources  = Binding of PodSpec-based workloads + Binding of non-PodSpec-based workloads , right? If yes, this line could be removed then?

PR #618, opened 2020-08-17
docs: Update best practices

sbose78:
.. to reflect that this project supports non-operator-backed-services as it was just another Kubernetes resource.

### DIFF
--- a/docs/OperatorBestPractices.md
+++ b/docs/OperatorBestPractices.md
@@ -1,25 +1,58 @@
-# Operator Best Practices
+# Backing Service Best Practices
 
 ## Introduction
 
 The goals of the Service Binding Operator is to make it easier for
 applications developers to bind applications with needed backing
-services, without having to perform manual configuration of secrets,
-configmaps, etc. and to assist operator providers in promoting and
+services, without having to perform manual configuration of `secrets`,
+`configmaps`, etc. and to assist operator providers in promoting and
 expanding the adoption of their operators.
 
-When a ServiceBindingRequest is created the Service Binding Operator
+When a `ServiceBindingRequest` is created the Service Binding Operator
 collects binding information and shares it with application. The
 Binding Service Operator's controller injects the binding information
-into the application's "DeploymentConfig", "Deployment" or “Replicaset"
-as environment variables via an intermediate Secret called "binding-request".
-The binding also works with Knative services as it works with any "deployment"
-controller which has the podspec defined in the its jsonpath as
+into the application's `DeploymentConfig`, `Deployment` or `Replicaset`
+as environment variables via an intermediate Secret.
+The binding also works with Knative `Services` as it works with any API which has the podspec defined in the its jsonpath as
 "spec.template.spec.containers".
 
 This document provides "best practices" guidelines for the development of
 Operators that manage backing services to be bound together with applications
 by the Service Binding Operator.
+
+## Making a Kubernetes Resource Bindable
+
+In order to make backing service bindable, the Kubernetes resource representing the same needs to be annotated as a means to express what is "interesting" for applications.
+
+Example, in an `Ingress` or `Route` resource representing a backing service, the `host` and `port` among other things would be "interesting" for applications for connecting to the backing service.
+
+To do so, the Kubernetes resource may be annotated to meaningfully convey what is "interesting" for binding.
+
+``` yaml
+kind: Route
+apiVersion: route.openshift.io/v1
+metadata:
+  name: example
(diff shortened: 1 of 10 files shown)


docs/OperatorBestPractices.md
```diff
@@ -1,26 +1,59 @@
-# Operator Best Practices
+# Backing Service Best Practices
```
pedjak, line 1: I would rename the file itself as well - instead of `OperatorBestPractices.md` to `BackingServicesBestPractices.md`.

docs/OperatorBestPractices.md
```diff
@@ -1,26 +1,59 @@
...
+into the application's `DeploymentConfig`, `Deployment` or `Replicaset`
+as environment variables via an intermediate Secret.
+The binding also works with Knative `Services` as it works with any API which has the podspec defined in the its jsonpath as
 "spec.template.spec.containers".
 
 This document provides "best practices" guidelines for the development of
 Operators that manage backing services to be bound together with applications
 by the Service Binding Operator.
 
+## Making a Kubernetes Resource Bindable
```
pedjak, line 23: By reading the document, I would say that the focus of the whole doc is to show the best practices how a Kubernets Resource can be consumed by an apllication. How about making this as the main title, for example:

"Best Practices for Making a Kube Resource Bindable"

This Section talks about adding annotations to help SBO to discover interesting data. So how about:

"Declare Bindable Data via Annotation"

docs/OperatorBestPractices.md
```diff
@@ -1,26 +1,59 @@
...
+
+``` yaml
+kind: Route
+apiVersion: route.openshift.io/v1
+metadata:
+  name: example
+  namespace: service-binding-demo
+  annotations:
+    openshift.io/host.generated: 'true'
+    servicebindingoperator.redhat.io/spec.host: 'binding:env:attribute' #annotate here.
```
pedjak, line 39: it would be great if we could have a doc section about all possible values an annotation can have.

docs/OperatorBestPractices.md
```diff
@@ -52,8 +85,8 @@ apiVersion: apiextensions.k8s.io/v1beta1
 metadata:
   name: databases.postgresql.baiju.dev
   annotations:
-    servicebindingoperator.redhat.io/status.dbConfigMap.password: 'binding:env:object:secret'
-    servicebindingoperator.redhat.io/status.dbConfigMap.username: 'binding:env:object:configmap'
+    servicebindingoperator.redhat.io/status.dbConfigMap-password: 'binding:env:object:secret'
```
pedjak, line 88: perhaps we could share more details about what is the semantic of these lines?

docs/OperatorBestPractices.md
sbose78: I would like to re-organize the doc to include the new annotations' format anyway in a separate doc since it might be too much information without linking out. I plan to bring in this

https://github.com/k8s-service-bindings/spec#annotation-examples

In most non-CR resources,

```
 servicebindingoperator.redhat.io/spec.SOMETHING: 'binding:env:attribute'
```
is that'll work.

docs/OperatorBestPractices.md
sbose78: Good point!

docs/OperatorBestPractices.md
sbose78: Also, we gotta document use of Secrets/Configmaps which I haven't tried out, but I heard works.

docs/OperatorBestPractices.md
sbose78: Good idea! 
Fixed up an annotation too. 
https://github.com/redhat-developer/service-binding-operator/pull/618/commits/2cf29e9a568009a4bf20dd40afb3dbf256f02b42

docs/OperatorBestPractices.md
sbose78: Done!

sbose78: need to update readme too..

docs/OperatorBestPractices.md
sbose78: update the title
https://github.com/redhat-developer/service-binding-operator/pull/618/commits/268279c3d0745a5c4bc460537c96761f0872c724

> "Declare Bindable Data via Annotation"

Annotation is more of a detail, I'd say.

sbose78:
@pedjak May I take this in given that this makes our docs better than what we have, if not perfect? :)

PR #597, opened 2020-08-04
Trigger reconcile on updates to services' status

Avni-Sharma:
Fix #508 

Follow steps mentioned here https://github.com/redhat-developer/service-binding-operator/issues/508#issuecomment-646777684
to test the functionality

Steps to reproduce:

Install the Strimzi Operator (v0.18.0).
Install the Service Binding Operator./ Run the SBO locally
Create ServiceBindingRequest, application, and the backing service (Kafka) resources:

```
apiVersion: apps.openshift.io/v1alpha1
kind: ServiceBindingRequest
metadata:
  name: kafka-producer-binding
spec:
  applicationSelector:
    resourceRef: nginx-deployment
    group: apps
    version: v1
    resource: deployments
  backingServiceSelectors:
    - group: kafka.strimzi.io
      kind: Kafka
      resourceRef: my-cluster
      version: v1beta1
      id: kaf
  customEnvVar:
    - name: BOOTSTRAP_SERVERS
      value: |-
        {{- range .kaf.status.listeners -}}
          {{- if and (eq .type "tls") (gt (len .addresses) 0) -}}
            {{- with (index .addresses 0) -}}
              {{ .host }}:{{ .port }}
            {{- end -}}
          {{- end -}}
        {{- end -}}
```

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 0 # tells deployment to deploy no pod
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
```

```
apiVersion: kafka.strimzi.io/v1alpha1
kind: Kafka
metadata:
  name: my-cluster
  labels:
    app: my-cluster
spec:
  kafka:
    replicas: 3
    resources:
      requests:
        memory: 2Gi
        cpu: 500m
      limits:
        memory: 2Gi
        cpu: "1"
    jvmOptions:
      -Xms: 1024m
      -Xmx: 1024m
    listeners:
      plain: {}
      tls:
        authentication:
          type: tls
    authorization:
      type: simple
      superUsers:
        - CN=my-connect
        - my-connect
        - ANONYMOUS
    config:
      offsets.topic.replication.factor: 3
      transaction.state.log.replication.factor: 3
      transaction.state.log.min.isr: 2
    storage:
      type: ephemeral
  zookeeper:
    replicas: 3
    resources:
[...]

PR #620, opened 2020-08-18
Acceptance test for validating binding using OLM descriptors

pratikjagrut:
### Motivation

Verifying the intermediate secret is getting populated with the bindable information present in backend service csv.
Ticket: https://issues.redhat.com/browse/APPSVC-645

### Testing

`make test-acceptance`

### DIFF
--- a/test/acceptance/features/bindAppToService.feature
+++ b/test/acceptance/features/bindAppToService.feature
@@ -11,7 +11,7 @@ Feature: Bind an application to a service
     Scenario: Bind an imported Node.js application to PostgreSQL database in the following order: Application, DB and Service Binding
         Given Imported Nodejs application "nodejs-rest-http-crud-a-d-s" is running
         * DB "db-demo-a-d-s" is running
-        When Service Binding is applied to connect the database and the application
+        When Service Binding is applied
             """
             apiVersion: operators.coreos.com/v1alpha1
             kind: ServiceBinding
@@ -47,7 +47,7 @@ Feature: Bind an application to a service
 
     Scenario: Bind an imported Node.js application to PostgreSQL database in the following order: Application, Service Binding and DB
         Given Imported Nodejs application "nodejs-rest-http-crud-a-s-d" is running
-        * Service Binding is applied to connect the database and the application
+        * Service Binding is applied
             """
             apiVersion: operators.coreos.com/v1alpha1
             kind: ServiceBinding
@@ -75,7 +75,7 @@ Feature: Bind an application to a service
     Scenario: Bind an imported Node.js application to PostgreSQL database in the following order: DB, Application and Service Binding
         Given DB "db-demo-d-a-s" is running
         * Imported Nodejs application "nodejs-rest-http-crud-d-a-s" is running
-        When Service Binding is applied to connect the database and the application
+        When Service Binding is applied
             """
             apiVersion: operators.coreos.com/v1alpha1
             kind: ServiceBinding
@@ -104,7 +104,7 @@ Feature: Bind an application to a service
     @disabled
     Scenario: Bind an imported Node.js application to PostgreSQL database in the following order: DB, Service Binding and Application
         Given DB "db-demo-d-s-a" is running
-        * Service Binding is applied to connect the database and the application
+        * Service Binding is applied
             """
             apiVersion: operators.coreos.com/v1alpha1
             kind: ServiceBinding
@@ -132,7 +132,7 @@ Feature: Bind an application to a service
     # Currently disabled as not supported by SBO
     @disabled
     Scenario: Bind an imported Node.js application to PostgreSQL database in the following order: Service Binding, Application and DB
-        Given Service Binding is applied to connect the database and the application
+        Given Service Binding is applied
             """
             apiVersion: operators.coreos.com/v1alpha1
             kind: ServiceBinding
@@ -160,7 +160,7 @@ Feature: Bind an application to a service
     # Currently disabled as not supported by SBO
     @disabled
     Scenario: Bind an imported Node.js application to PostgreSQL database in the following order: Service Binding, DB and Application
-        Given Service Binding is applied to connect the database and the application
+        Given Service Binding is applied
             """
             apiVersion: operators.coreos.com/v1alpha1
             kind: ServiceBinding
@@ -190,7 +190,7 @@ Feature: Bind an application to a service
     Scenario: Attempt to bind a non existing application to PostgreSQL database
         Given DB "db-demo-missing-app" is running
         * Imported Nodejs application "nodejs-missing-app" is not running
-        When Service Binding is applied to connect the database and the application
+        When Service Binding is applied
             """
             apiVersion: operators.coreos.com/v1alpha1
             kind: ServiceBinding
@@ -216,7 +216,7 @@ Feature: Bind an application to a service
     Scenario: Service Binding without application selector
         Given Imported Nodejs application "nodejs-empty-app" is running
         And DB "db-demo-empty-app" is running
-        When Service Binding is applied to connect the database and the application
+        When Service Binding is applied
             """
             apiVersion: operators.coreos.com/v1alpha1
             kind: ServiceBinding
@@ -236,7 +236,7 @@ Feature: Bind an application to a service
 
     Scenario: Backend Service status update gets propagated to the binding secret
         Given OLM Operator "backend" is running
-        * Backend CR "backend-demo" is applied
+        * The Custom Resource is present
             """
             apiVersion: "stable.example.com/v1"
             kind: Backend
@@ -248,7 +248,7 @@ Feature: Bind an application to a service
             spec:
                 host: example.common
             """
-        * Service Binding request is applied
+        * Service Binding is applied
             """
             apiVersion: operators.coreos.com/v1alpha1
             kind: ServiceBinding
@@ -270,7 +270,8 @@ Feature: Bind an application to a service
         Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-request-backend" should be changed to "True"
         And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-backend" should be changed to "False"
         Then Secret "binding-request-backend" contains "CustomReady" key with value "<no value>"
-        When Backend status in "backend-demo" is updated
+        # Backend status in "backend-demo" is updated
+        When The Custom Resource is present
             """
             apiVersion: "stable.example.com/v1"
             kind: Backend
@@ -289,7 +290,7 @@ Feature: Bind an application to a service
     Scenario: Custom environment variable is injected into the application under the declared name ignoring global and service env prefix
         Given Imported Nodejs application "nodejs-rest-http-crud-a-d-c" is running
         * DB "db-demo-a-d-c" is running
-        When Service Binding is applied to connect the database and the application
+        When Service Binding is applied
             """
             apiVersion: operators.coreos.com/v1alpha1
             kind: ServiceBinding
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -517,3 +517,31 @@ dev-release:
...
+## get_sbo_pkg
+get_sbo_pkg:
+	$(Q)wget https://raw.githubusercontent.com/redhat-developer/service-binding-operator-manifests/master/service-binding-operator.package.yaml -O $(SBO_PKG)
+
+
+.PHONY: get_csv_version
+## get_csv_version
+get_csv_version:
+	$(eval latestCSV := $(shell grep -wo "service-binding-operator\.v.*" $(SBO_PKG)))
+	$(eval latestCSVver := $(subst service-binding-operator.v, , $(latestCSV)))
```
pedjak, line 546: what is the purpose of this? we should not grab "the latest CSV" and deploy it to the cluster, because it might not correspond to the status we have on the branch under test. If we need to deploy SBO CSV (which I highly doubt), we should generate the actual version from the code in the given branch and then deploy it.

test/acceptance/features/bindAppToService.feature
```diff
@@ -217,3 +217,56 @@ Feature: Bind an application to a service
         And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding Request "binding-request-empty-app" should be changed to "False"
         And jq ".status.conditions[] | select(.type=="InjectionReady").reason" of Service Binding Request "binding-request-empty-app" should be changed to "EmptyApplicationSelector"
 
+    Scenario: Bind an imported Node.js application to PostgreSQL database in the following order: CSV, DB, App and SBR
```
pedjak, line 220: Unclear that is the scenario about, please rephrase. What are we testing here?

test/acceptance/features/bindAppToService.feature
```diff
@@ -217,3 +217,56 @@ Feature: Bind an application to a service
...
+            """
+        Then application should be re-deployed
+        And application should be connected to the DB "db-demo-c-d-a-s"
+        And jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding Request "binding-request-c-d-a-s" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding Request "binding-request-c-d-a-s" should be changed to "True"
+
+    Scenario: Bind an imported Node.js application to PostgreSQL database in the following order: CSV, APP, DB and SBR
+        Given SBO CSV is installed
+        * Imported Nodejs application "nodejs-rest-http-crud-c-a-d-s" is running
+        * DB "db-demo-c-a-d-s" is running
```
pedjak, line 250: this and previous test setup are identical, because at the time SBR is created, the system state is the same: nodejs application is there and db is present as well. Thus, one of these scenarios could be fully removed.

test/acceptance/features/bindAppToService.feature
```diff
@@ -217,3 +217,56 @@ Feature: Bind an application to a service
         And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding Request "binding-request-empty-app" should be changed to "False"
         And jq ".status.conditions[] | select(.type=="InjectionReady").reason" of Service Binding Request "binding-request-empty-app" should be changed to "EmptyApplicationSelector"
 
+    Scenario: Bind an imported Node.js application to PostgreSQL database in the following order: CSV, DB, App and SBR
+        Given SBO CSV is installed
```
pedjak, line 221: SBO CSV is not used during the lifetime of SBO at all. What we do use are CSV of operators managing backing services. In this case it would DB operator.

pedjak (changes requested):

Makefile
pratikjagrut: > what is the purpose of this?

We generate new CSV and push it to service-binding-operator-manifests repo so this target will download the latest CSV pushed. 

> If we need to deploy SBO CSV (which I highly doubt), we should generate the actual version from the code in the given branch and then deploy it.

This test was already there in e2e test suite, so according to ticket says I just converted it to the acceptance test.

test/acceptance/features/bindAppToService.feature
pratikjagrut: This is same as other scenarios in the list, just adding CSV step in it.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Yes the state is same but we are just adding CSV in the step, if this addition does not make any change then we can completely drop this conversion scenario.

test/acceptance/features/bindAppToService.feature
pratikjagrut: There was no sbo CSV assertion method in e2e so I kept it same other than successful installation in the start.

Makefile
pratikjagrut: code removed.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Removed gratuitous scenario.

test/acceptance/features/bindAppToService.feature
pratikjagrut: I've updated the setup, please review it.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Removed SBO csv.

test/acceptance/features/steps/steps.py
```diff
@@ -190,10 +190,38 @@ def then_sbo_jq_is(context, jq_expression, sbr_name, json_value_regex):
...
-    result | should.be_equal_to("[map[secretRef:map[name:binding-request-1]] map[secretRef:map[name:binding-request-2]]]")\
-        .desc(f'{app_name} deployment should contain secretRef: {sbr_name1} and {sbr_name2}')
+    result | should.be_equal_to(secret_ref)\
+        .desc(f'{app_name} deployment should contain secretRef: {secret_ref}')
+
+
+@given(u'Backend service "{csv_name}" CSV is installed')
+def install_csv(context, csv_name):
+    openshift = Openshift()
+    yaml_loc = "./test/acceptance/resources/backservClusterServiceVersion.yaml"
```
pedjak, line 205: `csv_name` should be used as the parameter to locate the right yaml file. Right now, not matter what we pass there, we always apply the same yaml file.

why do not we pass CSV content as the part of the step, like we do in other steps?

test/acceptance/features/bindAppToService.feature
```diff
@@ -217,3 +217,56 @@ Feature: Bind an application to a service
...
+                      storage: true
+                scope: Namespaced
+                names:
+                    plural: backservs
+                    singular: backserv
+                    kind: Backserv
+                    shortNames:
+                    - bs
+            """
+        * The application CR "csv-demo-backend" is present
```
pedjak, line 242: this is service CR not `application CR`

test/acceptance/features/bindAppToService.feature
```diff
@@ -217,3 +217,56 @@ Feature: Bind an application to a service
...
+                    group: apps
+                    version: v1
+                    resource: deployments
+                backingServiceSelector:
+                    group: service.example.com
+                    version: v1
+                    kind: Backserv
+                    resourceRef: csv-demo-backend
+            """
+        Then "csv-demo-app" deployment must contain secret ref "[map[secretRef:map[name:csv-demo-service-binding-request]]]"
```
pedjak, line 270: this check is not enough - we want to check if the secret contains values defined by binding declared in CSV. The current check does not tell us that at all.

test/acceptance/features/bindAppToService.feature
```diff
@@ -217,3 +217,56 @@ Feature: Bind an application to a service
...
+        * The application CR "csv-demo-backend" is present
+            """
+            apiVersion: service.example.com/v1
+            kind: Backserv
+            metadata:
+                name: csv-demo-backend
+            spec:
+                svcName: csv-demo-backend
+            """
+        * Imported Nodejs application "csv-demo-app" is running
```
pedjak, line 251: why do we install nodejs application if we do not check later on if it really works?

pedjak (changes requested):

pratikjagrut:
Resolved conflict and forced pushed the changes.

pratikjagrut:
@pmacik Consider this PR for review.

test/acceptance/features/bindAppToService.feature
pratikjagrut: I wrote it to reuse existing steps from [steps.py](https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/steps/steps.py#L291)

If it's not ok then I can update the steps.py and then other tests using it.

test/acceptance/features/steps/steps.py
pratikjagrut: > `csv_name` should be used as the parameter to locate the right yaml file. Right now, not matter what we pass there, we always apply the same yaml file.

I will rename the file, and will pass it a a variable in path.

> why do not we pass CSV content as the part of the step, like we do in other steps?
It's a file of 46 lines. Passing it directly will unnecessarily increase the length of feature file.

test/acceptance/features/bindAppToService.feature
pratikjagrut: We're asserting if its deployment contains the injected secretRef or not.

test/acceptance/features/steps/steps.py
pedjak: that could be then the argument for all yaml content we pass. Having CSV right in feature file enables the reader to immediately see what is going on. BTW, I am pretty sure that CSV file could be reduced by omitting unnecessary parts.

test/acceptance/features/bindAppToService.feature
pedjak: you can reuse the code logic, but logically the step is about adding service CR, and a reader should not be confused. You can make the step more generic as well.

test/acceptance/features/bindAppToService.feature
pedjak: yes, but saying that application is running mean that it runs correctly, and this is not the case since we do not connect it to any database.

test/acceptance/features/steps/steps.py
pratikjagrut: OK.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Removed application.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Removed this step.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Made this step generic.

test/acceptance/features/bindAppToService.feature
```diff
@@ -217,3 +217,177 @@ Feature: Bind an application to a service
...
+            metadata:
+                name: sbr-csv-secret-cm-descriptors
+            spec:
+                backingServiceSelector:
+                    group: service.example.com
+                    version: v1
+                    kind: Backserv
+                    resourceRef: demo-backserv-cr-2
+            """
+        Then Secret "sbr-csv-secret-cm-descriptors" must contain data from resource type "secret" cr name "csv-demo-secret" and jsonpath "{.data.username}"
```
pmacik, line 388: @pratikjagrut The acceptance tests are failing most likely a timing issue because you try to check for a secret that is supposed to be created by the binding, but you do so before that happens... You should probably verify anything about the secret only after the binding is finished, which is ensured by the `.status.condition[]` steps... So put the "Secret must contain data" steps after the SBR conditions. You know that after those steps passed, you are 100% sure that the secret is there.

Could you try this https://github.com/pratikjagrut/service-binding-operator/pull/1 ?

test/acceptance/features/bindAppToService.feature
```diff
@@ -217,3 +217,177 @@ Feature: Bind an application to a service
...
+                    group: service.example.com
+                    version: v1
+                    kind: Backserv
+                    resourceRef: demo-backserv-cr-1
+            """
+        Then Secret "sbr-csv-attribute" must contain data from resource type "Backserv" cr name "demo-backserv-cr-1" and jsonpath "{.spec.svcName}"
+        And jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding Request "sbr-csv-attribute" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding Request "sbr-csv-attribute" should be changed to "False"
+
+    Scenario: Create intermidiate secret using information mentioned in x-descriptors of backend service CSV
```
pedjak, line 292: what is the difference between this scenario and the previous one?

In both cases, binding definition is stored in x-descriptors.

I think we should call "binding secret" instead of "intermediate secret".

From user perspective, intermediate secrets are not important, i.e. they are not in focus of the use case. 

What user really wants to know is that a service managed by OLM operator can be bound to an application using binding definitions stored in OLM operator descriptor.

How about rephrasing to:

```
Binding secret gets created from the definitions managed in OLM operator descriptors
```

test/acceptance/features/bindAppToService.feature
```diff
@@ -217,3 +217,177 @@ Feature: Bind an application to a service
...
+            spec:
+                displayName: Some Backend Service
+                install:
+                    strategy: deployment
+                customresourcedefinitions:
+                    owned:
+                        - name: backservs.service.example.com
+                          version: v1
+                          kind: Backserv
+                          statusDescriptors:
```
pedjak, line 308: there is also `specDescriptors` - we should also have a test case covering that  as well.

test/acceptance/features/bindAppToService.feature
```diff
@@ -217,3 +217,177 @@ Feature: Bind an application to a service
...
+                      storage: true
+                scope: Namespaced
+                names:
+                    plural: backservs
+                    singular: backserv
+                    kind: Backserv
+                    shortNames:
+                    - bs
+            """
+        * The CR "demo-backserv-cr-2" for kind "Backserv" is present
```
pedjak, line 344: Do we really need to specify kind in step? How about saying something like:

```
Resource is applied
```

With such step we can apply anything - configmaps, secrets as well.

test/acceptance/features/bindAppToService.feature
```diff
@@ -217,3 +217,177 @@ Feature: Bind an application to a service
...
+            apiVersion: v1
+            kind: Secret
+            metadata:
+                name: csv-demo-secret
+            type: Opaque
+            stringData:
+                username: admin
+                password: secret123
+            """
+        When Service Binding Request is applied to connect the database and the application
```
pedjak, line 375: We should have a more generic step saying that `Service Binding Request is applied`

test/acceptance/features/bindAppToService.feature
pedjak, line 388: We could make it simpler - we do not really care what is the source of  data in binding secret, we just care about its value. So how about

```
Secret X contains Y key with value Z 
```

We have this approach in https://github.com/redhat-developer/service-binding-operator/pull/597/files

pedjak (changes requested):

test/acceptance/features/bindAppToService.feature
pmacik: https://github.com/redhat-developer/service-binding-operator/pull/666 fixes that

test/acceptance/features/bindAppToService.feature
pratikjagrut: SpecDescriptors is covered in the above scenario. https://github.com/redhat-developer/service-binding-operator/pull/620/files#diff-cccd4debaac09aaed0262fe2dbbad9e1R238

test/acceptance/features/bindAppToService.feature
pratikjagrut: If we don't specify kind here then it will be hard to assert if the resource is created or not. https://github.com/redhat-developer/service-binding-operator/pull/620/files#diff-ec891afbdb0cf5fd075cbf5ca535a45fR329

test/acceptance/features/bindAppToService.feature
pratikjagrut: Yes but I think it'll be a good idea to do this in separate PR and we can also refactor some more methods there. Anyway, I was thinking of refactoring acceptance test.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Yes, we can do it, but if you see it is tightly coupled with secrets data only which is base64 encoded whereas configmap and CR spec data will be in the plain string. So the specifying source will apprise about which data need to be encoded in base64 for the assertion. Which is what my step does. 
https://github.com/redhat-developer/service-binding-operator/pull/597/files#diff-ec891afbdb0cf5fd075cbf5ca535a45fR315

test/acceptance/features/bindAppToService.feature
pratikjagrut: > what is the difference between this scenario and the previous one?

The previous scenario is based on issue #656. So as it is a bug it's better to keep a test case for it.


> I think we should call "binding secret" instead of "intermediate secret".

We always have been calling it as an intermediate secret so I just continued it.

 
> 
> How about rephrasing to:
> 
> ```
> Binding secret gets created from the definitions managed in OLM operator descriptors
> ```

The name sounds cool.

test/acceptance/features/bindAppToService.feature
pratikjagrut: We might need to look into better assertion method here then.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Thanx @pmacik. Done

test/acceptance/features/bindAppToService.feature
pedjak: Fine.

test/acceptance/features/bindAppToService.feature
pedjak: ok

test/acceptance/features/bindAppToService.feature
pedjak: You can get kind info from the submitted yaml content.

test/acceptance/features/bindAppToService.feature
pedjak: We should do it here because the used step does not reflect the reality - there is not application selector in the given yaml - so we do not conntect database and application here.

test/acceptance/features/bindAppToService.feature
pedjak: I disagree - user do not care where the value originally comes - it is an implementation detail. From user perspective it is just important that given secret contains a key with some expected value. User do not need to know anything about encoding as well.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Done.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Done.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Done.

test/acceptance/features/bindAppToService.feature
```diff
@@ -305,3 +306,180 @@ Feature: Bind an application to a service
         Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding Request "binding-request-a-d-c" should be changed to "True"
         And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding Request "binding-request-a-d-c" should be changed to "True"
         And Secret "binding-request-a-d-c" contains "SOME_KEY" key with value "SOME_VALUE:5432:db-demo-a-d-c"
+
+
+    Scenario: Creating binding secret from the definitions managed in OLM operator descriptors
+        Given Backend service "some-backend-service.v0.2.0" CSV is installed
```
pedjak, line 312: I would drop service name from the step - it can be obtained from the yaml content.

test/acceptance/features/bindAppToService.feature
```diff
@@ -305,3 +306,180 @@ Feature: Bind an application to a service
...
+                              - binding:env:object:secret:password
+                            - description: Name of the ConfigMap to hold the DB config
+                              displayName: DB Config Map
+                              path: configmap
+                              x-descriptors:
+                              - urn:alm:descriptor:io.kubernetes:ConfigMap
+                              - binding:env:object:configmap:db_host
+                              - binding:env:object:configmap:db_port
+            """
+        * The CRD "backservs.service.example.com" is present
```
pedjak, line 343: same as above.

test/acceptance/features/bindAppToService.feature
```diff
@@ -305,3 +306,180 @@ Feature: Bind an application to a service
...
+                      storage: true
+                scope: Namespaced
+                names:
+                    plural: backservs
+                    singular: backserv
+                    kind: Backserv
+                    shortNames:
+                    - bs
+            """
+        * The CR "demo-backserv-cr-2" is present
```
pedjak, line 363: same as above.

test/acceptance/features/bindAppToService.feature
```diff
@@ -305,3 +306,180 @@ Feature: Bind an application to a service
...
+            """
+            apiVersion: service.example.com/v1
+            kind: Backserv
+            metadata:
+                name: demo-backserv-cr-2
+            status:
+                secret: csv-demo-secret
+                configmap: csv-demo-cm
+            """
+        * The CR "csv-demo-cm" is present
```
pedjak, line 373: to be precise - this is not a CR, it is a config map - so it reads better if we do `ConfigMap is present`

test/acceptance/features/bindAppToService.feature
```diff
@@ -305,3 +306,180 @@ Feature: Bind an application to a service
...
+            """
+            apiVersion: v1
+            kind: ConfigMap
+            metadata:
+                name: csv-demo-cm
+            data:
+                db_host: 172.72.2.0
+                db_port: "3306"
+            """
+        * The CR "csv-demo-secret" is present
```
pedjak, line 383: same above, it reads better like `Secret is present`.

test/acceptance/features/bindAppToService.feature
pratikjagrut: done.

test/acceptance/features/bindAppToService.feature
pratikjagrut: done

test/acceptance/features/bindAppToService.feature
pratikjagrut: done

test/acceptance/features/bindAppToService.feature
pratikjagrut: done

pedjak (approved):
conceptually looks good - squash the commits after tests become green

PR #629, opened 2020-08-20
[acceptance-tests] Use pre-built image of nodejs app instead of building it

pmacik:
### Motivation

<< Why this change is relevant, what's the "story" behind it? >>

### Changes

### Testing

### DIFF
--- a/test/acceptance/features/steps/nodejs_application.py
+++ b/test/acceptance/features/steps/nodejs_application.py
@@ -7,9 +7,40 @@ import time
 
 class NodeJSApp(object):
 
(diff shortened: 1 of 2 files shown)


akashshinde:
@pmacik why not use https://docs.openshift.com/enterprise/3.1/dev_guide/new_app.html#specifying-an-image ?
I think this will create Deployment, Service and Route for us

baijum:
> @pmacik why not use https://docs.openshift.com/enterprise/3.1/dev_guide/new_app.html#specifying-an-image ?
> I think this will create Deployment, Service and Route for us

Do we need Service and Route for tests?
Since `oc` is already a dependency, I don't see any problem to use that either.

Issue #476, opened 2020-05-19
Unable to delete project/namespace due to waiting for SBO's finalizers


pmacik:
After successfully executing the [`nodejs_postgresql`](https://github.com/redhat-developer/service-binding-operator/blob/master/examples/nodejs_postgresql/README.md) scenario I attempted to delete the example's `service-binding-demo` project by executing the `oc delete project service-binding-demo` command.

I expected the project to be deleted after 10s of seconds (while it is supposed to delete the dependent resources (such as `ImageStream`, `BuildConfig`, `Deployment`, `Database`, `ServiceBindingRequest`,...) created during the example. But after ~10 minutes the project is still in the `Terminating` phase.

While inspecting the project I found it waiting for SBO's finalizers:

`Some content in the namespace has finalizers remaining: finalizer.servicebindingrequest.openshift.io in 1 resource instances`

```yaml
apiVersion: project.openshift.io/v1
kind: Project
metadata:
  annotations:
    openshift.io/description: ""
    openshift.io/display-name: ""
    openshift.io/requester: system:admin
    openshift.io/sa.scc.mcs: s0:c24,c14
    openshift.io/sa.scc.supplemental-groups: 1000580000/10000
    openshift.io/sa.scc.uid-range: 1000580000/10000
  creationTimestamp: "2020-05-18T13:08:32Z"
  deletionTimestamp: "2020-05-18T14:39:43Z"
  name: service-binding-demo
  resourceVersion: "60881"
  selfLink: /apis/project.openshift.io/v1/projects/service-binding-demo
  uid: 412a7b80-7971-41ec-be02-5406390afd75
spec:
  finalizers:
  - kubernetes
status:
  conditions:
  - lastTransitionTime: "2020-05-18T14:39:49Z"
    message: All resources successfully discovered
    reason: ResourcesDiscovered
    status: "False"
    type: NamespaceDeletionDiscoveryFailure
  - lastTransitionTime: "2020-05-18T14:39:49Z"
    message: All legacy kube types successfully parsed
    reason: ParsedGroupVersions
    status: "False"
    type: NamespaceDeletionGroupVersionParsingFailure
  - lastTransitionTime: "2020-05-18T14:39:49Z"
    message: All content successfully deleted, may be waiting on finalization
    reason: ContentDeleted
    status: "False"
    type: NamespaceDeletionContentFailure
  - lastTransitionTime: "2020-05-18T14:39:49Z"
[...]

Issue #626, opened 2020-08-19
Remove finalisers in the controller


Avni-Sharma:
Now that https://github.com/redhat-developer/service-binding-operator/pull/596 is merged, we can consider removing finalisers as no cleanup is required on unbind(). The annotations has been removed and ownerRefs added to the intermediate secret which makes the presence of finalisers redundant

cc @isutton

sbose78:
Does this take into account that application objects need to be modified too ?

pedjak:
Obsolete by #919

PR #639, opened 2020-08-26
Remove SBR finalizers in the test namespace before deleting the namespace

pmacik:
### Motivation

Currently the acceptance tests setup deletes the test namespace and creates a new one. But if there are finalizers set on SBRs in that test namespace (for whatever reason) the deletion of the namespace is stuck in Terminating state.

Like in https://github.com/redhat-developer/service-binding-operator/issues/476

### Changes

This PR removes all the finalizers from all the SBRs in the test namespace just before it attempts the namespace deletion.

### Testing

`make e2e-cleanup`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,6 +222,7 @@ e2e-setup: e2e-cleanup e2e-create-namespace e2e-deploy-3rd-party-crds
 
 .PHONY: e2e-cleanup
 e2e-cleanup: get-test-namespace
+	$(Q)-TEST_NAMESPACE=$(TEST_NAMESPACE) $(HACK_DIR)/e2e-cleanup.sh
 	$(Q)-kubectl delete namespace $(TEST_NAMESPACE) --timeout=45s --wait
 
 OBSOLETE_E2E_MESSAGE=WARNING: e2e tests are obsolete and will be removed soon in favour of acceptance tests, \
@@ -437,12 +438,7 @@ deploy-crds:
 .PHONY: deploy-clean
 ## Deploy-Clean: Removing CRDs and CRs
 deploy-clean:
-	$(Q)-kubectl delete -f deploy/crds/apps_v1alpha1_servicebindingrequest_cr.yaml
-	$(Q)-kubectl delete -f deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
-	$(Q)-kubectl delete -f deploy/operator.yaml
-	$(Q)-kubectl delete -f deploy/role_binding.yaml
-	$(Q)-kubectl delete -f deploy/role.yaml
-	$(Q)-kubectl delete -f deploy/service_account.yaml
+	$(Q)-$(HACK_DIR)/deploy-clean.sh
 
 .PHONY: deploy
 ## Deploy:
--- a/hack/deploy-clean.sh
+++ b/hack/deploy-clean.sh
@@ -1,0 +1,14 @@
+#!/bin/bash -x
+
+HACK_DIR=${HACK_DIR:-$(dirname $0)}
+
+# Remove SBR finalizers
+$HACK_DIR/remove-sbr-finalizers.sh
+
+# Delete deployed resources
+kubectl delete -f deploy/crds/apps_v1alpha1_servicebindingrequest_cr.yaml
+kubectl delete -f deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
+kubectl delete -f deploy/operator.yaml
+kubectl delete -f deploy/role_binding.yaml
+kubectl delete -f deploy/role.yaml
+kubectl delete -f deploy/service_account.yaml
--- a/hack/e2e-cleanup.sh
+++ b/hack/e2e-cleanup.sh
@@ -1,0 +1,6 @@
+#!/bin/bash -xe
+
+HACK_DIR=${HACK_DIR:-$(dirname $0)}
+
+# Remove SBR finalizers
+NAMESPACE=$TEST_NAMESPACE $HACK_DIR/remove-sbr-finalizers.sh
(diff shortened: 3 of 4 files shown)


Avni-Sharma:
Should https://github.com/redhat-developer/service-binding-operator/issues/626 be considered in this regard? I think we should give that a look

Makefile
```diff
@@ -222,6 +222,7 @@ e2e-setup: e2e-cleanup e2e-create-namespace e2e-deploy-3rd-party-crds
 
 .PHONY: e2e-cleanup
 e2e-cleanup: get-test-namespace
+	$(Q)-TEST_NAMESPACE=$(TEST_NAMESPACE) $(HACK_DIR)/remove-sbr-finalizers.sh
```
pedjak, line 225: instead of piling up here various cleanup steps, how about that we create a script, e.g. `$(HACK_DIR)/e2e-cleanup.sh` that would contain this logic, and the following step removes the namespace?

hack/remove-sbr-finalizers.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+SBR=$(oc get sbrs)
```
pedjak, line 3: do we really need to use `oc` here or `kubectl` is enough? Asking because at line 6 we use `kubectl`. Let's consolidate.

We could save us one call here. How about the following:

```shell
#!/bin/bash -xe

SBRS=$(oc get sbrs -n $TEST_NAMESPACE -o jsonpath="{.items[*].metadata.name}")
for i in $SBRS; do
     kubectl patch sbr/$i -p '{"metadata":{"finalizers":[]}}' --type=merge;
done
```

mind `-xe` option on bash - `-e` exists the script if any any command returns non-zero exit code and `-x` displays on the console what commands are executed.

pedjak (changes requested):
`deploy-clean` make rule suffers from the same issue, because  SBR CRD cannot be removed if SBR CRs contain finalizers.

pedjak:
this PR tend to fix the issue that impacts local dev workflow. Hence, @DhritiShikhar @Avni-Sharma @akashshinde @baijum @isutton please test it locally. Right now, the following workflow is impacted and does not work on master:

1  start the operator with `make local`
2. follow up any of [examples](examples)
3. stop the operator
4. try to start the operator with `make local`, the process does not move forward because SBR CRD cannot be removed

but it should work on this PR.

Avni-Sharma:
> Should #626 be considered in this regard? I think we should give that a look

cc @pedjak @pmacik

DhritiShikhar:
I tried this PR locally.

```sh
➜  service-binding-operator git:(remove-finalizer) make test-acceptance VERBOSE=3
Requirement already up-to-date: setuptools in ./out/venv3/lib/python3.8/site-packages (50.0.0)
Requirement already up-to-date: pip in ./out/venv3/lib/python3.8/site-packages (20.2.2)
servicebindingrequest.apps.openshift.io/binding-request-1 patched
servicebindingrequest.apps.openshift.io/binding-request-2 patched
servicebindingrequest.apps.openshift.io/binding-request-a-d-s patched
servicebindingrequest.apps.openshift.io/binding-request-a-s-d patched
servicebindingrequest.apps.openshift.io/binding-request-d-a-s patched
servicebindingrequest.apps.openshift.io/binding-request-empty-app patched
servicebindingrequest.apps.openshift.io/binding-request-knative patched
servicebindingrequest.apps.openshift.io/binding-request-missing-app patched (no change)
namespace "test-namespace-6ec92571" deleted

```

SBR finalizers are removed.

hack/e2e-cleanup.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -xe
+
+# Remove SBR finalizers
+SBR=$(kubectl get sbrs)
+if [ $? -eq 0 ]; then
```
pedjak, line 5: no need to `if` check - invoking bash with `-xe` option will immediately exit the shell in case of non zero exit code on any subcommands

see https://github.com/redhat-developer/service-binding-operator/pull/639#discussion_r478371705

pedjak (changes requested):
https://github.com/redhat-developer/service-binding-operator/pull/639#pullrequestreview-476649110 is still not addressed

hack/e2e-cleanup.sh
pmacik: I've missed that... Thx

pmacik:
I've rebased and squashed the commits.

hack/deploy-clean.sh
```diff
@@ -0,0 +1,14 @@
...
+# Remove SBR finalizers
+$HACK_DIR/remove-sbr-finalizers.sh
+
+# Delete deployed resources
+kubectl delete -f deploy/crds/apps_v1alpha1_servicebindingrequest_cr.yaml
+kubectl delete -f deploy/crds/apps.openshift.io_servicebindingrequests_crd.yaml
+kubectl delete -f deploy/operator.yaml
+kubectl delete -f deploy/role_binding.yaml
+kubectl delete -f deploy/role.yaml
+kubectl delete -f deploy/service_account.yaml
```
pedjak, line 14: how about replacing this with:

```
kubectl delete -R -f deploy
```

pedjak (approved):
looks good - with an additional question

hack/deploy-clean.sh
pmacik: The `-R` would attempt to delete all of the yaml files under the `deploy` directory including those that are not Kubernetes resources, at all. Such as `deploy/olm-catalog/csv-config.yaml` and that would always be with error. We could of course move the `olm-catalog` specific directory somewhere else. But that would require additional changes at places that expect those files at that location, and those changes would be un-related to the point of this PR.

Avni-Sharma:
> > Should #626 be considered in this regard? I think we should give that a look
> 
> cc @pedjak @pmacik

Bumping my question up too...icymi :)

hack/deploy-clean.sh
pedjak: alright

pedjak:
/approve

@pmacik please squash the commits before we merge it.

PR #613, opened 2020-08-17
[acceptance-tests] Add scenario to validate etcd binding and use of detectBindingResources

shvasude:
### Motivation

https://issues.redhat.com/browse/APPSVC-644

### Changes

bind etcd cluster instance
- Given Etcd operator running
-  Given Etcd cluster "etcd-cluster-example" is created
-  Given Imported Nodejs application "nodejs-todo-etcd" is running

        
### Testing

make test-acceptance

For further more details refer the CONTRIBUTING.md

PR #644, opened 2020-08-27
[acceptance-tests] move get_deployment_with_secret to openshift class

shvasude:
### Motivation

For now, getting deployment name with intermediate secret is implemented in quarkus application class which would make it to use at quarkus application context level. any other similar usage from different application will have to have these methods inspired in those context. also, this function `get_deployment_with_intermediate_secret_of_given_pattern` is more relevant to openshift class

### Changes

Renamed `get_deployment_with_intermediate_secret` to `get_deployment_with_intermediate_secret_of_given_pattern` giving the application deployment pattern with respective namespace

### Testing

make test-acceptance

### DIFF
--- a/test/acceptance/features/steps/openshift.py
+++ b/test/acceptance/features/steps/openshift.py
@@ -333,3 +333,38 @@ spec:
             return lst
         print('Resource list is empty under namespace - {}'.format(namespace))
         return None
+
+    def get_deployment_names_of_given_pattern(self, deployment_name_pattern, namespace):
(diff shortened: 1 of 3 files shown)


pedjak:
My preference would be to see the immediate usage of it in some new scenario. Hence, I would do this refactoring in PR where we have such need.

pedjak:
included in https://github.com/redhat-developer/service-binding-operator/pull/613

shvasude:
covered this as part of [https://github.com/redhat-developer/service-binding-operator/pull/613](url)

PR #645, opened 2020-08-27
[acceptance-tests] Fix multiservice step to work with both oc v4.5 and v4.6

pmacik:
### Motivation

Currently the `"{app_name}" deployment must contain SBR name "{sbr_name1}" and "{sbr_name2}" step assumes that the output from the `oc get ... -o jsonpath=...` would be in the form of a go structure:

`[map[secretRef:map[name:{binding-request-1}]] map[secretRef:map[name:{binding-request-2}]]]`

which is what the `oc` v4.5 produces. However when `oc` v4.6 is used, the output changes into a valid JSON su:

`[{{"secretRef":{{"name":"{binding-request-1}"}}}},{{"secretRef":{{"name":"{binding-request-2}"}}}}]`

So if one runs the acceptance tests with `oc` v4.6, the scenarios fails on that step, because the expected and actual values of the result do not match even if the operator's behavior is correct.

### Changes

This PR:
* Adds a support for both variants of the result in the `"{app_name}" deployment must contain SBR name "{sbr_name1}" and "{sbr_name2}"` step
* Replaces hard-coded SBR names with the respective step input parameters

### Testing

`make test-acceptance`

### DIFF
--- a/test/acceptance/features/steps/steps.py
+++ b/test/acceptance/features/steps/steps.py
@@ -267,9 +267,13 @@ def then_envFrom_contains(context, app_name, sbr_name1, sbr_name2):
     time.sleep(60)
     openshift = Openshift()
     result = openshift.get_deployment_envFrom_info(app_name, context.namespace.name)
-    expected_result_1 = f'[map[secretRef:map[name:{sbr_name1}]] map[secretRef:map[name:{sbr_name2}]]]'
-    expected_result_2 = f'[{{"secretRef":{{"name":"{sbr_name1}"}}}},{{"secretRef":{{"name":"{sbr_name2}"}}}}]'
-    assert result == expected_result_1 or result == expected_result_2, \
+
+    # Expected result from 'oc' (openshift client) v4.5
+    expected_result_oc_45 = f'[map[secretRef:map[name:{sbr_name1}]] map[secretRef:map[name:{sbr_name2}]]]'
+    # Expected result from 'oc' (openshift client) v4.6+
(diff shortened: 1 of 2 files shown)


test/acceptance/features/steps/steps.py
```diff
@@ -267,8 +267,10 @@ def then_envFrom_contains(context, app_name, sbr_name1, sbr_name2):
     time.sleep(60)
     openshift = Openshift()
     result = openshift.get_deployment_envFrom_info(app_name, context.namespace.name)
-    result | should.be_equal_to("[map[secretRef:map[name:binding-request-1]] map[secretRef:map[name:binding-request-2]]]")\
-        .desc(f'{app_name} deployment should contain secretRef: {sbr_name1} and {sbr_name2}')
+    expected_result_1 = f'[map[secretRef:map[name:{sbr_name1}]] map[secretRef:map[name:{sbr_name2}]]]'
```
pedjak, line 270: for better readability, could you name these vars as `expected_result_ocp_45` and `expected_result_ocp_46` respectively?

pedjak (changes requested):

test/acceptance/features/steps/steps.py
pmacik: Good idea, thx

pmacik: It's just that the difference is in the `oc` tool itself, not the OCP... so I'd rather name it like `expected_result_oc_45`. And add a comment to mitigate any confusion.

Issue #648, opened 2020-08-28
service-binding-operator panics if there is a configMapRef in envFrom


qibobo:
To reproduce this issue.
Create an app with a configMapRef inside its envFrom, then after creating a SBR, the service-binding-operator panics.

```
  envFrom:
      - configMapRef:
          name: seed-defaults
```
The operator log:

```
{"level":"info","ts":1598593249.9040349,"logger":"reconciler","msg":"Saving data on intermediary secret...","Request.Namespace":"default","Request.Name":"binding-request-sample","ServiceBindingRequest.Name":"binding-request-sample"}
E0828 05:40:49.930262       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 1653 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x14e0540, 0x23fbce0)
	/go/src/github.com/redhat-developer/service-binding-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0xa3
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/src/github.com/redhat-developer/service-binding-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x82
panic(0x14e0540, 0x23fbce0)
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest.(*binder).appendEnvFrom(0xc00023a8c0, 0xc000117540, 0x1, 0x1, 0xc000420e00, 0x16, 0xc0002fac00, 0xc000bdcfc8, 0x118a195)
	/go/src/github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/binder.go:366 +0x47
github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest.(*binder).updateContainer(0xc00023a8c0, 0x14d5de0, 0xc00082cae0, 0x0, 0x0, 0x0)
	/go/src/github.com/redhat-developer/service-binding-operator/pkg/controller/servicebindingrequest/binder.go:414 +0xbf
[...]

PR #649, opened 2020-08-28
fix issue #648 , service-binding-operator panics if there is a configMapRef in envFrom

qibobo:
To fix issue #648

### DIFF
--- a/pkg/controller/servicebindingrequest/binder_test.go
+++ b/pkg/controller/servicebindingrequest/binder_test.go
@@ -182,18 +182,20 @@ func TestBinderNew(t *testing.T) {
 
 		require.NotNil(t, binder)
 		secretName := "secret"
+		configMapName := "configmap"
 		d := mocks.DeploymentMock("binder", "binder", map[string]string{})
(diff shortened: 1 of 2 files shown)


pkg/controller/servicebindingrequest/binder_test.go
```diff
@@ -171,6 +171,35 @@ func TestBinderNew(t *testing.T) {
...
+			ConfigMapRef: &corev1.ConfigMapEnvSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: "test-configmap",
+				},
+			},
+		})
+
+		list := binder.appendEnvFrom(envFrom, secretName)
+		require.Equal(t, 2, len(list))
+		require.Equal(t, secretName, list[1].SecretRef.Name)
```
DhritiShikhar, line 197: There are two items in `list` -> `require.Equal(t, 2, len(list))`

We should also have a `require.Equal` check for for `list[0]`

qibobo:
@DhritiShikhar fixed.

PR #659, opened 2020-09-02
Mark e2e testsuite obsolete and disable it before removal.

pmacik:
### Motivation

This is the first step of removing the `e2e` test suite in favour of `acceptance` tests (https://issues.redhat.com/browse/APPSVC-648)

### Changes

This PR disables the `e2e` tests by default unless `ENABLE_OBSOLETE_E2E` is set to `true`.
A warning like the following is shown instead:

```
$ make test-e2e
!WARNING: e2e tests are obsolete and will be removed soon in favour of acceptance tests,
!          so they are disabled by default. If you need to run it anyway, set ENABLE_OBSOLETE_E2E=true in the environment.
make: Nothing to be done for 'test-e2e'.
```

### Testing

`make test-e2e`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -265,6 +265,7 @@ test-unit:
 
 .PHONY: test-e2e-image
 ## Run e2e tests on operator image
+ifeq ($(ENABLE_OBSOLETE_E2E), true)
 test-e2e-image: push-image
 	$(info Running e2e test on operator image: $@)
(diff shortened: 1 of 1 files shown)


pedjak (approved):

pedjak (changes requested):
As CI shows, `e2e-log-parser.sh` file should be also made no-op.

PR #663, opened 2020-09-03
Support air gap scenarios

### LINKED ISSUE #557: Support air gap scenarios
We have recently [updated](https://github.com/application-stacks/runtime-component-operator/pull/146) the Runtime Component Operator to be compliant with Red Hat's requirement for air gap scenarios.

Given that SBO is an essential Operator for many scenarios, we would like to request for it to also comply with the air gap requirements.  

Additional details in 
* https://docs.openshift.com/container-platform/4.3/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-restricted-network_osdk-generating-csvs 
* https://docs.openshift.com/container-platform/4.3/openshift_images/image-configuration.html#images-configuration-registry-mirror_image-configuration

baijum:
### Motivation

Fixes #557

### Changes

After pushing the image to the registry, get the sha256 digest.
Pull it again using the digest to verify.

### Testing

1. Copy Quay token from here: https://quay.io/organization/redhat-developer?tab=robots
2. Set `QUAY_TOKEN` environment variable
3. Set `export OPERATOR_GROUP=redhat-developer`
4. Run `make merge-to-master-release`
5. Run `make push-to-manifest-repo`
6. Check the CSV file under `tmp/manifests` and verify the image location
7. Remove tag from here: https://quay.io/repository/redhat-developer/app-binding-operator?tab=tags

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -493,15 +493,18 @@ merge-to-master-release:
 	$(eval COMMIT_COUNT := $(shell git rev-list --count HEAD))
 	$(Q)docker build -f Dockerfile.rhel -t $(OPERATOR_IMAGE_REF) .
 	docker push "$(OPERATOR_IMAGE_REF)"
+
+.PHONY: verify-image-using-sha256-digest
+## The image pushed to registry should be verified
+verify-image-using-sha256-digest:
 	$(eval AIR_GAP_OPERATOR_IMAGE_REF := $(shell docker inspect --format="{{index .RepoDigests 0}}" $(OPERATOR_IMAGE_REF)))
 	docker rmi "$(AIR_GAP_OPERATOR_IMAGE_REF)"
 	docker rmi "$(OPERATOR_IMAGE_REF)"
 	docker pull "$(AIR_GAP_OPERATOR_IMAGE_REF)"
 
-
 .PHONY: push-to-manifest-repo
 ## Push manifest bundle to service-binding-operator-manifest repo
-push-to-manifest-repo:
+push-to-manifest-repo: verify-image-using-sha256-digest
 	@rm -rf $(MANIFESTS_TMP) || true
 	@mkdir -p ${MANIFESTS_TMP}/${BUNDLE_VERSION}
 	operator-sdk generate csv --csv-version $(BUNDLE_VERSION) --from-version=0.0.23
--- a/Makefile
+++ b/Makefile
@@ -496,11 +496,10 @@ merge-to-master-release:
 
 .PHONY: verify-image-using-sha256-digest
 ## The image pushed to registry should be verified
-verify-image-using-sha256-digest:
-	$(eval AIR_GAP_OPERATOR_IMAGE_REF := $(shell docker inspect --format="{{index .RepoDigests 0}}" $(OPERATOR_IMAGE_REF)))
-	docker rmi "$(AIR_GAP_OPERATOR_IMAGE_REF)"
-	docker rmi "$(OPERATOR_IMAGE_REF)"
-	docker pull "$(AIR_GAP_OPERATOR_IMAGE_REF)"
+verify-image-using-sha256-digest: merge-to-master-release
+	$(eval OPERATOR_IMAGE_DIGEST_REF := $(shell docker inspect --format="{{index .RepoDigests 0}}" $(OPERATOR_IMAGE_REF)))
+	docker rmi "$(OPERATOR_IMAGE_DIGEST_REF)" "$(OPERATOR_IMAGE_REF)"
+	docker pull "$(OPERATOR_IMAGE_DIGEST_REF)"
 
 .PHONY: push-to-manifest-repo
 ## Push manifest bundle to service-binding-operator-manifest repo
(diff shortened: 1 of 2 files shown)


baijum:
@pedjak In today's standup, I mentioned that `docker push` gives the sha256 digest. I found a reference for the same here:
https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier

It reads like this:

> Docker also prints the digest of an image when pushing to a registry. This may be useful if you want to pin to a version of the image you just pushed."

There is `docker push` at the end of `merge-to-master-release` Makefile target.

Makefile
```diff
@@ -480,7 +481,8 @@ push-to-manifest-repo:
 	cp -vrf $(OLM_CATALOG_DIR)/$(GO_PACKAGE_REPO_NAME)/$(BUNDLE_VERSION)/* $(MANIFESTS_TMP)/$(BUNDLE_VERSION)/
 	cp -vrf $(OLM_CATALOG_DIR)/$(GO_PACKAGE_REPO_NAME)/*package.yaml $(MANIFESTS_TMP)/
 	cp -vrf $(CRDS_DIR)/*_crd.yaml $(MANIFESTS_TMP)/${BUNDLE_VERSION}/
-	sed -i -e 's,REPLACE_IMAGE,$(OPERATOR_IMAGE_REF),g' $(MANIFESTS_TMP)/${BUNDLE_VERSION}/*.clusterserviceversion.yaml
+	$(eval AIR_GAP_OPERATOR_IMAGE_REF := $(shell docker inspect --format="{{index .RepoDigests 0}}" $(OPERATOR_IMAGE_REF)))
```
pedjak, line 484: IMHO, there is not need to introduce additional var, we can pack it as

```
sed -i -e 's,REPLACE_IMAGE,$(shell docker inspect --format="{{index .RepoDigests 0}}" $(OPERATOR_IMAGE_REF))),g' $(MANIFESTS_TMP)/${BUNDLE_VERSION}/*.clusterserviceversion.yaml
```

pedjak (changes requested):
It would good if we embed into the travis job a check at the end that would
* remove locally build image, and
* try to pull it from quay using the known digest

In this way, we will be confident even before nightly tests are performed.

In addition, please update PR description to reflect the change. We push images as before, this PR does not change that.

Makefile
baijum: I have moved the variable and reusing it now. See 18c60f5594bc79c1276273b46f41226f2e16ac97

baijum:
I have made the verification of image digest after deleting local image. See: 18c60f5594bc79c1276273b46f41226f2e16ac97

baijum:
I have created a separate target and made it as a dependency on the target to push manifests: 793074196c3f24587ae40a1e7dc6feb036ffe98b

Makefile
```diff
@@ -493,17 +494,24 @@ merge-to-master-release:
 	$(Q)docker build -f Dockerfile.rhel -t $(OPERATOR_IMAGE_REF) .
 	docker push "$(OPERATOR_IMAGE_REF)"
 
+.PHONY: verify-image-using-sha256-digest
+## The image pushed to registry should be verified
+verify-image-using-sha256-digest:
```
pedjak, line 499: this should depend on `merge-to-master-release`. Otherwise, this rule might fail.

Makefile
```diff
@@ -493,17 +494,24 @@ merge-to-master-release:
...
+	$(eval AIR_GAP_OPERATOR_IMAGE_REF := $(shell docker inspect --format="{{index .RepoDigests 0}}" $(OPERATOR_IMAGE_REF)))
+	docker rmi "$(AIR_GAP_OPERATOR_IMAGE_REF)"
+	docker rmi "$(OPERATOR_IMAGE_REF)"
```
pedjak, line 502: these two lines could be combined into single one:

```
docker rmi image1 image2
```

Makefile
```diff
@@ -132,6 +132,7 @@ COMMIT_COUNT := $(shell git rev-list --count HEAD)
 BASE_BUNDLE_VERSION ?= $(OPERATOR_VERSION)
 BUNDLE_VERSION ?= $(BASE_BUNDLE_VERSION)-$(COMMIT_COUNT)
 OPERATOR_IMAGE_REF ?= $(OPERATOR_IMAGE_REL):$(GIT_COMMIT_ID)
+AIR_GAP_OPERATOR_IMAGE_REF ?= ""
```
pedjak, line 135: This var keeps image digest ref. Hence, how about to call it `OPERATOR_IMAGE_DIGEST_REF`? `AIR_GAP_OPERATOR_IMAGE_REF` is too high level abstraction here, it describes a use case. Image digest refs are used both in air gap and regular case. Its usage just enable image caching in a registry accessible behind firewall.

pedjak (changes requested):
If I may suggest according to the change in PR: instead of

```
After pushing the image to the registry, get the sha256 digest.
Pull it again using the digest to verify.
```

the essential change is

```
published CSV refers operator image via sha256 digest ref.
```

Makefile
baijum: Done. a4617d0de4c0f7b730b1054e06d9d384d8a4be11

Makefile
baijum: Done. a4617d0de4c0f7b730b1054e06d9d384d8a4be11

Makefile
baijum: Done. a4617d0de4c0f7b730b1054e06d9d384d8a4be11

baijum:
/hold

This PR is going to affects the release process.

pedjak:
Fixed in operator-framework/community-operators#2796 - the published CSV refers the operator image using sha.

PR #671, opened 2020-09-09
e2e-cleanup make rule should not fail if testname does not exist

pedjak:
* removal of namespace moved to hack/e2e-cleanup.sh
* if namespace does not exit, do not try to remove it


pmacik:
@pedjak There's another part of the e2e-cleanup that is potentially problematic, I fixed that and created a sub-PR https://github.com/pedjak/service-binding-operator/pull/1 so you could include it in this, since it is related.

Could you look and rebase&merge that if you are fine with that?

pmacik (changes requested):
https://github.com/pedjak/service-binding-operator/pull/1

pedjak:
> There's another part of the e2e-cleanup that is potentially problematic,

Thanks, I have added it and polished things a bit further.

PR #674, opened 2020-09-10
Proposal to add an alternate service binding implementation

navidsh:
### Motivation

Proposing a mechanism to implement the service binding spec while keeping the current mechanism for binding services already implemented in the service binding operator.

### Changes

A new enhancement proposal to implement Service Binding Specification in the Service Binding Operator.

### Testing

N/A


docs/proposals/service-binding-specification/implementing-specification.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Proposal
+
+There are major differences between the specification and the current mechanism the Service Binding
+Operator employes such as how the application is defined in the CRD, how a binding secret is 
+projected in an application container and etc. The [Service Binding feature comparison document](./service-binding-feature-comparison.pdf) lists the differences between the two.
+
+To avoid tangling the implementation between the current codebase and the codebase to support the
+specification, this proposal suggests creating two different channels for the Service Binding
+Operator (e.g. `current` and `spec`) to avoid mixing of implementation details between the two
```
sbose78, line 79: Note, OLM is just a mode of packaging and delivering the controller+CRD.Could you help us with an enhancement proposal for the following:  

In an OLM agnostic way,
* How would an existing user of Service Binding migrate their (common) use cases from the existing CRD to the new CRD(s) ?
* How would the project transition its code/structure.

sbose78 (changes requested):
The current proposal effectively means delivering distinct controllers/CRDs using channels. That's not what my request for a proposal was :) 

We can't be maintaining distinct controllers/CRDs in one repo and delivering through different channels.

Check code below.

docs/proposals/service-binding-specification/implementing-specification.md
arthurdm: thanks for the feedback @sbose78 - those are fair points.  Here are my thoughts:

*  for the purposes of installation, we can probably mirror the OLM channel concept into a non-OLM `install.yaml` (Deployment/Role/RoleBinding/ServiceAccount/CRD) inside somewhere like `deploy/spec-impl`.  That way we're covered in both OLM and non-OLM scenarios.
* the biggest challenge is code co-existence and migration. @navidsh and @arturdzm outlined a nice [comparison](https://github.com/navidsh/service-binding-operator/blob/spec-proposal/docs/proposals/service-binding-specification/feature-comparison.pdf) of features between SBO's `ServiceBindingRequest` (SBR) and the spec's `ServiceBinding` (SB).
  * to maintain maximum stability of the current SBO codebase we could start a new controller and start pulling code from SBO as appropriate - for example, to handle a service provider's annotations, etc.  During this time SBO would have a controller for each of SBR and SB.
  * the next step is for the SB controller to automatically create  (1 or more)`ServiceBinding` CR(s) from a `ServiceBindingRequest` CR, so that customers can still use SBR for a while.  
  * after a certain period of time, SBR is deprecated, and eventually removed.

isutton: @arthurdm @navidsh

I am curious whether we have a comparison between one-to-one and one-to-many approaches and how the one-to-one approach is expected to work according to your architecture, do you have anything at hand?

arthurdm: hey @isutton - we don't have a specific document that clearly describes the 1..1 versus 1..N relationshipps, but that's probably something we should doc in the spec, since we spent quite a bit of time discussing this topic there.  In my mind it boils down to two points which I outlined [here](https://github.com/redhat-developer/service-binding-operator/pull/608#discussion_r482276288).

sbose78:
Hi @navidsh @arthurdm ,

Sorry if I wasn't clear on a couple of things with respect to the enhancement proposal approach:

This comparison is a good start
https://github.com/navidsh/service-binding-operator/blob/spec-proposal/docs/proposals/service-binding-specification/feature-comparison.pdf

I'm not too eager to have a second CRD and a controller in parallel ( renamed this proposal to reflect that) :) That feels like two different projects squeezed into one 🤷‍♂️ . 

However, a good contextual discussion would need **one** enhancement proposal **per** feature/area in the above doc that we wish to modify.

docs/proposals/service-binding-specification/implementing-specification.md
```diff
@@ -0,0 +1,111 @@
+---
+title: Implementing Service Binding Specification for Kubernetes
```
sbose78, line 2: Given that this doesn't describe implementation, rather describes how to deliver two sets of binding controller/CRDs, could you update the PR accordingly ?

( I've updated the title ..)

arthurdm:
hi @sbose78 - I think this point may be the key to focus in:

> I'm not too eager to have a second CRD and a controller in parallel 

As you know, the Service Binding Specification introduces the `ServiceBinding` CRD, with vendor portability being the main driver.  So it would be impossible to be fully be spec compliant without supporting this CRD.  =)

What we're proposing is a way to have the two CRDs co-exist in parallel during a period of migration, with the goal to eventually have just the `ServiceBinding` CRD.  

Do you have a suggestion for an alternative way to become spec compliant (not just spec equivalent)?

docs/proposals/service-binding-specification/implementing-specification.md
arthurdm: I think the old title correctly identified what we're trying to do:  implement the spec.  The two sets of controllers / CRDs is a temporary state until current SBO users can be migrated to the spec implementation.

sbose78:
We should implement new changes the same way we've been implementing so far :

Update the existing API with reasonable amount of backward compatibility where applicable. As an example, the API supported `backingServiceSelectors` and ` backingServiceSelector` for more than 6 months.

As an example, the enhancement proposal needs to make a case about why we should move to a singular `Service` v/s multiple services.

arthurdm:
> the enhancement proposal needs to make a case about why we should move to a singular Service v/s multiple services

That is a good example of a proposal that was made to the **specification** a few months ago, with corresponding discussion in various recorded hangouts, etc.  

We're trying to help SBO become spec compliant, which is different than going feature by feature and seeing which ones apply to SBO.  

For example:  an Ingress Controller doesn't pick and choose which parts of the core Ingress CRD it will implement - if the implementation has concerns, those are raised back into the specification community.  

My suggestion would be:  if there are concerns with the specification from SBO's point of view, let's definitely open those [here](https://github.com/k8s-service-bindings/spec/issues) and discuss at the next interlock.  It's a fully opened spec community, so you can directly add your item into the [agenda](https://docs.google.com/document/d/1rR0qLpsjU38nRXxeich7F5QUy73RHJ90hnZiFIQ-JJ8/).  The community would be happy to discuss any feedback, and that can help shape future changes.     👍

sbose78:
> That is a good example of a proposal that was made to the specification a few months ago, with corresponding discussion in various recorded hangouts, etc.

Let's have the arguments for the change recorded in an enhancement proposal.

arthurdm:
thanks for adding that @navidsh!

docs/proposals/service-binding-specification/implementing-specification.md
sbose78: The spec is not final. Parts of the spec (RC1) was being implemented when RC2 happened from June which is extremely distinct from RC1. 

If implementing the spec is the goal, I would say, let's wait for next year when it's stable and we've got enough feedback.

sbose78:
Hi, 
Could we please have distinct enhancement proposals so that we can have contextual discussions ?

I can easily see these broad areas:

1. Handling of projecting of information into applications as envs/volume mounts
2. Handling of multiple service resources ( not necessarily services )

Each of these changes have far-reaching effects on the use cases exposed by the projects and they should be considered, discussed and documented.

sbose78:
> if there are concerns with the specification from SBO's point of view, let's definitely open those here and discuss at the next interlock. It's a fully opened spec community, so you can directly add your item into the agenda. The community would be happy to discuss any feedback, and that can help shape future changes. 👍

My feedback is that a non-core-kube CRD shouldn't be part of the specification because the API maturity is very implementation specific in case of CRDs. 

Only **backing service**/subject and **application**/consumer contracts should be in the specification. 

I've conveyed the feedback to the spec group multiple times. This is a fundamental difference in opinion I have and I do not find the spec leads keen on making any changes to that. This is why I've given up on the spec conversation  :)

> an Ingress Controller doesn't pick and choose

Ingress is a core API, it's always there and hence custom controllers "must" be able to read it, we can't compare the ServiceBinding CRD with the Ingress API. Also, there are good technical/platform reasons why the Ingress controllers have their own implementations. Service Binding doesn't have that. We should worry about thousands of backing services out there instead of 2-3 implementations. 

If you wish to propose changes to the current API/CRD in the Service Binding Operator project, my request would be to make multiple enhancement proposals which would take into account :

* what the current API does 
* which use cases get affected by a modification to the API.

Let's not enforce a yet-to-be-final spec which in my experience has already changed significantly from RC1 to RC2.

arthurdm:
>  This is why I've given up on the spec conversation :)

It's definitely valid, and healthy, for opinions to differ. That's what makes open source communities great - it's an aggregation of the group's consensus.  

My main concern would be if the SBO community is blocked from moving forward in implementing the specification because of a single opinion.  A lot of SBO members have joined, and continue to join, the specification interlocks and we have not heard any concerns raised about the `ServiceBinding` CRD.  We're trying to foster a specification community so that decisions are made as a collective group, always seeking feedback as the first step in our hangouts.

We're quickly closing in 1.0.0-RC3, mainly for documentation purposes, so the spec becoming final is nearing very soon.  We could defer the implementation conversation until then, but that means SBO misses out in the early adoption process, which [at one point](https://github.com/operator-framework/enhancements/pull/12#issuecomment-610431986) was viewed as a good thing.

docs/proposals/service-binding-specification/implementing-specification.md
```diff
@@ -0,0 +1,129 @@
...
+### Goals
+
+Goals for this proposal are listed as follows:
+
+- Allow users of the Service Binding Operator to have a clear separation in choosing between the
+specification and the current mechanism for binding applications to services.
+
+### Non-Goals
+
+- Implementation details about how the specification should be implemented.
```
sbose78, line 91: Some details would be nice, especially how the existing implementation maps to the one proposed. https://github.com/redhat-developer/service-binding-operator/pull/674/files#diff-07b3f77fd05f37abf09d5a741a039c4b 

I think this is a good start
https://github.com/redhat-developer/service-binding-operator/pull/674/files#diff-07b3f77fd05f37abf09d5a741a039c4b

docs/proposals/service-binding-specification/implementing-specification.md
```diff
@@ -0,0 +1,129 @@
...
+
+- It is easier for tools created around the service binding to add and remove services by simply
+  adding or removing independent CRs instead of having to deal with editing an array of services
+  within a CR. In addition, editing the service array in a CR might cause CR to be corrupted unless
+  concurrent updates are well synchronized across a tool or tools. However, by defining a single
+  service per CR, adding or removing services simply becomes `kubectl apply ...` and
+  `kubectl delete ...`.
+
+- Defining one service per CR allows a cleaner association of attributes with the service. For example,\
+  when we identify a set of extra mappings or environment variables in a `ServiceBinding`, which
```
sbose78, line 73: Thank you for adding details.

> For example,\
  when we identify a set of extra mappings or environment variables in a `ServiceBinding`, which
  service does that get associated with if we have multiple services? Or if we want to change the
  service name or type that gets mounted, we again run into the issue that they all require a way
  to identify the service you want to apply those attributes against.

The existing implementation doesn't have this issue.

(1) you can unique identify keys coming from the a specific service by using service specific prefixes.
(2) you can use custom env vars where you can refer to the group.version.kind.name.service_specific_key

docs/proposals/service-binding-specification/implementing-specification.md
```diff
@@ -0,0 +1,129 @@
...
+  service does that get associated with if we have multiple services? Or if we want to change the
+  service name or type that gets mounted, we again run into the issue that they all require a way
+  to identify the service you want to apply those attributes against.
+
+Therefore, allowing users of the Service Binding Operator to choose between the specification and the
```
sbose78, line 78: I don't think users would be encouraged to use a service binding implementation which gets tainted to be not following a specification.

docs/proposals/service-binding-specification/implementing-specification.md
```diff
@@ -0,0 +1,129 @@
...
+current mechanism is likely to increase the user base of the operator, having a more vibrant community
+of service providers.
+
+### Goals
+
+Goals for this proposal are listed as follows:
+
+- Allow users of the Service Binding Operator to have a clear separation in choosing between the
```
sbose78, line 86: Same comment as above : 

I don't think users would be encouraged to use a service binding implementation which gets tainted to be not following a specification. Would you agree? :)

docs/proposals/service-binding-specification/implementing-specification.md
sbose78: Retitled to, "Proposal to add an alternate service binding implementation" because the content does say that the proposed implementation stays along with the existing one. 

Does this reflect the intent better? Also, I could also revert it to the original.

sbose78:
> My main concern would be if the SBO community is blocked from moving forward in implementing the specification because of a single opinion. A lot of SBO members have joined, and continue to join, the specification interlocks and we have not heard any concerns raised about the ServiceBinding CRD. We're trying to foster a specification community so that decisions are made as a collective group, always seeking feedback as the first step in our hangouts.

I'm not sure how many dissenting opinions are needed given that one is not enough for the spec community.

> but that means SBO misses out in the early adoption process

Let's be clear, the current state of the spec needs a fairly different implementation. There are two ways to get to it : 
1) Migrate area by area 
2) Implement the alternate implementation from the scratch.

Your are proposing (2), I am proposing (1).

docs/proposals/service-binding-specification/implementing-specification.md
arthurdm: I think there are 2 important points where the spec differs over a multi-service implementation:

(1) the `ServiceBinding` extra mappings become more readable because you keys are referenced simply by using `key` instead of `group.version.kind.name.service_specific_key`
(2) the `ServiceBinding` extra mappings are volume mounted with the same single Secret of the CR.  In the case of multiple services with `ServiceBindingRequest`, where do the extra mappings get volume mounted under?  (in the scenario where volume mount is picked instead of env vars)

docs/proposals/service-binding-specification/implementing-specification.md
arthurdm: I do agree. :)  hence our desire to get to spec compliance. 👍        If there are existing customers in prod they can have a nice stable period to make the necessary adjustments.

arthurdm:
I believe @navidsh's proposal stays clears of choosing between (1) and (2), as it states:

```
Providing implementation details is not the goal of this proposal
```

I actually believe we should be able to re-use a lot from the current SBO codebase - specially the annotation / x-descriptor support.  Having a second controller doesn't mean we can't share most of the underlying implementation.

docs/proposals/service-binding-specification/implementing-specification.md
pedjak: IMHO, a better title could be "Support Service Binding Specification for Kubernetes"

docs/proposals/service-binding-specification/implementing-specification.md
sbose78: >  the ServiceBinding extra mappings become more readable because you keys are referenced simply by using key instead of group.version.kind.name.service_specific_key

You could alias it as "x" or even use the service id "x" to do a x.username , y.password . We currently support 
* Aliases in the golang template 
* service id https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/apis/operators/v1alpha1/servicebinding_types.go#L59
* the full g.v.k.name.something 

> where do the extra mappings get volume mounted under

There are two levels of prefixes for both env var and vol paths: 
* Service specific, or 
* Global. In this case, this would be a service specific volume mount path prefix. This isn't implemented yet on a service level, but should be do-able given the direction things are going in.

docs/proposals/service-binding-specification/implementing-specification.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Proposal
+
+To avoid tangling the implementation between the current codebase and the codebase to support the
+specification, this proposal suggests creating two different channels for the Service Binding
+Operator (e.g. `current` and `spec`) to avoid mixing of implementation details between the two
+codebases. This way users can freely choose what set of resources they expect the operator to act on at
+the operator install time.
+
+### User Stories
```
sbose78, line 101: Hi @navidsh, 
Could you please provide additional user stories here to depict the usage of the new API/implementation ?

docs/proposals/service-binding-specification/implementing-specification.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### Story 1
+
+When installing the Service Binding Operator, a user can choose which of the two channels
+to install the Service Binding Operator from. Depending on the channel, the operator installs different
+sets of Custom Resource Definitions (CRDs) and would watch for different sets of resources in the
+cluster.
+
+### Implementation Details/Notes/Constraints
```
sbose78, line 110: As constraints/notes could we outline 
* Which use cases will be supported as is ( but with a different API ) and 
* Which ones the new implementation isn't planning on supporting.

This [doc](https://github.com/redhat-developer/service-binding-operator/blob/d82c45ffbe6cf982fb3bc5c35f0be7e6f241066a/docs/proposals/service-binding-specification/feature-comparison.pdf) this does a great job of API comparison, though.

Some questions I would have is:

1. How do we handle custom binding information ( known as customEnvVar here) which is constructed from the content in the spec/status and/or referenced secret/configmap ? ( no annotations/descriptors ) 

2. Would the new implementation have a way to auto-detect binding resources ( `detectBindingResources` ) ?

3. Multiple service resource use case: If a connection string has to be constructed from two resources namely, "DBInstance" and "DBCredentials" , how does a user do that?

cdlliuy:
Hi,  @sbose78   @navidsh  I am jumping in this thread a little bit more late ...  
Currently, we are using the current service binding operator features to build our service binding related offering by leveraging the features of multiple service binding,   customer env prefix,  envPrefix, id and etc ...

 The [doc](https://github.com/redhat-developer/service-binding-operator/blob/d82c45ffbe6cf982fb3bc5c35f0be7e6f241066a/docs/proposals/service-binding-specification/feature-comparison.pdf) is great,  but I still need more details to understand how the service binding spec can handle our use case below described in current SBR CR ..

```  
spec:
    applicationSelector:
      group: serving.knative.dev
      resource: services
      resourceRef: yingtestapp
      version: v1
    backingServiceSelectors:
    - envVarPrefix: SERVICE1
      group: fake-group
      id: fakeid1
      kind: fake-service
      resourceRef: fake-service-instance-1
      version: v1
    - envVarPrefix: SERVICE2
      group: fake-group
      id: fakeid2
      kind: fake-service
      resourceRef: fake-service-instance-2
      version: v1
    customEnvVar:
    - name: myENV1
      value: '{"credentials":{{json .fakeid1.status.secretName}}}'
```

Can you explain further based on the example above?

sbose78:
@navidsh Could you also help understand how https://github.com/strimzi/strimzi-kafka-operator/pull/2753 would work if the spec were to be used ?

docs/proposals/service-binding-specification/implementing-specification.md
arthurdm: hey @sbose78 - here are some thoughts on the questions:
1. The spec allows for the creation of custom mappings using hardcoded Strings + values available from the `binding Secret`.  So the scope of available data for mapping is the information the provisioned service has made available - not its spec/status.  An example [here](https://github.com/k8s-service-bindings/spec/blob/master/README.md#mappings-example-resource).   This custom mapping is then added to your available binding Secret.   For security reasons we volume mount all of the binding Secret's keys, but you can choose which ones you want to make available as env vars, sample [here](https://github.com/k8s-service-bindings/spec/blob/master/README.md#environment-variables-example-resource).

2.  This capability has been discussed in the spec and deemed that can be a useful value-add of implementations outside of the spec.  

3.  A user would have one `ServiceBinding` per binding resource, and could then choose how to concatenate the two inside the app - the choices would be via volume mount or env var.   Although I would add that this scenario hints at an opportunity for improving the provisioned service:  the information needed to connect should be available in a single resource (e.g. the `DBCredentials` should contain the information about which DB URL the credentials apply, hence not needing a separate `DBInstance` resource for this scenario).

sbose78: Note, everything in Service Binding will be supported with volume mounts and env vars, mostly in a mutually exclusive way.

> The spec allows for the creation of custom mappings using hardcoded Strings + values available from the binding Secret

Note, this pattern is supported in a resource-agnostic way : 

```
kind: ServiceBinding
metadata:
  name: binding-request
spec:
  application:
    name: java-app
    group: apps
    version: v1
    resource: deployments
  services:
  - group: ""
    version: v1
    kind: Secret
    name: any-secret
    id: postgresDB
  customEnvVar:
    - name: accountServiceUri
      value: ':{{ any-secret.data.key1:any-secret.data.key2 }}'
```
Are you proposing we drop support for spec/status/data-based introspection, or have both co-exist in the API?


> A user would have one ServiceBinding per binding resource, and could then choose how to concatenate the two inside the app 

That implies that the user has to
* Create 2 `ServiceBindings`
* Modify application code to stitch the information from individual secrets


> Although I would add that this scenario hints at an opportunity for improving the provisioned service: the information needed to connect should be available in a single resource (e.g. the DBCredentials should contain the information about which DB URL the credentials apply, hence not needing a separate DBInstance resource for this scenario)

This implies, that the service provider would have to make changes to the controller code, even though separating instance creation and credential creation is often a conscious design choice.

Did I understand you correctly ? :)

arthurdm: > Are you proposing we drop support for spec/status/data-based introspection, or have both co-exist in the API?

I think this is a case where the functionality of SBO is a superset of what the spec is proposing, so we should be able to happily comply with the spec while keeping the value-add functionality.

> even though separating instance creation and credential creation is often a conscious design choice.

I would assert that this provider design choice forces consumers to deal with two CRs, so it's acceptable that the app would have these as separate values.  In many cases the app won't even be concatenating these values together into a single string, they would use them in separate places - e.g. adding the credentials in headers, and URL in the REST request.  I think this goes down the road of multiple vs single provisioned service references =)

arthurdm:
regarding strimzi - the current proposal could still work as-is, since the spec also supports x-descriptors / annotations.  I think there may be room for improvement in terms of complying with the duck type that the spec proposes, for a cleaner approach.

docs/proposals/service-binding-specification/implementing-specification.md
sbose78: > I would assert that this provider design choice forces consumers to deal with two CRs, so it's acceptable that the app would have these as separate values. 

A conscious design choice by the provider doesn't necessarily mean the complexity has to be passed on to the app developer. The current ServiceBinding API implementation shows that consumer workloads *don't* have to deal with two different `ServiceBindings` if designed well.

One can make do with one `ServiceBinding` CR and construct amazing custom variables that could be used by both env vars & volume mounts.

arthurdm: yup, that's true.  Although what I have seen is that the advantages of having separate ServiceBinding CRs outweigh the advantages of having single multi-service ServiceBinding CR (summarized [here](https://github.com/redhat-developer/service-binding-operator/pull/608#discussion_r482276288)).

sbose78: > it's easier for a tool to add and remove services by simply adding or removing independent CRs, instead of having to deal with editing (and potentially corrupting, if concurrent updates are not well synchronized) an array inside a CR. It simply becomes "kubectl apply .." and "kubectl delete .."

You can run into issues with concurrent updates to any CR's spec if the status is constantly changing - try updating a _just created_ `Deployment`. Also, supporting multiple services doesn't force any tool or anyone to also use multiple services - it can still refer to one service if it wants to ;)

> cleaner association of attributes with the service. For example, when we identify a set of extra mappings or environment variables in a ServiceBinding, which service does that get associated with if we have multiple services? Or if we want to change the service name or type that gets mounted, we again run into the issue that they all require a way to identify the service you want to apply those attributes against.

This is a solved problem - we already support this today in a very clean way using the group.version.kind.name notation or by simply using the service "id". You can use the "id" (or an equivalent identifier) in the mount path if you wish to.

arthurdm: I think a big issue with using `id` to distinguish multiple services within a single CR is that every single attribute of your CR which impacts the projection now needs to be "id aware".  So it's the mappings, the env vars, the name of the projection, the status, etc.  Actually, the status of the CR is probably a whole separate topic of complexity, since now there would be a partial success cases where maybe some of the services were not yet ready, etc. 

In the spec we had a few hangouts where we discussed this, and arrived at the conclusion that a single service CR was also much simpler to work with, even when adding mappings and env vars: [sample](https://github.com/k8s-service-bindings/spec#environment-variables-example-resource)

sbose78: > I think a big issue with using id to distinguish multiple services within a single CR is that every single attribute of your CR which impacts the projection now needs to be "id aware". So it's the mappings, the env vars, the name of the projection, the status, etc. 

If a workload talks to multiple service resources, you have to _namespace_ things on the workload side  **anyway** :) Also, our code already has binding secret generation strategies that support this :) 

> a single service CR was also much simpler to work with

* Was it hard to implement ? Yes, there's work.
* Is it easy to use as a user ? Yes :) ( less code changes on the user's app, multiple Service Binding CRs **not** needed, will work with *any* backing service )

> the status of the CR is probably a whole separate topic of complexity, since now there would be a partial success cases where maybe some of the services were not yet ready, etc.

There's prior art we can take inspiration from https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate , again, it is a bit of work for the implementor, but do-able in a clean way to help the user.

docs/proposals/service-binding-specification/implementing-specification.md
navidsh: The intention of this proposal is really to suggest a way to support the spec. I think "Support Service Binding Specification for Kubernetes" summarizes the intention nicely.

sbose78: Sure sure!

docs/proposals/service-binding-specification/implementing-specification.md
navidsh: hi @sbose78, sure thing. Adding this now...

navidsh: Sorry for the delay... Addressed in https://github.com/redhat-developer/service-binding-operator/pull/674/commits/d82c45ffbe6cf982fb3bc5c35f0be7e6f241066a.

navidsh:
> Any idea to support our use case in the future service binding spec design?

Hi @cdlliuy, the [Service Binding Specification for Kubernetes](https://github.com/k8s-service-bindings/spec) doesn't support one-to-one binding. Here is some info on why this decision was made: https://github.com/redhat-developer/service-binding-operator/pull/674#discussion_r488580263.

To work around your use case, you can create multiple `ServiceBinding` CRs and concatenate info from multiple sources either in your applications or by adding a `postStart` hook to your application to concatenate various env variables into json.

sbose78:
Update: As a project we are looking to have a more first class support for the spec. There are separate issues being tracked by @jasperchui in this repo to get this moving. We are also working on collaborating on a single implementation with vmware.

As much as possible, we are trying to avoid different APIs. Therefore, I'm going to close this since this is no longer being tracked :)

PR #676, opened 2020-09-14
[acceptance-tests] Fix python linting dependencies.

pmacik:
### Motivation

Recently the python linter started to require `flake8_polyfill` module which is not installed. 

### Changes

This PR adds `flake8_polyfill` module into the installed dependencies.

### Testing

`make lint-python-code`

### DIFF
--- a/hack/check-python/prepare-env.sh
+++ b/hack/check-python/prepare-env.sh
@@ -9,5 +9,5 @@ function prepare_venv() {
     for req in $(find . -name 'requirements.txt'); do
         python3 "$(which pip3)" install -q -r $req;
     done
-    python3 "$(which pip3)" install -q pydocstyle pyflakes vulture radon
+    python3 "$(which pip3)" install -q -r $(dirname $0)/requirements.txt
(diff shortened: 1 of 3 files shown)


hack/check-python/prepare-env.sh
```diff
@@ -9,5 +9,5 @@ function prepare_venv() {
     for req in $(find . -name 'requirements.txt'); do
         python3 "$(which pip3)" install -q -r $req;
     done
-    python3 "$(which pip3)" install -q pydocstyle pyflakes vulture radon
+    python3 "$(which pip3)" install -q pydocstyle pyflakes vulture radon flake8_polyfill
```
pedjak, line 12: In order to avoid similar surprise in the future, we should pin the dependency versions. I would propose that we extract the dependencies into `requirement.txt` and pin the versions. We should apply the same to `test/acceptance/features/requirements.txt` as well.

pedjak (approved):
please update PR description and summary and we are good to go

PR #677, opened 2020-09-14
[acceptance-tests] Update README.md with description and tutorial for acceptance tests

pmacik:
### Motivation

Currently there is effectively no docs nor tutorials available for the acceptance tests.

### Changes

This PR adds a description and tutorials for the acceptance tests.

### Testing

Read and follow the instructions [`./tests/acceptance/README.md`](https://github.com/pmacik/service-binding-operator/blob/acceptance-tests-readme/test/acceptance/README.md)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,8 @@ LOGS_DIR ?= $(OUTPUT_DIR)/logs
 GOLANGCI_LINT_BIN=$(OUTPUT_DIR)/golangci-lint
 PYTHON_VENV_DIR=$(OUTPUT_DIR)/venv3
 
+CONTAINER_RUNTIME ?= docker
+
 # -- Variables for uploading code coverage reports to Codecov.io --
 # This default path is set by the OpenShift CI
 CODECOV_TOKEN_PATH ?= "/usr/local/redhat-developer-service-binding-operator-codecov-token/token"
@@ -161,6 +163,7 @@ PULL_NUMBER := $(shell echo $$CLONEREFS_OPTIONS | jq '.refs[0].pulls[0].number')
 # -- Variables for acceptance tests
 TEST_ACCEPTANCE_START_SBO ?= local
 TEST_ACCEPTANCE_OUTPUT_DIR ?= $(OUTPUT_DIR)/acceptance-tests
+TEST_ACCEPTANCE_REPORT_DIR ?= $(OUTPUT_DIR)/acceptance-tests-report
 TEST_ACCEPTANCE_ARTIFACTS ?= /tmp/artifacts
 
 TEST_ACCEPTANCE_TAGS ?=
@@ -294,6 +297,16 @@ endif
 test-acceptance-smoke:
 	$(Q)TEST_ACCEPTANCE_TAGS=@smoke $(MAKE) test-acceptance
 
+.PHONY: test-acceptance-generate-report
+## Generate acceptance tests report
+test-acceptance-generate-report:
+	$(Q)CONTAINER_RUNTIME=$(CONTAINER_RUNTIME) $(HACK_DIR)/allure-report.sh generate
+
+.PHONY: test-acceptance-serve-report
+## Serves acceptance tests report at http://localhost:8088
+test-acceptance-serve-report:
+	$(Q)CONTAINER_RUNTIME=$(CONTAINER_RUNTIME) $(HACK_DIR)/allure-report.sh serve
+
 .PHONY: test-acceptance-artifacts
 ## Collect artifacts from acceptance tests to be archived in CI
 test-acceptance-artifacts:
@@ -364,9 +377,9 @@ push-operator: prepare-csv
 
 ## Push-Image: push container image to upstream, including latest tag.
 push-image: build-image
-	podman tag "$(OPERATOR_IMAGE):$(OPERATOR_TAG_LONG)" "$(OPERATOR_IMAGE):latest"
-	-podman push "$(OPERATOR_IMAGE):$(OPERATOR_TAG_LONG)"
-	-podman push "$(OPERATOR_IMAGE):latest"
+	$(CONTAINER_RUNTIME) tag "$(OPERATOR_IMAGE):$(OPERATOR_TAG_LONG)" "$(OPERATOR_IMAGE):latest"
+	-$(CONTAINER_RUNTIME) push "$(OPERATOR_IMAGE):$(OPERATOR_TAG_LONG)"
+	-$(CONTAINER_RUNTIME) push "$(OPERATOR_IMAGE):latest"
 
 ## -- Local deployment targets --
 
@@ -441,10 +454,10 @@ endif
 .PHONY: merge-to-master-release
 ## Make a dev release on every merge to master
 merge-to-master-release:
-	echo "${QUAY_TOKEN}" | docker login -u "redhat-developer+travis" --password-stdin quay.io
+	echo "${QUAY_TOKEN}" | $(CONTAINER_RUNTIME) login -u "redhat-developer+travis" --password-stdin quay.io
 	$(eval COMMIT_COUNT := $(shell git rev-list --count HEAD))
-	$(Q)docker build -f Dockerfile.rhel -t $(OPERATOR_IMAGE_REF) .
-	docker push "$(OPERATOR_IMAGE_REF)"
+	$(Q)$(CONTAINER_RUNTIME) build -f $(CONTAINER_RUNTIME)file.rhel -t $(OPERATOR_IMAGE_REF) .
+	$(CONTAINER_RUNTIME) push "$(OPERATOR_IMAGE_REF)"
 
 
 .PHONY: push-to-manifest-repo
--- a/hack/allure-report.sh
+++ b/hack/allure-report.sh
@@ -1,0 +1,29 @@
+#!/bin/bash -ex
+
+CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-docker}
+
+$CONTAINER_RUNTIME info 2>/dev/null >/dev/null || NO_CONTAINER_RUNTIME=$?
+if [ -n "$NO_CONTAINER_RUNTIME" ]; then
+    echo "WARNING $CONTAINER_RUNTIME is required but not installed, skipping..."
+    exit 1
+fi
+
+ACTION=${1:-generate}
+
+NAME=test-acceptance-report
+
+REPORT_DIR=$(pwd)/out/acceptance-tests-report
+mkdir -p $REPORT_DIR
+
+$CONTAINER_RUNTIME build -f test/acceptance/Dockerfile.allure -t $NAME test/acceptance
+
+$CONTAINER_RUNTIME rm -f $NAME || true
+
+if [ $ACTION == "generate" ]; then
+    $CONTAINER_RUNTIME run --rm --name $NAME -v $(pwd)/out/acceptance-tests:/allure/results -v $REPORT_DIR:/allure/report $NAME $ACTION
+    echo -e "\nAcceptance tests report was generated into $REPORT_DIR directory" 
+elif [ $ACTION == "serve" ]; then
+    $CONTAINER_RUNTIME run -d --rm --name $NAME -v $(pwd)/out/acceptance-tests:/allure/results -p 8088:8080 $NAME $ACTION
+    CID=$($CONTAINER_RUNTIME ps -a -f name=$NAME -q)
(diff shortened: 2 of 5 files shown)


Avni-Sharma:
This looks good. Anything else to be added to this @pmacik ?

pmacik:
> This looks good. Anything else to be added to this @pmacik ?

@Avni-Sharma Nope, it's done, thx

hack/generate-allure-report.sh
```diff
@@ -0,0 +1,23 @@
...
+ALLURE_VERSION=2.13.5
+ALLURE_DIR=out/allure
+ALLURE_BIN=$ALLURE_DIR/allure-$ALLURE_VERSION/bin
+
+if [ -z $JAVA_HOME ]; then
+    mkdir -p out
+    curl -s -L https://download.java.net/java/GA/jdk15.0.1/51f4f36ad4ef43e39d0dfdbaf6549e32/9/GPL/openjdk-15.0.1_linux-x64_bin.tar.gz -o out/java.tar.gz
+    tar -C out -xvf out/java.tar.gz
+    export JAVA_HOME=$(readlink -m out/jdk-15.0.1)
+fi
```
pedjak, line 12: I am not sure if people would like to install java just for being able to visualize reports.. IMHO, it is a too hard expectation. Instead, how about crafting a Dockerfile that would have all tools needed in it, and then use it from this script to generate reports?

Check https://github.com/fescobar/allure-docker-service

test/acceptance/README.md
```diff
@@ -1,10 +1,230 @@
...
+
+1 feature passed, 0 failed, 3 skipped
+1 scenario passed, 0 failed, 16 skipped
+10 steps passed, 0 failed, 179 skipped, 0 undefined
+Took 0m49.231s
+```
+
+## Inspect test results
+
+During the test execution the result of individual steps are shown as the execution goes. That is the most obvious way to see the results, but it is far from elegant - especially with a growing number of scenarios. However, the framework also records the test results in a standardized way of xUnit format, which can be found under `$WORKSPACE/out/acceptance-tests` directory.
```
pedjak, line 214: Should introduce somewhere `test-acceptance-artifacts` rule and its usage? should we really call it explicitly or that could a part of test-acceptance rule execution?

test/acceptance/README.md
```diff
@@ -1,10 +1,230 @@
...
+### Install Allure tool
+
+Follow the [instructions](https://docs.qameta.io/allure/#_installing_a_commandline).
+
+### Show test result report
+
+```bash
+allure serve out/acceptance-tests
+```
+
```
pedjak, line 229: See above the idea bout using a Docker image...

pedjak (changes requested):

test/acceptance/README.md
```diff
@@ -1,10 +1,230 @@
...
+Let's use `WORKSPACE` as a placeholder for the path of your local Service Binding Operator repository.
+
+```bash
+export WORKSPACE=<path-to-sbo-repo>
+cd $WORKSPACE
+```
+
+Every output artifact (such as SBO's logs or test reports) related to the acceptance tests can be found under `$WORKSPACE/out/acceptance-tests` directory during and after the test execution.
+
+The acceptance test scenarios are defined in a [Gherkin Syntax](https://cucumber.io/docs/gherkin/), that is `Given -> When -> Then` while the intividual steps are in a English-like, human readable, language (e.g. [bindAppToService.feature](https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToService.feature)).
```
Avni-Sharma, line 87: Nit - individual

test/acceptance/README.md
```diff
@@ -1,10 +1,230 @@
...
+.
+```
+
+This command will prepare the environment similarly to the previous one and:
+* Builds and start a local instance of the operator
+* Executes the `smoke` sub-set of acceptance tests [features and scenarios](https://github.com/redhat-developer/service-binding-operator/tree/master/test/acceptance/features) - marked with `@smoke` tag.
+
+### Run acceptance tests with remote instance of SBO
+
+It is possible to run the acceptance tests with SBO already running, that is without building and starting the local instance and without installing the dependent resources (such as CRD, role, role-binding, service-account, etc.) required when the local instance is used. That is particulary useful for testing SBO installed on a remote cluster or running the acceptance tests with various versions of the SBO.
```
Avni-Sharma, line 146: Nit - particularly

test/acceptance/Dockerfile.allure
```diff
@@ -0,0 +1,5 @@
+FROM yhxue911/python-allure
```
pedjak, line 1: we need to copy this image to quay due to the recent pull limits introduced by DockerHub

test/acceptance/Dockerfile.allure
```diff
@@ -0,0 +1,5 @@
...
+
+RUN mkdir -p /allure/results && mkdir -p /allure/report
```
pedjak, line 3: you do not need to create these folder since you are mounting them always from host in `generate-allure-report.sh`

test/acceptance/Dockerfile.allure
pedjak: this is not interactive session, so `-it` option can be removed.

hack/generate-allure-report.sh
pmacik: Good idea, thx... done

test/acceptance/README.md
```diff
@@ -1,10 +1,212 @@
...
+.
+.
+
+1 feature passed, 0 failed, 3 skipped
+1 scenario passed, 0 failed, 16 skipped
+10 steps passed, 0 failed, 179 skipped, 0 undefined
+Took 0m49.231s
+```
+
+That's all Folks! ...well, at least for now.
```
pedjak, line 212: Should we mention somewhere `test-acceptance-report` rule and its usage? Or how to visualize the reports?

Makefile
```diff
@@ -293,9 +294,14 @@ endif
...
 
+.PHONY: test-acceptance-report
+## Generate acceptance test report
+test-acceptance-report:
+	$(Q)$(HACK_DIR)/generate-allure-report.sh
+
 .PHONY: test-acceptance-artifacts
 ## Collect artifacts from acceptance tests to be archived in CI
-test-acceptance-artifacts:
+test-acceptance-artifacts: test-acceptance-report
```
pedjak, line 304: generating reports should optional IMHO, otherwise, it is going to fail where docker is not installed.

test/acceptance/README.md
pmacik: The `test-acceptance-artifacts` is a separated target [used in CI](https://github.com/openshift/release/blob/3a0faadab90334bc3295700207f30f44f0f3359c/ci-operator/config/redhat-developer/service-binding-operator/redhat-developer-service-binding-operator-master__4.5.yaml#L31) to collect artifacts from acceptance tests to be archived. It is a separated target because we want to ensure the artifacts are collected no matter how the actual acceptance test execution goes - We wouldn't want to loose the artifacts if the `test-acceptance` fails.

test/acceptance/README.md
pmacik: I've introduced a docker image for generating the allure reports...

test/acceptance/Dockerfile.allure
pmacik: good points, will update mirror the image and update the command

test/acceptance/README.md
pmacik: Thx, fixed

test/acceptance/README.md
pmacik: Thx, fixed

test/acceptance/Dockerfile.allure
pmacik: good point, thx.

hack/allure-report.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -ex
+
+podman version 2>/dev/null >/dev/null || NO_DOCKER=$?
```
pedjak, line 3: can we make `podman` as default choice but also allow using `docker` through configuring some env var?

test/acceptance/Dockerfile.allure
```diff
@@ -0,0 +1,10 @@
+FROM quay.io/pmacik/yhxue911-python-allure
+
+RUN echo -e "#!/bin/bash\n\
+    if [ \$1 == \"generate\" ]; then allure generate /allure/results -o /allure/report --clean; fi\n\
+    if [ \$1 == \"serve\" ]; then allure serve -p 8080 /allure/results; fi" >> /entrypoint.sh && chmod +x /entrypoint.sh
```
pedjak, line 5: Instead of crafting `entrypoint.sh` during the build, please add it to the repo under `test/acceptance/entrypoint.sh` - then you need to do only:

```
COPY entrypoint.sh /entrypoint.sh
```

pmacik: done

hack/allure-report.sh
pmacik: done

hack/allure-report.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -ex
+
+CE=${CE:-podman}
```
pedjak, line 3: can we come up with some better name and document it perhaps on the level of Makefile? How about `CONTAINER_RUNTIME`?

hack/allure-report.sh
```diff
@@ -0,0 +1,26 @@
...
+
+$CE info 2>/dev/null >/dev/null || NO_CE=$?
+if [ -n "$NO_CE" ]; then
+    echo "WARNING $CE is required but not installed, skipping..."
+    exit 0
```
pedjak, line 8: This can make the issue silent and not visible. How about that we exit here with `exit 1`?

Makefile
pmacik: good point

hack/allure-report.sh
pmacik: Good idea

hack/allure-report.sh
pmacik: good pint

test/acceptance/README.md
pmacik: Done

Makefile
```diff
@@ -363,9 +376,9 @@ push-operator: prepare-csv
 
 ## Push-Image: push container image to upstream, including latest tag.
 push-image: build-image
-	podman tag "$(OPERATOR_IMAGE):$(OPERATOR_TAG_LONG)" "$(OPERATOR_IMAGE):latest"
-	-podman push "$(OPERATOR_IMAGE):$(OPERATOR_TAG_LONG)"
-	-podman push "$(OPERATOR_IMAGE):latest"
+	$(CONTAINER_RUNTIME) tag "$(OPERATOR_IMAGE):$(OPERATOR_TAG_LONG)" "$(OPERATOR_IMAGE):latest"
```
pedjak, line 379: travis CI conf should be updated because it depends on docker. Or to make `docker` as default?

hack/allure-report.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash -ex
+
+CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-podman}
```
pedjak, line 3: is this line needed now?

pmacik: It is for the case you want to run the script standalone

Makefile
pmacik: good point... let's switch to `docker` as default

hack/allure-report.sh
```diff
@@ -0,0 +1,29 @@
...
+ACTION=${1:-generate}
+
+NAME=test-acceptance-report
+
+REPORT_DIR=$(pwd)/out/acceptance-tests-report
+mkdir -p $REPORT_DIR
+
+$CONTAINER_RUNTIME build -f test/acceptance/Dockerfile.allure -t $NAME test/acceptance
+
+if [ $($CONTAINER_RUNTIME ps -a -f name=$NAME | grep $NAME | wc -l) != 0 ]; then $CONTAINER_RUNTIME rm -f $NAME; fi
```
pedjak, line 20: should not we stop container only if action is `generate`?

pmacik: We are running a container with the same `$NAME` in case of both actions. The point of the step is to ensure that there is no such container running when we start it (in both actions).

pmacik: +1 on simplifying it

PR #679, opened 2020-09-17
Fixed: owned resources are listed again in CSV

pedjak:
`pkg/apis/apps` folder renamed to `pkg/apis/operators` to match operator-sdk conventions, and hence reenable listing owned resources into the generated CSVs.
With this fix, we can again push images to quay. Currently Travis jobs are failing, e.g.:

https://travis-ci.org/github/redhat-developer/service-binding-operator/builds/728020714

In addition, the following tweaks are made:
* clean-up of the required rules
* `service-binding-operator.v0.0.23.clusterserviceversion.yaml` serves as template and contains again the owned resources
* `prepare-bundle-to-quay` make target requires just python3 as other similar targets


deploy/olm-catalog/service-binding-operator/0.0.23/service-binding-operator.v0.0.23.clusterserviceversion.yaml
```diff
@@ -41,7 +41,14 @@ metadata:
   namespace: placeholder
 spec:
   apiservicedefinitions: {}
-  customresourcedefinitions: {}
+  customresourcedefinitions:
+      owned:
+      - description: ServiceBinding expresses intent to bind an operator-backed service
```
akashshinde, line 46: @pedjak do we need to mention `operator-backed service` ? since we also support non-operator backed services as well.

akashshinde (changes requested):

deploy/olm-catalog/service-binding-operator/0.0.23/service-binding-operator.v0.0.23.clusterserviceversion.yaml
pedjak: You are right, but the source of this line is https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/apis/apps/v1alpha1/servicebinding_types.go#L105 - it would be good revisit the doc/descriptions in a dedicated PR. This PR is just about bringing the things back, i.e. to make CSV valid agin.

akashshinde (approved):

PR #680, opened 2020-09-21
Add validation to use 'services' and not 'service'.

pratikjagrut:
### Motivation

We are deprecating service and supporting services so updating CRD to enforce the use of CSV.

In the current situation, CR gets applied if it has service instead of services and on reconciling the operator crashes. So this PR will add validation to use services and if not used then CR will not be applied and will throw an error.

### DIFF
--- a/deploy/crds/operators.coreos.com_servicebindings_crd.yaml
+++ b/deploy/crds/operators.coreos.com_servicebindings_crd.yaml
@@ -13,6 +13,7 @@ spec:
     - sbrs
     singular: servicebinding
   scope: Namespaced
+  preserveUnknownFields: false
   subresources:
     status: {}
   validation:
@@ -259,6 +260,8 @@ spec:
                 type: object
               type: array
           type: object
+          required:
+              - "services"
         status:
           description: ServiceBindingStatus defines the observed state of ServiceBinding
           properties:


deploy/crds/operators.coreos.com_servicebindings_crd.yaml
```diff
@@ -259,6 +260,8 @@ spec:
                 type: object
               type: array
           type: object
+          required:
+              - "services"
```
sbose78, line 264: Oh I thought we had removed "service" 🤔

deploy/crds/operators.coreos.com_servicebindings_crd.yaml
```diff
@@ -13,6 +13,7 @@ spec:
     - sbrs
     singular: servicebinding
   scope: Namespaced
+  preserveUnknownFields: false
```
pedjak, line 16: we need to specify the annotation on `servicebinding_types.go`, otherwise, this is going to removed by the next invocation of `operator-sdk generate crds`

deploy/crds/operators.coreos.com_servicebindings_crd.yaml
pedjak, line 264: They are actually optional, see: https://github.com/redhat-developer/service-binding-operator/blob/6d7a3aa2a396d7d02458a3f898807cdd71a0bae2/pkg/apis/apps/v1alpha1/servicebinding_types.go#L26

pedjak (changes requested):

deploy/crds/operators.coreos.com_servicebindings_crd.yaml
pratikjagrut: yes, but if I'm not wrong Kuberners allows us to make CR with singular and plural fields and we're not dealing with service so need to add validation to use services.

pratikjagrut: Yes, but I could not find any other way to avoid using service and enforce use of services. Do you have anything in mind which I can try?

deploy/crds/operators.coreos.com_servicebindings_crd.yaml
pratikjagrut: Ohh true I'll need to add it there as well, nice catch thank you.

deploy/crds/operators.coreos.com_servicebindings_crd.yaml
pedjak: Not sure what you are trying to solve in this PR. `service` fields are simply ignored now, also not part of the schema. Hence, why do we need to enforce anything?

pratikjagrut: If we use the service field instead of services in SBR then the operator will crash at reconciling so to avoid happening that I'm trying to make sure we're using services not service.

Issue #665, opened 2020-09-07
Create an overall `Ready` to indicate whether the SBR reconcile is completed successfully


cdlliuy:
According to the discuss on last call for  https://github.com/redhat-developer/service-binding-operator/pull/565#issuecomment-684356894 , I would like to get more comments on the proposal listed by @DhritiShikhar  
![Screenshot from 2020-09-03 17-30-47](https://user-images.githubusercontent.com/17263036/92346066-3df44700-f0fe-11ea-9cef-aea7ddcf81bc.png)

Any objection to go ahead with this table for the `ready` status under different condition?

isutton:
@cdlliuy 

To confirm, the idea would be re-introduce the `Ready` condition, and a `True` status might be set in the following circumstances:

* `CollectionReady` is `True` and `InjectionReady` is `False` with the `EmptyApplicationSelector` reason, indicating the user **has not declared an application**; or
* `CollectionReady` is `True` and `InjectionReady` is also `True`.

Does this match your expectations?

cdlliuy:
yes,  @isutton ,  exactly.

cdlliuy:
@isutton , if we already get agreement on this, maybe we can put effort to get start to add `ready` ?  Let us decide this on Thursday's call?

DhritiShikhar:
@cdlliuy Sure

DhritiShikhar:
@pedjak 

Based on the discussion in the last community call,
there are two happy paths:

[1] CollectionReady -> True + InjectionReady -> False
Intermediate secret is populated and ready to be used. 
This is generic binding -> https://github.com/redhat-developer/service-binding-operator/pull/350

[2] CollectionReady -> True + InjectionReady -> True

Another `Ready` condition was proposed because as far as I understand, @cdlliuy 's team needs something to represent these two scenario. Refer this -> https://github.com/redhat-developer/service-binding-operator/pull/565#issuecomment-684356894

pedjak:
By looking at the condition table above, it appears to me that they are actually not independent from each other, i.e. some combinations are not possible. Hence, we can simplify the model a lot: we can keep just `Ready` type and incorporate other two as reasons:

Type|Status|Reason|Message
-|-|-|-
Ready|True|AppBound|Application Foo bound to service(s)
Ready|False|EmptyServiceSelector|Empty service selector
Ready|False|ServiceNotFound|Service Bar not found
Ready|False|EmptyApplicationSelector|Empty application selector
[...]

PR #681, opened 2020-09-21
Create an overall `Ready` to indicate whether the SBR reconcile is completed successfully #665

qibobo:
According to the discussion in https://github.com/redhat-developer/service-binding-operator/issues/665, we need to add a overall Ready condition to show what is the status of the ServiceBinding CR.

### DIFF
--- a/pkg/controller/servicebinding/reconciler.go
+++ b/pkg/controller/servicebinding/reconciler.go
@@ -6,6 +6,7 @@ import (
 
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -15,10 +16,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/redhat-developer/service-binding-operator/pkg/apis/operators/v1alpha1"
+	"github.com/redhat-developer/service-binding-operator/pkg/converter"
 	"github.com/redhat-developer/service-binding-operator/pkg/log"
 )
 
 const (
+	// BindingReady indicates that the overall sbr succeeded
+	BindingReady conditionsv1.ConditionType = "Ready"
 	// CollectionReady indicates readiness for collection and persistance of intermediate manifests
 	CollectionReady conditionsv1.ConditionType = "CollectionReady"
 	// InjectionReady indicates readiness to change application manifests to use those intermediate manifests
@@ -32,6 +36,8 @@ const (
 	EmptyApplicationReason = "EmptyApplication"
 	// ApplicationNotFoundReason is used when the application is not found.
 	ApplicationNotFoundReason = "ApplicationNotFound"
+	// ServiceNotFoundReason is used when the service is not found.
+	ServiceNotFoundReason = "ServiceNotFound"
 )
 
 // Reconciler reconciles a ServiceBinding object
@@ -131,10 +137,12 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 				Message: errEmptyServices.Error(),
 			},
 			conditionsv1.Condition{
-				Type:    InjectionReady,
-				Status:  corev1.ConditionFalse,
-				Reason:  EmptyServiceSelectorsReason,
-				Message: errEmptyServices.Error(),
+				Type:   InjectionReady,
+				Status: corev1.ConditionFalse,
+			},
+			conditionsv1.Condition{
+				Type:   BindingReady,
+				Status: corev1.ConditionFalse,
 			},
 		)
 		if updateErr == nil {
@@ -157,9 +165,30 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		r.restMapper,
 	)
 	if err != nil {
+		//handle service not found error
+		if k8serrors.IsNotFound(err) {
+			err = updateSBRConditions(r.dynClient, sbr,
+				conditionsv1.Condition{
+					Type:    CollectionReady,
+					Status:  corev1.ConditionFalse,
+					Reason:  ServiceNotFoundReason,
+					Message: err.Error(),
+				},
+				conditionsv1.Condition{
+					Type:   InjectionReady,
+					Status: corev1.ConditionFalse,
+				},
+				conditionsv1.Condition{
+					Type:   BindingReady,
+					Status: corev1.ConditionFalse,
+				},
+			)
+			if err != nil {
+				logger.Error(err, "Failed to update SBR conditions", "sbr", sbr)
+			}
+		}
 		return requeueError(err)
 	}
-
 	binding, err := buildBinding(
 		r.dynClient,
 		sbr.Spec.CustomEnvVar,
@@ -219,4 +248,22 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	logger.Info("Binding applications with intermediary secret...")
 	return sb.bind()
+}
+
+func updateSBRConditions(dynClient dynamic.Interface, sbr *v1alpha1.ServiceBinding, conditions ...conditionsv1.Condition) error {
+	for _, v := range conditions {
+		conditionsv1.SetStatusCondition(&sbr.Status.Conditions, v)
+	}
+	u, err := converter.ToUnstructured(sbr)
+	if err != nil {
+		return err
+	}
+
+	nsClient := dynClient.
+		Resource(groupVersion).
+		Namespace(sbr.GetNamespace())
+
+	_, err = nsClient.UpdateStatus(u, metav1.UpdateOptions{})
+
+	return err
 }
--- a/pkg/controller/servicebinding/reconciler_test.go
+++ b/pkg/controller/servicebinding/reconciler_test.go
@@ -133,6 +133,7 @@ func TestApplicationByName(t *testing.T) {
 
 		requireConditionPresentAndTrue(t, CollectionReady, sbrOutput.Status.Conditions)
 		requireConditionPresentAndTrue(t, InjectionReady, sbrOutput.Status.Conditions)
+		requireConditionPresentAndTrue(t, BindingReady, sbrOutput.Status.Conditions)
 
 		require.Equal(t, 1, len(sbrOutput.Status.Applications))
 		expectedStatus := v1alpha1.BoundApplication{
@@ -196,6 +197,7 @@ func TestReconcilerReconcileUsingSecret(t *testing.T) {
 
 		requireConditionPresentAndTrue(t, CollectionReady, sbrOutput.Status.Conditions)
 		requireConditionPresentAndTrue(t, InjectionReady, sbrOutput.Status.Conditions)
+		requireConditionPresentAndTrue(t, BindingReady, sbrOutput.Status.Conditions)
 
 		require.Equal(t, reconcilerName, sbrOutput.Status.Secret)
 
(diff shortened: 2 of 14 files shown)


pedjak (changes requested):
acceptance tests need to updated to reflect and check this new flag.

qibobo:
@isutton 
Could you help review this PR?

qibobo:
@pedjak done

pkg/controller/servicebinding/reconciler.go
```diff
@@ -31,6 +36,10 @@ const (
 	EmptyApplicationReason = "EmptyApplication"
 	// ApplicationNotFoundReason is used when the application is not found.
 	ApplicationNotFoundReason = "ApplicationNotFound"
+	// FailedToGatherServiceInformationReason is used when the it is failed to fetch backend service information.
+	FailedToGatherServiceInformationReason = "FailedToGatherServiceInformation"
+	// FailedToSaveSecretReason is used when the it is failed to save intermediate secret.
```
Avni-Sharma, line 41: Do we need this ? Could not find any reference to this particular FailedToSaveReason in this PR.

qibobo: removed

qibobo:
@Avni-Sharma @isutton @pmacik @pedjak 
The acceptance test is failed, but seems I did not modify the error part. Could you help confirm whether it is the acceptance test problem or the issue of my PR?

```
test/acceptance/features/steps/steps.py:225
      Assertion Failed: 
      Expected: not SBO jq result "{result}" should match "{json_value_regex}"
           but: was <None>
```

pkg/controller/servicebinding/reconciler.go
```diff
@@ -157,16 +170,55 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+				Status:  corev1.ConditionFalse,
+				Reason:  FailedToGatherServiceInformationReason,
+				Message: err.Error(),
+			},
+			conditionsv1.Condition{
+				Type:    BindingReady,
+				Status:  corev1.ConditionFalse,
+				Reason:  FailedToGatherServiceInformationReason,
+				Message: err.Error(),
+			},
```
pedjak, line 192: We did not even reach the state when we can say if something is injection ready or not, because we did not gather service data at all. Conditions must not exist at all. We should update them only if there is a change in status, and in that case we are going to update last transition time as well. In this PR, it looks like we have 3 conditions, but they all share value and reason. Why then not collapse it into single condition.

BTW, excellent read about conditions: https://dev.to/maelvls/what-the-heck-are-kubernetes-conditions-for-4je7

pkg/controller/servicebinding/reconciler.go
```diff
@@ -157,16 +170,55 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+			conditionsv1.Condition{
+				Type:    BindingReady,
+				Status:  corev1.ConditionFalse,
+				Reason:  FailedToGatherServiceInformationReason,
+				Message: err.Error(),
+			})
```
pedjak, line 221: same comments as above. Error in `buildServiceContexts` or `buildBinding` does not move the process into the same condition.

pkg/controller/servicebinding/reconciler.go
```diff
@@ -185,6 +237,27 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+				Status:  corev1.ConditionFalse,
+				Reason:  FailedToGatherServiceInformationReason,
+				Message: err.Error(),
+			},
+			conditionsv1.Condition{
+				Type:    BindingReady,
+				Status:  corev1.ConditionFalse,
+				Reason:  FailedToGatherServiceInformationReason,
+				Message: err.Error(),
+			},
```
pedjak, line 258: same comment  as above.

pkg/controller/servicebinding/sbrcontroller.go
```diff
@@ -106,6 +106,12 @@ func isOfKind(obj runtime.Object, kind string) bool {
 	return strings.EqualFold(obj.GetObjectKind().GroupVersionKind().Kind, kind)
 }
 
+func createFunc(logger *log.Log) func(event.CreateEvent) bool {
+	return func(e event.CreateEvent) bool {
+		return true
+	}
+}
+
```
pedjak, line 114: why do we need this?

pedjak (changes requested):
This PR should introduce new condition, but should alte the logic around the existing conditions.

sbose78:
@qibobo Could you please add a description why this is needed :) ?

pkg/controller/servicebinding/reconciler_test.go
```diff
@@ -257,6 +259,63 @@ func TestReconcilerReconcileUsingVolumes(t *testing.T) {
 	})
 }
 
+func TestServiceNotFound(t *testing.T) {
+	backingServiceResourceRef := "backingService1"
+	matchLabels := map[string]string{
```
Avni-Sharma, line 264: Do we require matchLabels?

pkg/controller/servicebinding/reconciler_test.go
```diff
@@ -257,6 +259,63 @@ func TestReconcilerReconcileUsingVolumes(t *testing.T) {
...
+		"connects-to": "database",
+		"environment": "reconciler",
+	}
+	f := mocks.NewFake(t, reconcilerNs)
+	f.AddMockedUnstructuredServiceBinding(reconcilerName, backingServiceResourceRef, "", deploymentsGVR, matchLabels)
+	f.AddMockedUnstructuredCSV("cluster-service-version-list")
```
Avni-Sharma, line 270: CSV might not be necessary
depends whether we are checking via annotations on CRDs or via x-descriptors

pkg/controller/servicebinding/servicebinder.go
```diff
@@ -273,14 +279,7 @@ func removeFinalizer(sbr *v1alpha1.ServiceBinding) {
 // handleApplicationError handles scenarios when:
 // 1. application not declared in the Service Binding
 // 2. application not found
-func (b *serviceBinder) handleApplicationError(reason string, applicationError error, sbrStatus *v1alpha1.ServiceBindingStatus) (reconcile.Result, error) {
```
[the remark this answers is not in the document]
qibobo: @Avni-Sharma 
Put the condition setting out of handleApplicationError function as there are two callers:
caller1: https://github.com/redhat-developer/service-binding-operator/blob/2f561b7bb5cecb0e2d497d3bff352b1edf578ff6/pkg/controller/servicebinding/servicebinder.go#L327
caller2: https://github.com/redhat-developer/service-binding-operator/blob/2f561b7bb5cecb0e2d497d3bff352b1edf578ff6/pkg/controller/servicebinding/servicebinder.go#L343

caller1 is for application empty so InjectionReady is false and BindingReady is true.
caller2 is for application not found so both InjectionReady and BindingReady are false.
So here we need to put the condition setting out of this function.

qibobo:
> @qibobo Could you please add a description why this is needed :) ?

@sbose78 Done. Add the issue that for the discussion of this PR.

pkg/controller/servicebinding/reconciler_test.go
```diff
@@ -447,6 +520,33 @@ func TestEmptyApplication(t *testing.T) {
 
 	requireConditionPresentAndTrue(t, CollectionReady, sbrOutput.Status.Conditions)
 	requireConditionPresentAndFalse(t, InjectionReady, sbrOutput.Status.Conditions)
+	requireConditionPresentAndTrue(t, BindingReady, sbrOutput.Status.Conditions)
+}
+
+// TestEmptyServiceSelector tests that Status is successfully updated when ServiceSelector is empty
+func TestEmptyServiceSelector(t *testing.T) {
```
pedjak, line 527: please rename this function to reflect what the comment above says.

pkg/controller/servicebinding/sbrcontroller.go
```diff
@@ -145,6 +145,7 @@ func buildGVKPredicate(logger *log.Log) predicate.Funcs {
 		UpdateFunc: updateFunc(logger),
 		DeleteFunc: func(e event.DeleteEvent) bool {
 			// evaluates to false if the object has been confirmed deleted
+			logger.Debug("Resource update event received", "GVK", e.Object.GetObjectKind(), "name", e.Meta.GetName())
```
pedjak, line 148: do we really need this change in PR? BTW, the message is misleading, since we receive here delete event.

pkg/controller/servicebinding/servicebinder.go
```diff
@@ -326,13 +325,33 @@ func (b *serviceBinder) bind() (reconcile.Result, error) {
...
+				Type:    InjectionReady,
+				Status:  corev1.ConditionFalse,
+				Reason:  ApplicationNotFoundReason,
+				Message: errApplicationNotFound.Error(),
+			})
+			conditionsv1.SetStatusCondition(&sbrStatus.Conditions, conditionsv1.Condition{
+				Type:   BindingReady,
+				Status: corev1.ConditionFalse,
+			})
+			return b.handleApplicationError(errApplicationNotFound, sbrStatus)
```
pedjak, line 354: this was part of `handleApplicationError` - why taking it out from the function and placing here?

pedjak (changes requested):

pedjak:
Does this PR include the work in https://github.com/redhat-developer/service-binding-operator/pull/502 ?

pkg/controller/servicebinding/servicebinder.go
qibobo: @pedjak 
https://github.com/redhat-developer/service-binding-operator/pull/681#discussion_r496023405

pkg/controller/servicebinding/reconciler_test.go
qibobo: @pedjak 
What name should it be instead of TestEmptyServiceSelector?

qibobo:
> Does this PR include the work in #502 ?

@pedjak 
Both this PR and 502 are not merged yet, so we can not include each other.

pkg/controller/servicebinding/reconciler_test.go
pedjak: the comment line above has a few hints how the test function name could look like. It should contain the context (empty selector) and the consequence (status is successfully updated)

pedjak:
> Both this PR and 502 are not merged yet, so we can not include each other.

#502 is about handling situation when service is not found. Does your PR cover such situation?

qibobo:
> > Both this PR and 502 are not merged yet, so we can not include each other.
> 
> #502 is about handling situation when service is not found. Does your PR cover such situation?

@pedjak 
It does not cover the scenario. In PR 502,  a new ServiceNotFoundReason is introduced, so either my PR got merged first then PR502 should add relevant change, or PR502 got merged first and then I will do the change. Is it reasonable?

pedjak:
> It does not cover the scenario. In PR 502, a new ServiceNotFoundReason is introduced, so either my PR got merged first then PR502 should add relevant change, or PR502 got merged first and then I will do the change. Is it reasonable?

@qibobo thanks for the clarification - IMHO your PR should go first.

pkg/controller/servicebinding/reconciler.go
```diff
@@ -131,10 +138,12 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-				Type:    InjectionReady,
-				Status:  corev1.ConditionFalse,
-				Reason:  EmptyServiceSelectorsReason,
-				Message: errEmptyServicess.Error(),
+				Type:   InjectionReady,
+				Status: corev1.ConditionFalse,
+			},
+			conditionsv1.Condition{
+				Type:   BindingReady,
+				Status: corev1.ConditionFalse,
```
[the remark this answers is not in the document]
qibobo: @pedjak 
According to https://github.com/redhat-developer/service-binding-operator/issues/665#issue-694667113,
when service selector is empty, CollectionReady, InjectionReady and BindingReady should be updated to False.

pkg/controller/servicebinding/reconciler.go
```diff
@@ -157,9 +166,27 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+			_ = updateSBRConditions(r.dynClient, sbr,
+				conditionsv1.Condition{
+					Type:    CollectionReady,
+					Status:  corev1.ConditionFalse,
+					Reason:  ServiceNotFoundReason,
+					Message: err.Error(),
+				},
+				conditionsv1.Condition{
+					Type:   InjectionReady,
+					Status: corev1.ConditionFalse,
```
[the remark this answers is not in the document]
qibobo: @pedjak 
https://github.com/redhat-developer/service-binding-operator/issues/665#issue-694667113
when service is not found, three False, too.

qibobo:
@pedjak @Avni-Sharma any more comments?

pkg/controller/servicebinding/reconciler.go
```diff
@@ -220,3 +246,21 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	logger.Info("Binding applications with intermediary secret...")
 	return sb.bind()
 }
+
+func updateSBRConditions(dynClient dynamic.Interface, sbr *v1alpha1.ServiceBinding, conditions ...conditionsv1.Condition) error {
```
Avni-Sharma, line 250: This function is returning an error type, but from where the function is being called, the err is not being handled(if an err is returned).

pkg/controller/servicebinding/servicebinder_test.go
```diff
@@ -513,6 +525,10 @@ func TestServiceBinder_Bind(t *testing.T) {
 				Reason:  EmptyApplicationReason,
 				Message: errEmptyApplication.Error(),
 			},
+			{
+				Type:   BindingReady,
+				Status: corev1.ConditionTrue,
```
Avni-Sharma, line 530: According to https://github.com/redhat-developer/service-binding-operator/issues/665#issue-694667113 CollectionReady : true and InjectionReady : False gives Ready : False . @qibobo Please let me know if I am missing something here

Avni-Sharma: Because in https://github.com/redhat-developer/service-binding-operator/issues/665#issuecomment-692596417 I see that 2nd row and 4th row (last row) have collectionReady :true InjectionReady : false but Ready is both True and False for 2 similar granularities. Idk what was eventually agreed upon(there are 2 tables demonstrated in the issue). We need to look into that and accordingly change and also update the comment here as well https://github.com/redhat-developer/service-binding-operator/pull/681/files#diff-c9af90357b70995ba06d05f3e039ec5bed01449ec59b7318238804f523ba6364R24

qibobo: @Avni-Sharma 
CollectionReady=True, InjectionReady=False with emptyAppSelector, then Ready is true as there in app to do bind, so the whole process is successful.
CollectionReady=True, InjectionReady=False with appNotFound, then Ready is false because there is error to do injection, so the whole process is failed.

pkg/controller/servicebinding/reconciler.go
```diff
@@ -15,10 +16,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/redhat-developer/service-binding-operator/pkg/apis/operators/v1alpha1"
+	"github.com/redhat-developer/service-binding-operator/pkg/converter"
 	"github.com/redhat-developer/service-binding-operator/pkg/log"
 )
 
 const (
+	// BindingReady indicates that the overall sbr succeeded
```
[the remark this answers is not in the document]
qibobo: @Avni-Sharma Sorry, do not get you.

pkg/controller/servicebinding/reconciler.go
qibobo: @Avni-Sharma 
There may be error when do the status updating, but in where call this function, seems we do not need check the error.
So your suggestion is it should not return an error right?

Avni-Sharma: returning an err is good, but we might need to inform the err to the user

pkg/controller/servicebinding/reconciler.go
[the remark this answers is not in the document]
Avni-Sharma: Seems like the link is broken. Basically a better description where we can inform that ready is CollectionReady true and Injection ready true or something on those lines

pkg/controller/servicebinding/servicebinder_test.go
Avni-Sharma: Thanks @qibobo  . In that case, can we pls also have the appNotFound reason test case as well. wherein we show CollectionReady = True, InjectionRedy = false with appNotFound and then show `ready` as false

pkg/controller/servicebinding/reconciler.go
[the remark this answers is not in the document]
pedjak: ok

pkg/controller/servicebinding/servicebinder.go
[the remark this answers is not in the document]
pedjak: Thanks. However, in order to improve readability of the code, and avoid copy/pasting `conditionsv1.SetStatusCondition` blocks around, how about to define `InjectionReadyStatus(...)`, `ReadyStatus`, ... on `ServiceBindingStatus` struct and use them across the code? IMHO, it would improve readbility and reusability of the code. Also, Ready condition could even autoderived from other two conditions in that case.

Avni-Sharma:
I also think that we can have acceptance tests for `ready` condition. For example https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToService.feature#L70

qibobo:
@Avni-Sharma 
updated.

qibobo:
> I also think that we can have acceptance tests for `ready` condition. For example https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToService.feature#L70

@Avni-Sharma I think they have been added.

Avni-Sharma:
> > I also think that we can have acceptance tests for `ready` condition. For example https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToService.feature#L70
> 
> @Avni-Sharma I think they have been added.

Oh , I meant specifically checking for `Ready` under conditions

qibobo:
@Avni-Sharma squashed the commits.

Avni-Sharma:
@qibobo Please feel free to create an issue regarding https://github.com/redhat-developer/service-binding-operator/pull/681#discussion_r510158311 . You can perhaps take it up later

PR #682, opened 2020-09-21
Remove obsolete e2e suite from repository

pmacik:
### Motivation

Currently the obsolete `e2e` test suite is no longer used nor maintained. The acceptance tests replaced it.

Concludes [APPSVC-648](https://issues.redhat.com/browse/APPSVC-648).

### Changes

This PR:
* Removes the `e2e` test suite from the codebase.
* Renames formely e2e-related `Makefile` targets to a more general names.
* Updates docs where e2e is mentioned with the acceptance tests instead.


pedjak:
could we submit commit 007f0330eab0d457d238f146f6fceddaf0806bd3 as a separate PR? IMHO, it is not the related to the scope of this PR.

pmacik:
> could we submit commit [007f033](https://github.com/redhat-developer/service-binding-operator/commit/007f0330eab0d457d238f146f6fceddaf0806bd3) as a separate PR? IMHO, it is not the related to the scope of this PR.

Ah, yes...

PR #687, opened 2020-09-24
[acceptance-tests] Fix timeout waiting for serverless/knative to be imported

pmacik:
### Motivation

Currently the knative-related acceptance tests scenarios fails for timeout sometimes.

### Changes
This PR:
* Increases various timeouts around quarkus app steps
* Changes the way to detect knative app is imported - uses number of replicas on deployment instead of checking `Available` status condition
* Changes the way to detect that serverless operator is running but checking for CSV and then for all the expected pods from it's deployments given by the CSV.

### Testing

`make test-acceptance`

### DIFF
--- a/test/acceptance/features/requirements.txt
+++ b/test/acceptance/features/requirements.txt
@@ -1,3 +1,4 @@
 behave==1.2.6
 pyshould==0.7.1
 requests==2.24.0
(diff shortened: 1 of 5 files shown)


pmacik:
Squashed the commits and rebased...

PR #628, opened 2020-08-20
Add support for new annotation and OLM descriptor format

isutton:
### Motivation

Refer to https://github.com/k8s-service-bindings/spec#annotation-examples for the format.

### Changes

This pull request substitutes the annotations' syntax with those specified in https://github.com/k8s-service-bindings/spec#annotation-examples.

The following use cases have been covered:

#### Contribute an entire Secret data to the binding

The annotation indicates that `.status.data.dbCredentials` in the Service resource must exist and be the name of a Secret in the same namespace.

```yaml
service.binding: path={.status.data.dbCredentials},objectType=Secret
```

The following is present in the Service resource, the contents of `dbCredentials` being a string with the name of a Secret existing in the same namespace as the Service itself.

```yaml
status:
  data:
    dbCredentials: database-cred-Secret  # Secret
```

The following content is present in the `database-cred-Secret` Secret resource:

```yaml
data:
  username: <base64 encoded string>
  password: <base64 encoded string>
```

Once the `ServiceBindingRequest` is reconciled, the following service binding values should be exposed to the application as environment variables:

| Variable | Value |
|---|---|
| `DBCREDENTIALS_USERNAME` | decoded string |
| `DBCREDENTIALS_PASSWORD` | decoded string |

---

#### Contribute an entire ConfigMap data to the binding

The annotation indicates that `.status.data.dbConfiguration` in the Service resource must exist and be the name of a ConfigMap in the same namespace.

```yaml
service.binding: path={.status.data.dbConfiguration},objectType=ConfigMap
```

The following is present in the Service resource, the contents of `dbConfiguration` being a string with the name of a ConfigMap existing in the same namespace as the Service itself.

```yaml
status:
  data:
    dbConfiguration: database-config     # ConfigMap
```

The following content is present in the `database-config` ConfigMap resource:

```yaml
data:
  port: 5432
  ipAddress: 192.168.1.250
  certificate: <base64 encoded certificate>
  db_timeout: 30
```

Once the `ServiceBindingRequest` is reconciled, the following service binding values should be exposed to the application as environment variables:

| Variable | Value |
|---|---|
| `DBCONFIGURATION_PORT` | `5432` |
| `DBCONFIGURATION_IPADDRESS` | `192.168.1.250` |
| `DBCONFIGURATION_CERTIFICATE` | `<base64 encoded certificate>` |
| `DBCONFIGURATION_DB_TIMEOUT` | `30` |

---

#### Contribute a single entry from a ConfigMap

[...]

PR #690, opened 2020-09-25
fix: there's no need to requeue if app is not found

pedjak:
Since there's nothing to recover from and there's a watch already
configured for further reverse lookups, the operator should not
requeue the work item.

Also, there's no need to update the service binding twice in the same
process because the finalizer has been already added during `bind`.


pedjak:
got merged as a part of #628

Issue #661, opened 2020-09-02
Binding Service using volumes broken


pedjak:
The only thing that might be fixed is `binding:volumemount:secret` option, but even that does not work, i.e. operator crashes with panic due to wrong type casting in https://github.com/redhat-developer/service-binding-operator/blob/f0748bbdc5d34adeb3ff215be55a9c053c194feb/pkg/controller/servicebindingrequest/binder.go#L176 

The functionalities introduced in https://github.com/redhat-developer/service-binding-operator/pull/83 are accompanied with a single unit test that cover the case with annotations on CSV: https://github.com/redhat-developer/service-binding-operator/blob/f0748bbdc5d34adeb3ff215be55a9c053c194feb/pkg/controller/servicebindingrequest/reconciler_test.go#L217 not sure how the test is green, need to investigate further, but setting on annotation something like binding:volumemount:secret:user simply does not work, I see in the log the error saying that this is unsupported format. 

Restoring this functionality should be performed according the annotation format documented in the [proposed descriptor/annotations' format](https://github.com/k8s-service-bindings/spec#olm-operator-descriptors)

sbose78:
@baijum 's original PR https://github.com/redhat-developer/service-binding-operator/pull/83 
@baijum how would we enhance this to acknowledge `$SERVICE_BINDING_ROOT` ?

sbose78:
Note from team discussion
```
 Scenario: Binding is injected as file into application pod
        Given OLM Operator "backend" is running
        * Imported test application "test-app" is running
        * Backend CR "backend-demo" is applied
            """
            apiVersion: "stable.example.com/v1"
            kind: Backend
            metadata:
                name: backend-demo
                annotations:
                    "service.binding/host": "path={.spec.host}"
            spec:
                host: example.common
            """
        When Service Binding request is applied
            """
            apiVersion: operators.coreos.com/v1alpha1
            kind: ServiceBinding
            metadata:
                name: binding-backend
            spec:
                mountPathPrefix: "/foo/bar"   # turn on volume mounts
                services:
                -   group: stable.example.com
                    version: v1
                    kind: Backend
                    name: backend-demo
                    bindAs:
                          file
                             -  host
                application:
                    name: test-app
                    group: apps
                    version: v1
                    resource: deployments
            """
        Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-backend" should be changed to "True"
[...]

PR #694, opened 2020-09-30
Feature list & Sample CR

sbose78:
* Add a list of key features and a sample CR.
* The list would help a reader understand what this projects aims to support in master.
* Given that our PR checks validated against OKD 4.4, I've added a dependency on kubernetes v1.17+

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,10 +100,12 @@ EOS
 
 ## Key Features
 
-* Binding with backing services represented by Kubernetes resources including third-party CRD-backed resources.
-* Binding secret generation based on OLM descriptors and CRD/CR annotations.
-* Binding with multiple-backing services.
-* Binding as volume mounts and environment variables.
+* Support Binding with backing services represented by Kubernetes resources including third-party CRD-backed resources.
+* Support binding with multiple-backing services.
+* Extract binding information based on annotations present in CRDs/CRs/resources. 
+* Extract binding values based on annotations present in OLM descriptors.
+* Project binding values as volume mounts.
+* Project binding values as environment variables.
 * Binding of PodSpec-based workloads.
 * Binding of non-PodSpec-based Kubernetes resources.
 * Custom binding variables composed from one or more backing services.
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-# The Service Binding Operator for Kubernetes
-## Connecting Applications with Services
+# The Service Binding Operator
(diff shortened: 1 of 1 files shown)


akashshinde:
/lgtm

@sbose78 can you please squash the commits ?

README.md
```diff
@@ -64,6 +97,19 @@ spec:
 EOS
 ```
 
+
+## Key Features
+
+* Binding with backing services represented by Kubernetes resources including third-party CRD-backed resources.
+* Binding secret generation based on OLM descriptors and CRD/CR annotations.
+* Binding with multiple-backing services.
+* Binding as volume mounts and environment variables.
```
Avni-Sharma, line 106: afaik for volume mount we have an issue https://github.com/redhat-developer/service-binding-operator/issues/661 . For annotations and descriptors also the `bindAs` is not being used.

README.md
```diff
@@ -64,6 +97,19 @@ spec:
 EOS
 ```
 
+
+## Key Features
+
+* Binding with backing services represented by Kubernetes resources including third-party CRD-backed resources.
+* Binding secret generation based on OLM descriptors and CRD/CR annotations.
```
baijum, line 104: Since both OLM descriptors and annotations provide the same functionality, why do we need to support both?

isutton: OLM descriptors give us both the same semantics annotations give us to extract binding values from resources, with the added benefit of providing hints for OpenShift's UI to enrich the user's experience.

README.md
isutton: I believe each feature should have its own bullet:

```md
* Project binding values as volume mounts.
* Project binding values as environment variables.
```

README.md
isutton: I believe each feature should have its own bullet:

```md
* Extract binding values based on annotations present in CRD/CR
* Extract binding values based on OLM descriptors 
```

README.md
sbose78: @Avni-Sharma This list is more of a "what we intend to support over time" so that a user can make decisions about considering this project for adoption. Therefore, I'm hoping to have this as a static list. What do you think?

As a separate task, we should have detailed docs on the extent of support for each that must vary between releases/PRs.

@isutton +1 will do!

sbose78:
Given that this is a readme change, and there weren't any obvious objections to the content, I'm taking this is.

README.md
sbose78: Done!

PR #698, opened 2020-10-01
Make services a required field in sbr spec

pratikjagrut:
### Motivation

https://issues.redhat.com/browse/APPSVC-696

### Testing

`make test-acceptance VERBOSE=3`

### DIFF
--- a/pkg/apis/operators/v1alpha1/servicebinding_types.go
+++ b/pkg/apis/operators/v1alpha1/servicebinding_types.go
@@ -23,7 +23,8 @@ type ServiceBindingSpec struct {
 	CustomEnvVar []corev1.EnvVar `json:"customEnvVar,omitempty"`
 
 	// Services is used to identify multiple backing services.
-	Services *[]Service `json:"services,omitempty"`
+	// +kubebuilder:validation:MinItems:=1
+	Services []Service `json:"services"`
 
 	// Application is used to identify the application connecting to the
 	// backing service operator.
--- a/pkg/controller/servicebinding/mapper.go
+++ b/pkg/controller/servicebinding/mapper.go
@@ -42,8 +42,7 @@ func isSecret(obj runtime.Object) bool {
 
 // isSBRService checks whether the given obj is a service in given sbr.
 func isSBRService(sbr *v1alpha1.ServiceBinding, obj runtime.Object) bool {
-	services := servicesOrEmptyList(sbr.Spec.Services)
-	for _, svc := range services {
+	for _, svc := range sbr.Spec.Services {
 		svcGVK := schema.GroupVersionKind{Group: svc.Group, Version: svc.Version, Kind: svc.Kind}
 		if obj.GetObjectKind().GroupVersionKind() == svcGVK {
 			return true
--- a/pkg/controller/servicebinding/reconciler.go
+++ b/pkg/controller/servicebinding/reconciler.go
@@ -78,14 +78,6 @@ func (r *reconciler) getServiceBinding(
 	return sbr, nil
 }
 
-func servicesOrEmptyList(maybeServices *[]v1alpha1.Service) []v1alpha1.Service {
-	services := []v1alpha1.Service{}
-	if maybeServices != nil {
-		services = *maybeServices
-	}
-	return services
-}
-
 // Reconcile a ServiceBinding by the following steps:
 // 1. Inspecting SBR in order to identify backend service. The service is composed by a CRD name and
 //    kind, and by inspecting "connects-to" label identify the name of service instance;
@@ -128,9 +120,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	ctx := context.Background()
 
-	selectors := servicesOrEmptyList(sbr.Spec.Services)
-
-	if len(selectors) == 0 {
+	if len(sbr.Spec.Services) == 0 {
 		_, updateErr := updateServiceBindingStatus(
 			r.dynClient,
 			sbr,
@@ -162,7 +152,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		logger.WithName("buildServiceContexts"),
 		r.dynClient,
 		sbr.GetNamespace(),
-		selectors,
+		sbr.Spec.Services,
 		sbr.Spec.DetectBindingResources,
 		r.restMapper,
 	)
--- a/test/acceptance/features/bindAppToService.feature
+++ b/test/acceptance/features/bindAppToService.feature
@@ -670,3 +670,61 @@ Feature: Bind an application to a service
         And The application env var "BACKEND_SPEC_DBNAME" has value "db-demo"
         And jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "rsa-2" should be changed to "True"
         And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "rsa-2" should be changed to "True"
+
+        @negative
+        Scenario: Service Binding with empty services is not allowed in the cluster
+        When Invalid Service Binding is applied
+            """
+            apiVersion: operators.coreos.com/v1alpha1
+            kind: ServiceBinding
+            metadata:
+                name: binding-request-empty-services
+            spec:
+                services:
+            """
+        Then Error message "invalid: spec.services: Invalid value: \"null\"" is thrown
+        And Service Binding "binding-request-empty-services" is not persistent in the cluster
+
+        @negative
+        Scenario: Service Binding without gvk of services is not allowed in the cluster
+        When Invalid Service Binding is applied
+            """
+            apiVersion: operators.coreos.com/v1alpha1
+            kind: ServiceBinding
+            metadata:
+                name: binding-request-without-gvk
+            spec:
+                services:
+                -   name: backend-demo
+            """
+        Then Error message "spec.services.group: Required value" is thrown
+        And Error message "spec.services.kind: Required value" is thrown
+        And Error message "spec.services.version: Required value" is thrown
+        And Service Binding "binding-request-without-gvk" is not persistent in the cluster
+
+        @negative
+        Scenario: Removing service from services field from existing serivce binding is not allowed
+        Given Service Binding is applied
+            """
+            apiVersion: operators.coreos.com/v1alpha1
+            kind: ServiceBinding
+            metadata:
+                name: binding-request-remove-service
+            spec:
+                services:
+                -   group: service.example.com
+                    version: v1
+                    kind: Backserv
+                    name: demo-backserv-cr-2
+            """
+        When Invalid Service Binding is applied
+            """
+            apiVersion: operators.coreos.com/v1alpha1
+            kind: ServiceBinding
+            metadata:
+                name: binding-request-remove-service
+            spec:
+                services:
+            """
+        Then Error message "invalid: spec.services: Required value" is thrown
+        And Service Binding "binding-request-remove-service" is not updated
--- a/test/acceptance/features/steps/service_binding.py
+++ b/test/acceptance/features/steps/service_binding.py
@@ -8,5 +8,11 @@ class ServiceBinding(object):
     openshift = Openshift()
 
     def create(self, yaml):
-        sbr_create_output = self.openshift.oc_apply(yaml)
-        return re.search(r'.*servicebinding.operators.coreos.com/.*(created|unchanged)', sbr_create_output)
+        return re.search(r'.*servicebinding.operators.coreos.com/.*(created|unchanged)', self.attempt_to_create(yaml))
+
+    def attempt_to_create(self, yaml):
+        return self.openshift.oc_apply(yaml)
+
+    def get_servicebinding_info_by_jsonpath(self, servicebinding_name, namespace, json_path):
+        openshift = Openshift()
+        return openshift.get_resource_info_by_jsonpath("servicebinding", servicebinding_name, namespace, json_path)
--- a/test/acceptance/features/steps/steps.py
+++ b/test/acceptance/features/steps/steps.py
@@ -174,8 +174,11 @@ sbr_is_applied_step = u'Service Binding is applied'
 
 @given(sbr_is_applied_step)
 @when(sbr_is_applied_step)
+@then(sbr_is_applied_step)
 def sbr_is_applied(context):
     sbr_yaml = context.text
+    metadata_name = re.sub(r'.*: ', '', re.search(r'name: .*', sbr_yaml).group(0))
+    context.sbr_name = metadata_name
     sbr = ServiceBinding()
     if context.__contains__("application") and context.__contains__("application_type"):
         application = context.application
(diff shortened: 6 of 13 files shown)


test/acceptance/features/bindAppToService.feature
```diff
@@ -585,3 +585,33 @@ Feature: Bind an application to a service
         And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-backend-a" should be changed to "False"
         And Secret "binding-request-backend-a" contains "BACKEND_READY" key with value "true"
         And Secret "binding-request-backend-a" contains "BACKEND_HOST" key with value "example.common"
+
+        @negative
+        Scenario: Create Service Binding with empty services
```
pedjak, line 590: you cannot create it, as you show later on. Hence, the scenario should reflect that, something like:

"Service Binding with empty services is not allowed in the cluster"

test/acceptance/features/bindAppToService.feature
```diff
@@ -585,3 +585,33 @@ Feature: Bind an application to a service
...
+        When Invalid Service Binding is applied
```
pedjak, line 591: We can reuse already existing step `Service Binding is applied`. What is the benefit of the new step?

test/acceptance/features/bindAppToService.feature
```diff
@@ -585,3 +585,33 @@ Feature: Bind an application to a service
...
+            """
+            apiVersion: operators.coreos.com/v1alpha1
+            kind: ServiceBinding
+            metadata:
+                name: binding-request-empty-services
+            spec:
+                services:
+            """
+        Then Openshift throws an error "invalid: spec.services: Invalid value: \"null\""
```
pedjak, line 600: Let be more generic here, for future plain k8s support: how about "Cluster reports the error "invalid: spec.services: Invalid value: \"null\""

Also, we really need to check that the submitted SB instance is not persisted in the cluster

test/acceptance/features/bindAppToService.feature
```diff
@@ -585,3 +585,33 @@ Feature: Bind an application to a service
...
+
+        @negative
+        Scenario: Create Service Binding without GVK of service
```
pedjak, line 603: see above.

pedjak (changes requested):
in addition to other comments:

* please squash commits
* we need additional test case where `services` fields is not present in SB at all
* we need scenario where existing valid SB gets updated so that `services` fields are removed.

test/acceptance/features/steps/steps.py
```diff
@@ -425,3 +426,17 @@ def etc_cluster_is_running(context, etcd_name):
...
         assert etcd_cluster.create() is True, "etcd cluster is not created"
         assert etcd_cluster.is_present() is True, "etcd cluster is not present"
+
+
+@when(u'Invalid Service Binding is applied')
+def invalid_sbr_is_applied(context):
+    yaml = context.text
+    sbr = ServiceBinding()
+    output = sbr.attempt_to_create(yaml)
+    context.error = output
```
pmacik, line 436: Please rename the context variable to `context.expected_error` so that it is clear that it is expected instead of being confused with an ordinary error.

test/acceptance/features/steps/service_binding.py
```diff
@@ -10,3 +10,6 @@ class ServiceBinding(object):
     def create(self, yaml):
         sbr_create_output = self.openshift.oc_apply(yaml)
```
pmacik, line 11: Please, replace this line with
```go
sbr_create_output = self.attempt_to_create(yaml)
```
to avoid duplicating pieces of code.

test/acceptance/features/steps/steps.py
```diff
@@ -425,3 +426,17 @@ def etc_cluster_is_running(context, etcd_name):
...
+
+
+@then(u'Openshift throws an error "{err_msg}"')
```
pmacik, line 439: Let's be more generic here in the step definition... since we are actually asserting the content of `context.expected_error` here, the step should rather be something like `Error message "{err_msg}" is thrown` instead of `Openshift throws error...`. 

So please rename the stes to the `Error message "{err_msg}" is thrown` so that it can be re-used by other scenarios. Where the previous steps would set `context.expected_error` to be checked later with this step.

pmacik (changes requested):

test/acceptance/features/bindAppToService.feature
pmacik: @pedjak We can be even more generic here, see https://github.com/redhat-developer/service-binding-operator/pull/698#discussion_r498217503

test/acceptance/features/bindAppToService.feature
pmacik: The new step has a different assertion expectations. The original step expect no error to be thrown, while the new step expect error to be thrown... so the behavior of each variant is different, therefore it should be split.

test/acceptance/features/bindAppToService.feature
pmacik: +1 on checking SB is not persisted.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Ahh, you got a point here, I'll update it.

pratikjagrut: Done.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Done.

test/acceptance/features/steps/steps.py
pratikjagrut: Done.

test/acceptance/features/steps/service_binding.py
pratikjagrut: Done.

test/acceptance/features/steps/steps.py
pratikjagrut: Done.

pratikjagrut: PTAL

test/acceptance/features/bindAppToService.feature
pratikjagrut: I made some changes PTAL.

pkg/apis/operators/v1alpha1/servicebinding_types.go
```diff
@@ -112,7 +113,8 @@ type ServiceBinding struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   ServiceBindingSpec   `json:"spec,omitempty"`
+	// +required
+	Spec   ServiceBindingSpec   `json:"spec"`
```
pedjak, line 117: is this change really necessary in this PR?

test/acceptance/features/bindAppToService.feature
```diff
@@ -6,9 +6,8 @@ Feature: Bind an application to a service
     Background:
         Given Namespace [TEST_NAMESPACE] is used
         * Service Binding Operator is running
-        * PostgreSQL DB operator is installed
+        # * PostgreSQL DB operator is installed
```
pedjak, line 9: please remove if not needed.

test/acceptance/features/bindAppToService.feature
```diff
@@ -586,3 +585,86 @@ Feature: Bind an application to a service
         And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-backend-a" should be changed to "False"
         And Secret "binding-request-backend-a" contains "BACKEND_READY" key with value "true"
         And Secret "binding-request-backend-a" contains "BACKEND_HOST" key with value "example.common"
+
+        @negative1
```
pedjak, line 589: `@negative1` is the new annotation?

test/acceptance/features/bindAppToService.feature
```diff
@@ -586,3 +585,86 @@ Feature: Bind an application to a service
...
+            """
+            apiVersion: operators.coreos.com/v1alpha1
+            kind: ServiceBinding
+            metadata:
+                name: binding-request-without-spec
+            """
+        Then Error message "spec: Required value" is thrown
+        And Service Binding "binding-request-without-spec" is not persistent in the cluster
+
+        @negative1
```
pedjak, line 601: same as above

test/acceptance/features/bindAppToService.feature
```diff
@@ -586,3 +585,86 @@ Feature: Bind an application to a service
...
+            kind: ServiceBinding
+            metadata:
+                name: binding-request-empty-services
+            spec:
+                services:
+            """
+        Then Error message "invalid: spec.services: Invalid value: \"null\"" is thrown
+        And Service Binding "binding-request-empty-services" is not persistent in the cluster
+
+        @negative1
```
pedjak, line 628: same as above

test/acceptance/features/bindAppToService.feature
```diff
@@ -586,3 +585,86 @@ Feature: Bind an application to a service
...
+            spec:
+                services:
+                -   name: backend-demo
+            """
+        Then Error message "spec.services.group: Required value" is thrown
+        And Error message "spec.services.kind: Required value" is thrown
+        And Error message "spec.services.version: Required value" is thrown
+        And Service Binding "binding-request-without-gvk" is not persistent in the cluster
+
+        @negative1
```
pedjak, line 645: same as above

test/acceptance/features/bindAppToService.feature
```diff
@@ -586,3 +585,86 @@ Feature: Bind an application to a service
...
+            metadata:
+                name: persisting-binding-request
+            spec:
+                services:
+                -   group: service.example.com
+                    version: v1
+                    kind: Backserv
+                    name: demo-backserv-cr-2
+            """
+        Then Get "{.metadata.resourceVersion}" of "servicebinding" "persisting-binding-request"
```
pedjak, line 660: what is the purpose of this step?

test/acceptance/features/bindAppToService.feature
```diff
@@ -586,3 +585,86 @@ Feature: Bind an application to a service
...
+                services:
+                -   name: backend-demo
+            """
+        Then Error message "spec.services.group: Required value" is thrown
+        And Error message "spec.services.kind: Required value" is thrown
+        And Error message "spec.services.version: Required value" is thrown
+        And Service Binding "binding-request-without-gvk" is not persistent in the cluster
+
+        @negative1
+        Scenario: Configuring exsting service binding with invalid spec configuration is not allowed
```
pedjak, line 646: better reads as "Removing services field from existing serivce binding is not allowed"

test/acceptance/features/bindAppToService.feature
pratikjagrut: This step will save the resourceVersion of first CR in context and we'll later verify if the resourceVersion is same or changed to validate the unchanged state of service binding.

pkg/apis/operators/v1alpha1/servicebinding_types.go
pratikjagrut: Yes, I tried creating SBR without spec and I was successful in doing so. Making services required will only be applicable if the spec is present in SBR if spec itself is not present then Kubernetes will not object to it.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Sorry commenting it is a mistake. Reverting it back.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Removed.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Done.

pkg/apis/operators/v1alpha1/servicebinding_types.go
pratikjagrut: Do you feel like I should address it another PR?

pedjak: Either that or change the scope of the PR updating its description. If you decide to keep this change, we should then add additional acceptance tests.

pkg/controller/servicebinding/reconciler.go
```diff
@@ -78,10 +78,10 @@ func (r *reconciler) getServiceBinding(
 	return sbr, nil
 }
 
-func servicesOrEmptyList(maybeServices *[]v1alpha1.Service) []v1alpha1.Service {
+func servicesOrEmptyList(maybeServices []v1alpha1.Service) []v1alpha1.Service {
```
pedjak, line 81: this function is now pointless and can be removed fully, because `maybeServices` is not pointer anymore, this cannot be `nil`.

test/acceptance/features/bindAppToService.feature
```diff
@@ -670,3 +670,86 @@ Feature: Bind an application to a service
...
+            kind: ServiceBinding
+            metadata:
+                name: binding-request-without-gvk
+            spec:
+                services:
+                -   name: backend-demo
+            """
+        Then Error message "spec.services.group: Required value" is thrown
+        And Error message "spec.services.kind: Required value" is thrown
+        And Error message "spec.services.version: Required value" is thrown
```
pedjak, line 727: Does user see 3 error messages? or it is just single multiline message?

test/acceptance/features/bindAppToService.feature
pedjak: This smells like abstraction leakage, i.e. implementation details. User should not know anything about that. The only thing what users sees is an error after trying to update the service binding. Think about the scenario as list of instruction we give user/tester so he is able to verify the scenario. Whatever we do in the line 732 should gather enough information (for example this step could be a part of that). so that we can perform later steps.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Single multiline error message.

pratikjagrut: Should I coalesce these in one multiline error step?

pkg/apis/operators/v1alpha1/servicebinding_types.go
isutton: I'd say move this to another PR.

pkg/controller/servicebinding/reconciler.go
pratikjagrut: Removed.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Redesigned my approach PTAL.

pkg/apis/operators/v1alpha1/servicebinding_types.go
pratikjagrut: Okay, I'll create another PR for this.

pratikjagrut: Raised another PR for this: https://github.com/redhat-developer/service-binding-operator/pull/712 PTAL

pkg/controller/servicebinding/mapper.go
```diff
@@ -42,7 +42,7 @@ func isSecret(obj runtime.Object) bool {
 
 // isSBRService checks whether the given obj is a service in given sbr.
 func isSBRService(sbr *v1alpha1.ServiceBinding, obj runtime.Object) bool {
-	services := servicesOrEmptyList(sbr.Spec.Services)
+	services := sbr.Spec.Services
```
pedjak, line 45: do we really need `services` var? We can iterate directly `sbr.Spec.Services` in for loop?

test/acceptance/features/bindAppToService.feature
pedjak: What other messages? @pmacik wdyt?

pedjak (commented):
In general looks good, need to be squashed.

test/acceptance/features/bindAppToService.feature
pratikjagrut: All the scenarios related to this PR uses the same step for error validation step, which produces single line error. So to confirm the reusability of steps I'll need to update those scenarios as well.

pkg/controller/servicebinding/mapper.go
pratikjagrut: No, we don't need it, I'll update it.

pratikjagrut: Done.

pedjak (approved):

PR #704, opened 2020-10-01
Trigger reconcile on updates to services' metadata.Annotations specific to service bindings

Avni-Sharma:
Refer https://issues.redhat.com/browse/APPSVC-709

As an extension to https://github.com/redhat-developer/service-binding-operator/pull/597 the new changes can now trigger reconcile on changes to services' metadata.Annotations

Run
Test - `make test-acceptance`

### DIFF
--- a/pkg/controller/servicebinding/sbrcontroller.go
+++ b/pkg/controller/servicebinding/sbrcontroller.go
@@ -2,6 +2,7 @@ package servicebinding
 
 import (
 	"os"
+	"reflect"
 	"strings"
 
 	olmv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
@@ -20,6 +21,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/redhat-developer/service-binding-operator/pkg/apis/operators/v1alpha1"
+	"github.com/redhat-developer/service-binding-operator/pkg/controller/servicebinding/binding"
 	"github.com/redhat-developer/service-binding-operator/pkg/log"
 )
 
@@ -106,6 +108,16 @@ func isOfKind(obj runtime.Object, kind string) bool {
 	return strings.EqualFold(obj.GetObjectKind().GroupVersionKind().Kind, kind)
 }
 
+// check whether annotations specify bindings
+func isAnnotationValid(ann map[string]string) bool {
+	for k := range ann {
+		if strings.HasPrefix(k, binding.AnnotationPrefix) {
+			return true
+		}
+	}
+	return false
+}
+
 // updateEvent returns a predicate handler function.
 func updateFunc(logger *log.Log) func(updateEvent event.UpdateEvent) bool {
 	return func(e event.UpdateEvent) bool {
@@ -141,8 +153,13 @@ func updateFunc(logger *log.Log) func(updateEvent event.UpdateEvent) bool {
 		} else {
 			statusAreEqual = statusComparison.Success
 		}
-
-		shouldReconcile := !specsAreEqual || !statusAreEqual
+		annotationsAreEqual := true
+		annOld := e.MetaOld.GetAnnotations()
+		annNew := e.MetaNew.GetAnnotations()
+		if isAnnotationValid(annOld) && isAnnotationValid(annNew) {
+			annotationsAreEqual = reflect.DeepEqual(annOld, annNew)
+		}
+		shouldReconcile := !specsAreEqual || !statusAreEqual || !annotationsAreEqual
 
 		logger.Debug("Resource update event received",
 			"GVK", e.ObjectNew.GetObjectKind().GroupVersionKind(),
--- a/test/acceptance/features/bindAppToService.feature
+++ b/test/acceptance/features/bindAppToService.feature
@@ -804,3 +804,105 @@ Feature: Bind an application to a service
             """
         Then Error message "spec: Required value" is thrown
         And Service Binding "binding-request-remove-spec" is not updated
+
+
+   Scenario: Backend Service metadata annotations update for service bindings gets propagated to the binding secret
+        Given OLM Operator "backend" is running
+        * The Custom Resource is present
+            """
+            apiVersion: "stable.example.com/v1"
+            kind: Backend
+            metadata:
+                name: backend-demo-2
+                annotations:
+                    service.binding/host: path={.spec.host}
+            spec:
+                host: example.com
+            status:
+                ready:true
+            """
+        * Service Binding is applied
+            """
+            apiVersion: operators.coreos.com/v1alpha1
+            kind: ServiceBinding
+            metadata:
+                name: binding-request-backend-ann-sb
+            spec:
+                services:
+                -   group: stable.example.com
+                    version: v1
+                    kind: Backend
+                    name: backend-demo-2
+            """
+        And Secret "binding-request-backend-ann-sb" contains "BACKEND_HOST" key with value "example.com"
+        And Secret "binding-request-backend-ann-sb" does not contain "BACKEND_READY"
+        # Backend metadata.annotations for service binding is updated
+        When The Custom Resource is present
+            """
+            apiVersion: "stable.example.com/v1"
+            kind: Backend
+            metadata:
+                name: backend-demo-2
+                annotations:
+                    service.binding/host: path={.spec.host}
+                    service.binding/ready: path={.status.ready}
+            spec:
+                host: example.com
+            status:
+                ready: true
+            """
+        Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-request-backend-ann-sb" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-backend-ann-sb" should be changed to "False"
+        And Secret "binding-request-backend-ann-sb" contains "BACKEND_READY" key with value "true"
+        And Secret "binding-request-backend-ann-sb" contains "BACKEND_HOST" key with value "example.com"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").reason" of Service Binding "binding-request-backend-ann-sb" should be changed to "EmptyApplication"
+
+
+   @negative
+   Scenario: Backend Service metadata annotations update not specific to service bindings does not get propagated to the binding secret
+        Given OLM Operator "backend" is running
+        * The Custom Resource is present
+            """
+            apiVersion: "stable.example.com/v1"
+            kind: Backend
+            metadata:
+                name: backend-demo-3
+                annotations:
+                    host : "demo.com"
+            spec:
+                host: example.com
+            status:
+                ready:true
+            """
+        * Service Binding is applied
+            """
+            apiVersion: operators.coreos.com/v1alpha1
+            kind: ServiceBinding
+            metadata:
+                name: binding-request-backend-ann
+            spec:
+                services:
+                -   group: stable.example.com
+                    version: v1
+                    kind: Backend
+                    name: backend-demo-3
+            """
+        And jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-request-backend-ann" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-backend-ann" should be changed to "False"
+        And Secret "binding-request-backend-ann" is empty
+        And jq ".status.conditions[] | select(.type=="InjectionReady").reason" of Service Binding "binding-request-backend-ann" should be changed to "EmptyApplication"
+        # Backend metadata.annotations not pertaining to service binding is updated
+        When The Custom Resource is present
+            """
+            apiVersion: "stable.example.com/v1"
+            kind: Backend
+            metadata:
+                name: backend-demo-3
+                annotations:
+                    host : "example.common"
+            spec:
+                host: example.com
+            status:
+                ready: true
+            """
+        Then Secret "binding-request-backend-ann" is empty
--- a/test/acceptance/features/steps/steps.py
+++ b/test/acceptance/features/steps/steps.py
@@ -219,6 +219,7 @@ def then_app_is_connected_to_db(context, db_name):
 
 
 # STEP
+@given(u'jsonpath "{json_path}" of Service Binding "{sbr_name}" should be changed to "{json_value_regex}"')
 @then(u'jsonpath "{json_path}" of Service Binding "{sbr_name}" should be changed to "{json_value_regex}"')
 def then_sbo_jsonpath_is(context, json_path, sbr_name, json_value_regex):
     openshift = Openshift()
@@ -436,3 +437,24 @@ def validate_persistent_sb(context, sb_name):
         assert True
     else:
         assert False, "Service Binding got updated"
+
+
+@given(u'Secret "{secret_name}" does not contain "{key}"')
+def check_secret_key(context, secret_name, key):
+    openshift = Openshift()
+    json_path = f'{{.data.{key}}}'
+    polling2.poll(lambda: openshift.get_resource_info_by_jsonpath("secrets", secret_name, context.namespace.name,
+                                                                  json_path) == "",
+                  step=5, timeout=120, ignore_exceptions=(binascii.Error,))
+
+
+@step(u'Secret "{secret_name}" is empty')
+def validate_secret_empty(context, secret_name):
+    openshift = Openshift()
+    try:
+        polling2.poll(lambda: json.loads(
+            openshift.get_resource_info_by_jq("secrets", secret_name, context.namespace.name, ".data",
+                                              wait=False)) == "null", step=5, timeout=20,
+                      ignore_exceptions=(json.JSONDecodeError,))
+    except polling2.TimeoutException:
+        pass
--- a/pkg/controller/servicebinding/sbrcontroller_test.go
+++ b/pkg/controller/servicebinding/sbrcontroller_test.go
@@ -146,6 +146,63 @@ func TestSBRControllerBuildGVKPredicate(t *testing.T) {
 				AvailableReplicas: 3,
 			},
 		}
+		deploymentD := &appsv1.Deployment{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "Deployment",
+				APIVersion: "apps/v1",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Annotations: map[string]string{
+					"service": "binding",
+				},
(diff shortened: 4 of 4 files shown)


pkg/controller/servicebinding/sbrcontroller.go
```diff
@@ -43,8 +44,8 @@ var _ ResourceWatcher = (*sbrController)(nil)
 // controllerName common name of this controller
 const controllerName = "servicebinding-controller"
 
-// compareObjectFields compares a nested field of two given objects.
-func compareObjectFields(objOld, objNew runtime.Object, fields ...string) (bool, error) {
+// compareRuntimeObjectFields compares a nested field of two given runtime objects.
+func compareRuntimeObjectFields(objOld, objNew runtime.Object, fields ...string) (bool, error) {
```
pedjak, line 48: why do we need to rename this function?

Avni-Sharma: I had planned to use this function for ObjectMeta as well, but then later realized that only Runtime Object is the param for this function. So to avoid confusion in the future I added the Runtime as a prefix to the Object.

pkg/controller/servicebinding/sbrcontroller.go
```diff
@@ -113,27 +114,29 @@ func updateFunc(logger *log.Log) func(updateEvent event.UpdateEvent) bool {
...
 		if err != nil {
 			logger.Error(err, "error comparing object's status fields: %s", err.Error())
 			return false
 		}
-		shouldReconcile := !specsAreEqual || !statusAreEqual
-		logger.Debug("Resource update event received", "GVK", e.ObjectNew.GetObjectKind(), "spec changed", !specsAreEqual, "status changed", !statusAreEqual, "should reconcile", shouldReconcile)
-
+		annOld := e.MetaOld.GetAnnotations()
+		annNew := e.MetaNew.GetAnnotations()
+		annotationsAreEqual := reflect.DeepEqual(annOld, annNew)
```
pedjak, line 137: we should trigger reconcile only if relevant annotations are changed, i.e. annotations specifying bindings.

Avni-Sharma: Done https://github.com/redhat-developer/service-binding-operator/pull/704/commits/9ec2a90e4799408910298274c6fca5b1a1f29d4c

test/acceptance/features/bindAppToService.feature
```diff
@@ -585,3 +585,55 @@ Feature: Bind an application to a service
         And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-backend-a" should be changed to "False"
         And Secret "binding-request-backend-a" contains "BACKEND_READY" key with value "true"
         And Secret "binding-request-backend-a" contains "BACKEND_HOST" key with value "example.common"
+
+
+   Scenario: Backend Service metadata annotations update gets propagated to the binding secret
```
pedjak, line 590: so to verify that only changes on service binding annotations trigger reconcile, please add a negative scenario where we show that reconcile is not triggered if non-service binding annotations are updated/added/removed.

Avni-Sharma: Updated https://github.com/redhat-developer/service-binding-operator/pull/704/commits/3f8111378c1613385009313ad144a92750b0220f

test/acceptance/features/bindAppToService.feature
```diff
@@ -585,3 +585,103 @@ Feature: Bind an application to a service
...
+            spec:
+                services:
+                -   group: stable.example.com
+                    version: v1
+                    kind: Backend
+                    name: backend-demo-3
+            """
+        Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-request-backend-ann" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-backend-ann" should be changed to "False"
+        And Secret "binding-request-backend-ann" contains "BACKEND_HOST" key with value ""
```
pedjak, line 672: why should secret contain at all `BACKEND_HOST` key? If I understand correctly, that should not be the case.

pkg/controller/servicebinding/sbrcontroller.go
```diff
@@ -106,34 +108,49 @@ func isOfKind(obj runtime.Object, kind string) bool {
...
 		}
-		shouldReconcile := !specsAreEqual || !statusAreEqual
-		logger.Debug("Resource update event received", "GVK", e.ObjectNew.GetObjectKind(), "spec changed", !specsAreEqual, "status changed", !statusAreEqual, "should reconcile", shouldReconcile)
-
+		annotationsAreEqual := true
+		annOld := e.MetaOld.GetAnnotations()
+		annNew := e.MetaNew.GetAnnotations()
+		if isAnnotationValid(annOld) && isAnnotationValid(annNew) {
+			annotationsAreEqual = reflect.DeepEqual(annOld, annNew)
+		}
```
isutton, line 151: Could you elaborate what are you trying to achieve in here? 

My interpretation is that annotations are deeply compared if and only if both annotation maps have at least one valid service binding annotation; if this is indeed true, then both annotation maps are deeply compared and, if any change is identified by `reflect.DeepEqual`, the resource will be reconciled.

I expected to read here something along the following lines:

```go
annotationsAreEqual, err := compareAnnotations(e.MetaOld.GetAnnotations(), e.MetaNew.GetAnnotations())
if err != nil { ... }
```
`compareAnnotations` could then perform, such as `compareRuntimeObjectFields` the appropriate comparison for annotations and return an error in situations the comparison could not be performed.

Additionally, having a broader API surface with reasonable tests increase the understanding the reasons behind a failed comparison.

isutton (changes requested):

test/acceptance/features/bindAppToService.feature
Avni-Sharma: "BACKEND_HOST" key with value "" means that the key also does not exist. If the key exists and the value does not exist then the value is `<no-value>` whereas when both key and value are not present then for whatever key we put we get a value as `""` So this step means that there is no key and value .

pedjak: IMHO, this interpretation is confusing, especially for users. If a key does not exist in the secret, we should state that in the appropriate step. For example:

```
Secret XXX does not contain key KEY
```

Please implement such a step.

pkg/controller/servicebinding/sbrcontroller.go
Avni-Sharma: runtime.Object is being used as parameters for `compareRuntimeObjectFields`  and we need meta.Object. Hence I didn't use that function.
I can have a compareAnnotations function performing the reflect.DeepEqual. Would that suffice?

test/acceptance/features/bindAppToService.feature
Avni-Sharma: updated

test/acceptance/features/bindAppToService.feature
```diff
@@ -804,3 +804,103 @@ Feature: Bind an application to a service
...
+            metadata:
+                name: binding-request-backend-ann-sb
+            spec:
+                services:
+                -   group: stable.example.com
+                    version: v1
+                    kind: Backend
+                    name: backend-demo-2
+            """
+        Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-request-backend-ann-sb" should be changed to "True"
```
pedjak, line 837: use `And` because we are still in the preparation phase here. I would even say that is not really required to check the content of the secret, because we are interested to see what is going to happen after update.

test/acceptance/features/bindAppToService.feature
```diff
@@ -804,3 +804,103 @@ Feature: Bind an application to a service
...
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-backend-ann-sb" should be changed to "False"
+        And Secret "binding-request-backend-ann-sb" contains "BACKEND_HOST" key with value "example.com"
```
pedjak, line 839: see above, IMHO it can be removed here.

test/acceptance/features/bindAppToService.feature
```diff
@@ -804,3 +804,103 @@ Feature: Bind an application to a service
...
+                annotations:
+                    service.binding/host: path={.spec.host}
+                    service.binding/ready: path={.status.ready}
+            spec:
+                host: example.com
+            status:
+                ready: true
+            """
+        Then Secret "binding-request-backend-ann-sb" contains "BACKEND_READY" key with value "true"
+        And Secret "binding-request-backend-ann-sb" contains "BACKEND_HOST" key with value "example.com"
```
pedjak, line 857: and on top of this we should check SB conditions.

test/acceptance/features/bindAppToService.feature
```diff
@@ -804,3 +804,103 @@ Feature: Bind an application to a service
...
+            metadata:
+                name: binding-request-backend-ann
+            spec:
+                services:
+                -   group: stable.example.com
+                    version: v1
+                    kind: Backend
+                    name: backend-demo-3
+            """
+        Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-request-backend-ann" should be changed to "True"
```
pedjak, line 889: Use `And`

test/acceptance/features/bindAppToService.feature
```diff
@@ -804,3 +804,103 @@ Feature: Bind an application to a service
...
+   Scenario: Backend Service metadata annotations update not specific to service bindings does not get propagated to the binding secret
+        Given OLM Operator "backend" is running
+        * The Custom Resource is present
+            """
+            apiVersion: "stable.example.com/v1"
+            kind: Backend
+            metadata:
+                name: backend-demo-3
+                annotations:
+                    host : 'binding:env:attribute'
```
pedjak, line 870: the value of the annotation could be confusing for users, because looks similar to the old annotation format, not in the use anymore. Please change to something else.

test/acceptance/features/bindAppToService.feature
```diff
@@ -804,3 +804,103 @@ Feature: Bind an application to a service
...
+            spec:
+                services:
+                -   group: stable.example.com
+                    version: v1
+                    kind: Backend
+                    name: backend-demo-3
+            """
+        Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-request-backend-ann" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-backend-ann" should be changed to "False"
+        And Secret "binding-request-backend-ann" does not contain "BACKEND_HOST"
```
pedjak, line 891: why do we need this step?

Avni-Sharma: In the scenario where this step is used, there we are applying SBR 2 times. Once with a non specific service-binding-annotation and then update to the similar non specific service binding annotation to check that both the times the secret key should not contain anything. This scenario was added for https://github.com/redhat-developer/service-binding-operator/pull/704#discussion_r498896479

test/acceptance/features/bindAppToService.feature
```diff
@@ -804,3 +804,103 @@ Feature: Bind an application to a service
...
+            metadata:
+                name: backend-demo-3
+                annotations:
+                    host : "example.common"
+            spec:
+                host: example.com
+            status:
+                ready: true
+            """
+        Then Secret "binding-request-backend-ann" does not contain "BACKEND_HOST"
```
pedjak, line 906: why is this step necessary? how user can detect if SB did not trigger reconciliation? The scenario is about that.

test/acceptance/features/bindAppToService.feature
Avni-Sharma: hence this shows that when annotations are updated which are non specific to binding, then retrigger does not occur. And to check that we check the content of secret again

test/acceptance/features/bindAppToService.feature
Avni-Sharma: refer https://github.com/redhat-developer/service-binding-operator/pull/704#discussion_r501084165 and https://github.com/redhat-developer/service-binding-operator/pull/704#discussion_r501084957

test/acceptance/features/bindAppToService.feature
Avni-Sharma: Updated

test/acceptance/features/bindAppToService.feature
Avni-Sharma: Retriggering of the reconcile can be seen from the changes in the secret's data. Something like what was followed here https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToService.feature#L277  We are concerned with the secret contents to be updated , not exactly how the app users will use the secret. hence for if we are interested to see what is going to happen after update, it means to check the secret contents. Please let me know what you suggest here. According to my this checking is required.

test/acceptance/features/bindAppToService.feature
Avni-Sharma: ?

test/acceptance/features/bindAppToService.feature
pedjak: My point was that these checks could be dropped because you do the final check at the end - what happened inbetween is not super important - only what happens at the end counts :) but I am ok if you want to leave them.

test/acceptance/features/bindAppToService.feature
pedjak: Everything what comes before `When` step is just establishing test fixture - in Given-When-Then format. Therefore, this step should start with `And` like the other one before it.

test/acceptance/features/bindAppToService.feature
pedjak: ah, so secret should not contain anything - but the steps says `Secret "binding-request-backend-ann" does not contain "BACKEND_HOST`. Hence, if it should not contain anything, then the step should be

```
Secret "binding-request-backend-ann" is empty
```

The other question is - should secret be created at all? Should the application deployment be modified at all?

test/acceptance/features/bindAppToService.feature
```diff
@@ -804,3 +804,105 @@ Feature: Bind an application to a service
...
+                name: binding-request-backend-ann-sb
+            spec:
+                services:
+                -   group: stable.example.com
+                    version: v1
+                    kind: Backend
+                    name: backend-demo-2
+            """
+        Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-request-backend-ann-sb" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-backend-ann-sb" should be changed to "False"
```
pmacik, line 838: If you expect some negative outcome or a failure (like here where you expect the status to be `False`) you should also make sure the reason (especially with multiple options), so that any other failure do not pass the test.
Please add a step such as the following (or one that is appropriate) after the check that the status is `False` to verify the expected reason.:

```feature
And jq ".status.conditions[] | select(.type=="InjectionReady").reason" of Service Binding "binding-request-backend-ann-sb" should be changed to "EmptyApplication"
```

test/acceptance/features/bindAppToService.feature
```diff
@@ -804,3 +804,105 @@ Feature: Bind an application to a service
...
+                annotations:
+                    service.binding/host: path={.spec.host}
+                    service.binding/ready: path={.status.ready}
+            spec:
+                host: example.com
+            status:
+                ready: true
+            """
+        Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-request-backend-ann-sb" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-backend-ann-sb" should be changed to "False"
```
pmacik, line 857: Same as above...

Please add a step such as the following (or one that is appropriate) after the check that the status is `False` to verify the expected reason:

```feature
And jq ".status.conditions[] | select(.type=="InjectionReady").reason" of Service Binding "binding-request-backend-ann-sb" should be changed to "EmptyApplication"
```

test/acceptance/features/bindAppToService.feature
```diff
@@ -804,3 +804,105 @@ Feature: Bind an application to a service
...
+                name: binding-request-backend-ann
+            spec:
+                services:
+                -   group: stable.example.com
+                    version: v1
+                    kind: Backend
+                    name: backend-demo-3
+            """
+        Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-request-backend-ann" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-backend-ann" should be changed to "False"
```
pmacik, line 892: Same as above...

Please add a step such as the following (or one that is appropriate) after the check that the status is `False` to verify the expected reason.:

```feature
And jq ".status.conditions[] | select(.type=="InjectionReady").reason" of Service Binding "binding-request-backend-ann" should be changed to "EmptyApplication"
```

pmacik (changes requested):

test/acceptance/features/bindAppToService.feature
Avni-Sharma: Currently, if SBR is there then the secret is created even with data 0.

test/acceptance/features/bindAppToService.feature
Avni-Sharma: Updated

test/acceptance/features/bindAppToService.feature
Avni-Sharma: Updated

test/acceptance/features/bindAppToService.feature
Avni-Sharma: updated

test/acceptance/features/steps/steps.py
```diff
@@ -436,3 +437,25 @@ def validate_persistent_sb(context, sb_name):
...
+    openshift = Openshift()
+    json_path = f'{{.data.{key}}}'
+    polling2.poll(lambda: openshift.get_resource_info_by_jsonpath("secrets", secret_name, context.namespace.name,
+                                                                  json_path) == "",
+                  step=5, timeout=120, ignore_exceptions=(binascii.Error,))
+
+
+@given(u'Secret "{secret_name}" is empty')
+@then(u'Secret "{secret_name}" is empty')
+def validate_secret_empty(context, secret_name):
```
baijum, line 453: Instead of two decorators (`given` and `then`), you can use `step` as the decorator.

Avni-Sharma: https://github.com/redhat-developer/service-binding-operator/pull/704/commits/13bd37716c3aab868aa8ae9729e83d038405f96a

PR #712, opened 2020-10-05
Make spec a required field in SBR

pratikjagrut:
### Motivation

Making spec a required field
https://issues.redhat.com/browse/APPSVC-711

### DIFF
--- a/test/acceptance/features/steps/steps.py
+++ b/test/acceptance/features/steps/steps.py
@@ -180,7 +180,7 @@ def sbr_is_applied(context):
     metadata_name = re.sub(r'.*: ', '', re.search(r'name: .*', sbr_yaml).group(0))
     context.sbr_name = metadata_name
     sbr = ServiceBinding()
-    if context.__contains__("application") and context.__contains__("application_type"):
+    if "application" in context and "application_type" in context:
         application = context.application
         if context.application_type == "nodejs":
             context.application_original_generation = application.get_observed_generation()
@@ -403,13 +403,11 @@ def etc_cluster_is_running(context, etcd_name):
 def invalid_sbr_is_applied(context):
     sbr = ServiceBinding()
     # Get resource version of sbr if sbr is available
-    if context.__contains__("sbr_name"):
+    if "sbr_name" in context:
         json_path = "{.metadata.resourceVersion}"
         rv = sbr.get_servicebinding_info_by_jsonpath(context.sbr_name, context.namespace.name, json_path)
         context.resource_version = rv
-    yaml = context.text
-    output = sbr.attempt_to_create(yaml)
-    context.expected_error = output
+    context.expected_error = sbr.attempt_to_create(context.text)
 
 
 @then(u'Error message "{err_msg}" is thrown')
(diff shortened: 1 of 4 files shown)


test/acceptance/features/steps/steps.py
```diff
@@ -394,3 +397,40 @@ def etc_cluster_is_running(context, etcd_name):
         print("etcd cluster not present, creating etcd cluster")
         assert etcd_cluster.create() is True, "etcd cluster is not created"
         assert etcd_cluster.is_present() is True, "etcd cluster is not present"
+
+
+@when(u'Invalid Service Binding is applied')
+def invalid_sbr_is_applied(context):
+    sbr = ServiceBinding()
+    # Get resource version of sbr if sbr is available
+    if context.__contains__("sbr_name"):
```
baijum, line 406: I can see many uses of the special method `__contains__`. Is there any problem to use `in` operator?

Ref. https://stackoverflow.com/questions/1964934/what-does-contains-do-what-can-call-contains-function

pratikjagrut: I didn't really think about it `__contains__` was already in use so I simply continued it but will look into it and if `in` works will replace `__contains__` with `in`.

pratikjagrut: Hey updated as you suggested. PTAL.

test/acceptance/features/steps/steps.py
```diff
@@ -394,3 +397,40 @@ def etc_cluster_is_running(context, etcd_name):
...
+def invalid_sbr_is_applied(context):
+    sbr = ServiceBinding()
+    # Get resource version of sbr if sbr is available
+    if "sbr_name" in context:
+        json_path = "{.metadata.resourceVersion}"
+        rv = sbr.get_servicebinding_info_by_jsonpath(context.sbr_name, context.namespace.name, json_path)
+        context.resource_version = rv
+    yaml = context.text
+    output = sbr.attempt_to_create(yaml)
+    context.expected_error = output
```
pedjak, line 412: we can join these two lines:

```
context.expect_error = sbr.attemp_to_create(yaml)
```

test/acceptance/features/steps/steps.py
```diff
@@ -394,3 +397,40 @@ def etc_cluster_is_running(context, etcd_name):
...
+
+
+@then(u'Error message "{err_msg}" is thrown')
+def validate_error(context, err_msg):
+    search = re.search(rf'.*{err_msg}.*', context.expected_error)
+    assert search is not None, "Actual error: {context.expected_error}, Expected error: {err_msg}"
+
+
+@then(u'Service Binding "{sb_name}" is not persistent in the cluster')
+def get_sb(context, sb_name):
```
pedjak, line 422: can we update the function name to reflect what is going on? the function is not about getting  service binding - it is about checking if the given SB exists in the cluster.

test/acceptance/features/steps/steps.py
```diff
@@ -394,3 +397,40 @@ def etc_cluster_is_running(context, etcd_name):
...
+    openshift = Openshift()
+    output = openshift.search_resource_in_namespace("servicebindings", sb_name, context.namespace.name)
+    assert output is None, "Service Binding {sb_name} is present in namespace {context.namespace.name}"
+
+
+@then(u'Service Binding "{sb_name}" is not updated')
+def get_persistent_sb(context, sb_name):
```
pedjak, line 429: same as above.

pedjak (changes requested):
see the comments.

test/acceptance/features/steps/steps.py
pratikjagrut: updated.

test/acceptance/features/steps/steps.py
pratikjagrut: Updated PTAL.

test/acceptance/features/steps/steps.py
pratikjagrut: Updated PTAL.

pratikjagrut:
Forced push for squashing the commits.

PR #713, opened 2020-10-05
Use CRD name instead of hard-coded "sbr" or "sbrs" in hack/remove-sbr-finalizers.sh

pmacik:
### Motivation

Jira: [APPSVC-726](https://issues.redhat.com/browse/APPSVC-726)

Currently the `hack/remove-sbr-finalizers.sh` script uses hardcoded short names for Service Binding resources `sbr`/`sbrs` that is not reliable - it fails sometimes to find the resource by that short name. And also it can change in the future.
 
### Changes

This PR changes the script to use exact name of the CRD instead of the short name.

### Testing

`make test-cleanup`

### DIFF
--- a/hack/remove-sbr-finalizers.sh
+++ b/hack/remove-sbr-finalizers.sh
@@ -7,11 +7,12 @@ else
 fi
 
 # Remove SBR finalizers if CRD exists
-[ -z "$(kubectl get -f deploy/crds/*crd.yaml -o jsonpath="{.metadata.name}" --ignore-not-found)" ] && exit 0
+CRD_NAME=$(kubectl get -f deploy/crds/*crd.yaml -o jsonpath="{.metadata.name}" --ignore-not-found)
+[ -z $CRD_NAME ] && exit 0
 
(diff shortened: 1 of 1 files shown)


hack/remove-sbr-finalizers.sh
```diff
@@ -7,11 +7,12 @@ else
 fi
 
 # Remove SBR finalizers if CRD exists
-[ -z "$(kubectl get -f deploy/crds/*crd.yaml -o jsonpath="{.metadata.name}" --ignore-not-found)" ] && exit 0
+CRD=$(kubectl get -f deploy/crds/*crd.yaml -o jsonpath="{.metadata.name}" --ignore-not-found)
+[ -z $CRD ] && exit 0
```
pedjak, line 11: `CRD` contains CRD name, not CRD itself. How about that we rename it ot `CRD_NAME`?

pedjak (approved):
nit: rename var for better reading.

pedjak:
Can you also update PR summary to reflect the exact change?

hack/remove-sbr-finalizers.sh
pmacik: Good point

PR #714, opened 2020-10-05
[acceptance-tests] Do not install any resource if SBO is remote.

pmacik:
### Motivation

Currently there are some resources (such as CRD, role, role-binding,...) that are installed for acceptance tests before the actual tests are execued. This should only by the case when SBO is running `local`. In case SBO is running `remote`ly (i.e. installed in OpenShift beforehand) all of the resources should be already there by the installation.

### Changes

This PR:
* Skips steps in Makefile that install resources
* Introduces `SBO_NAMESPACE` variable to specify in which namespace the remote SBO is running. 

### Testing

1. Install SBO via OperatorHub (`beta` channel)
2. Run  `TEST_ACCEPTANCE_START_SBO=remote SBO_NAMESPACE=openshift-operators make test-acceptance-smoke
`

### DIFF
--- a/test/acceptance/features/environment.py
+++ b/test/acceptance/features/environment.py
@@ -13,13 +13,24 @@ before_all(context), after_all(context)
 """
 
 import subprocess
-from pyshould import should
+import os
+
+
+def before_all(_context):
+    start_sbo = os.getenv("TEST_ACCEPTANCE_START_SBO")
+    assert start_sbo is not None, "TEST_ACCEPTANCE_START_SBO is not set. It should be one of local, remote or operator-hub"
+    assert start_sbo in {"local", "remote", "operator-hub"}, "TEST_ACCEPTANCE_START_SBO should be one of local, remote or operator-hub"
+
+    if start_sbo == "local":
+        assert not os.getenv("TEST_ACCEPTANCE_SBO_STARTED").startswith("FAILED"), "TEST_ACCEPTANCE_SBO_STARTED shoud not be FAILED."
+    elif start_sbo == "remote":
+        sbo_namespace = os.getenv("SBO_NAMESPACE")
+        assert sbo_namespace is not None, "SBO_NAMESPACE is required but not set."
+        _context.sbo_namespace = sbo_namespace
+    else:
+        assert False, f"TEST_ACCEPTANCE_START_SBO={start_sbo} is currently unsupported."
 
 
 def before_scenario(_context, _scenario):
-    print("Getting OC status before {} scenario".format(_scenario))
-    code, output = subprocess.getstatusoutput('oc get project default')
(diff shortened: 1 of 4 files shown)


test/acceptance/features/steps/steps.py
```diff
@@ -83,8 +83,11 @@ def sbo_is_running_in_namespace_from_env(context, operator_namespace_env):
 @given(sbo_is_running_step)
 @when(sbo_is_running_step)
 def sbo_is_running(context):
-    context.namespace | should_not.be_none.desc("Namespace set in context")
-    sbo_is_running_in_namespace(context, context.namespace.name)
+    if context.__contains__("sbo_namespace"):
```
baijum, line 86:
```suggestion
    if "sbo_namespace" in context:
```

pmacik: This is better, thx

test/acceptance/features/environment.py
```diff
@@ -13,13 +13,24 @@
...
+    assert start_sbo in {"local", "remote", "operator-hub"}, "TEST_ACCEPTANCE_START_SBO should be one of local, remote or operator-hub"
+
+    if start_sbo == "local":
+        assert not os.getenv("TEST_ACCEPTANCE_SBO_STARTED").startswith("FAILED"), "TEST_ACCEPTANCE_SBO_STARTED shoud not be FAILED."
+    elif start_sbo == "remote":
+        sbo_namespace = os.getenv("SBO_NAMESPACE")
+        assert sbo_namespace is not None, "SBO_NAMESPACE is required but not set."
+        _context.sbo_namespace = sbo_namespace
+    elif start_sbo == "operator-hub":
+        return False
```
pedjak, line 31: I would fail here - with some assertion or something, letting developer that `operator-hub` case is still unsupported.

test/acceptance/features/steps/servicebindingoperator.py
```diff
@@ -9,15 +9,17 @@ class Servicebindingoperator():
     name = ""
     namespace = ""
 
+    name_pattern = f"{name}.*"
```
pedjak, line 12: why we cannot use one exact value? reg ex will match against many namespaces here, is that really necessary? Why we cannot read it from env var `SBO_NAMESPACE`?

pmacik: the `name_pattern` is for name of resources, not namespace... the namespace is exact

test/acceptance/features/environment.py
pmacik: Good point

pmacik:
Acceptance tests failed for OpenShift CI cluster failure.
/retest

Issue #723, opened 2020-10-13
Controller doesn't reconcile when BackingService CR got deleted and recreated


qibobo:
steps to reproduce this issue:

I use [bindings.ibmcloud](https://github.com/IBM/cloud-operators/blob/master/config/crd/bases/ibmcloud.ibm.com_bindings.yaml) as the backing service. And I have added the annotations as below to the bindings CRD.

```
    service.binding/secretName: path={.status.secretName},objectType=Secret
    service.binding/serviceName: path={.spec.serviceName}
```

1 create the BackingServiceCR:

```
cat <<EOF | kubectl apply -f -
apiVersion: ibmcloud.ibm.com/v1
kind: Binding
metadata:
    name: thetranslator
spec:
    serviceName: thetranslator
EOF
```

2 create the knative service as the app

```
cat <<EOF | kubectl apply -f -
apiVersion: serving.knative.dev/v1
kind: Service 
metadata:
  name: knative-vcap
spec:
  template:
    metadata:
      annotations:
        autoscaling.knative.dev/minScale: "1"
    spec:
      containers:
        - image: docker.io/qibobo/go4autoscaler:latest
          ports:
          - containerPort: 8080
EOF
```

3 create the SBR
```
cat <<EOF | kubectl apply -f -
apiVersion: operators.coreos.com/v1alpha1
kind: ServiceBinding
metadata:
  name: binding-request-knative
spec:
  application:
    group: serving.knative.dev
    version: v1
    resource: services
    name: knative-vcap
  services:
  - group: ibmcloud.ibm.com
    version: v1alpha1
    kind: Binding
    name: thetranslator
    envVarPrefix: serviceprefix
    id: thetranslator
  customEnvVar:
    - name: VCAP_SERVICES
      value: '{"language-translator":[{"credentials":{{json .thetranslator.status.secretName}},"name":"thetranslator","plan":"standard"}]}'
    - name: TRANSLATOR_NAME
      value: "thetranslator"
    - name: URL
      value: '{{.thetranslator.status.secretName.url}}'
EOF
```

Now there are 2 secrets named thetranslator(generated by BackingService) and binding-request-knative(generated by SBO) and the content of the 2 secrets are equal.

5 delete and re-create the BackingServiceCR
```
kubectl delete binding.ibmcloud thetranslator
cat <<EOF | kubectl apply -f -
apiVersion: ibmcloud.ibm.com/v1
kind: Binding
metadata:
    name: thetranslator
[...]

PR #726, opened 2020-10-14
fix issue #723, trigger reconcile when backing services get deleted and recreated

qibobo:


cdlliuy:
@isutton  @Avni-Sharma  @baijum ,  can you have a look and share your comment?   If the changes  is fine for you , we will go ahead to add related test and so on.

Avni-Sharma:
Thanks for the PR folks! Perhaps we have a description to the PR, wherein we link the issue number, and describe the steps and results which we are trying to achieve, so that going further, we can use the same as a Scenario for our acceptance tests.

Avni-Sharma:
Follow up from community meeting- acceptance tests steps would be added to this PR :rocket:

pedjak (changes requested):
This PR need acceptance test(s) demonstrating the expected behaviour.

qibobo:
In our scenario, when the backing service is deleted and then recreated, the secret belongs to it will change, so the SBO must reconcile to pick the new secret. In current acceptance tests, there is no one for this scenario.
Below is a method maybe work.
we can define a CR as below that has a status.secretName which is the name of secret that the Backend owns. 
```
            apiVersion: "stable.example.com/v1"
            kind: Backend
            metadata:
                name: backend-demo
            spec:
                host: example.common
            status:
               secretName: thename
```

For this PR, the steps of acceptance can be: 
```
1 create the Backend and the secret1
2 create the SBR
3 assert that the SBO create a secret that has the same content as secret1
4 delete Backend, modify the secret1 content and create Backend again
5 assert that the SBO create a new secret that has the same content with new secret1
```

qibobo:
@Avni-Sharma @pedjak 
Acceptance test added and passed.

Issue #728, opened 2020-10-14
Binding using annotation with elementType=sliceOfStrings fails with `unsupported type` error


pmacik:
# Steps to reproduce
## Scenario 1:  `sliceOfStrings` from array
Given a CRD is annotated with service binding annotation using the `sliceOfStrings` element type like in the following snippet:

```yaml
...
metadata:
   annotations:
      service.binding/tags: path={.spec.tags},elementType=sliceOfStrings
```

and the respective CR contains the following spec:

```yaml
...
kind: Backend
...
spec:
   tags:
      - knowledge
      - is
      - power
```
when a service binding is created to bind the service CR to an application I'd expect that the application would be enriched with the following env variables by the binding:

```shell
BACKEND_TAGS_0="knowledge"
BACKEND_TAGS_1="is"
BACKEND_TAGS_2="power"
```

However the following error is being thrown instead:

```
2020-10-14T13:28:35.086+0200	ERROR	controller-runtime.controller	Reconciler error	{"controller": "servicebinding-controller", "request": "service-binding-demo/binding-request", "error": "unsupported type"}
github.com/go-logr/zapr.(*zapLogger).Error
	/home/pmacik/.go/src/github.com/redhat-developer/service-binding-operator/vendor/github.com/go-logr/zapr/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/home/pmacik/.go/src/github.com/redhat-developer/service-binding-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:258
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/home/pmacik/.go/src/github.com/redhat-developer/service-binding-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:232
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
	/home/pmacik/.go/src/github.com/redhat-developer/service-binding-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:211
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
	/home/pmacik/.go/src/github.com/redhat-developer/service-binding-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:152
k8s.io/apimachinery/pkg/util/wait.JitterUntil
[...]

PR #729, opened 2020-10-14
Fix annotations with sliceOfStrings

pmacik:
### Motivation

Implements:
* https://github.com/k8s-service-bindings/spec/issues/131
* [APPSVC-739](https://issues.redhat.com/browse/APPSVC-739)

Fixes:
* https://github.com/redhat-developer/service-binding-operator/issues/728
* [APPSVC-729](https://issues.redhat.com/browse/APPSVC-729)

The `unsupported type` error is thrown because when using a list of strings (a.k.a. slice of strings) the env variable building piece of SBO does not recognize it because it ends up as `[]string` which is not the case of `[]interface{}` or any other considered case.

### Changes

This PR:
* Adds a case for `[]string` in a similar way as `[]interface{}` is in the env variable build part of SBO.
* Improves a bit of error logs related to the issue.
* Implements case for `sliceOfStrings` without `sourceValue`
* Adds acceptance tests

### Testing

`make test-unit`
`TEST_ACCEPTANCE_TAGS=@annotations make test-acceptance`

### DIFF
--- a/pkg/controller/servicebinding/envvars/envvars.go
+++ b/pkg/controller/servicebinding/envvars/envvars.go
@@ -2,6 +2,7 @@ package envvars
 
 import (
 	"errors"
+	"fmt"
 	"sort"
 	"strconv"
 	"strings"
@@ -66,12 +67,14 @@ func Build(obj interface{}, path ...string) (map[string]string, error) {
 		return buildString(strconv.FormatInt(val, 10), path), nil
 	case float64:
 		return buildString(strconv.FormatFloat(val, 'f', -1, 64), path), nil
+	case []string:
+		return buildSliceOfStrings(val, path)
 	case []interface{}:
 		return buildSliceOfInterface(val, path)
 	case bool:
 		return buildString(strconv.FormatBool(val), path), nil
 	default:
-		return nil, errUnsupportedType
+		return nil, fmt.Errorf("%v: %v", errUnsupportedType, val)
 	}
 }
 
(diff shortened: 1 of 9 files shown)


DhritiShikhar:
@pmacik 

Is there even a use-case for `[]interface{}`?
I referred this -> https://github.com/golang/go/wiki/InterfaceSlice

DhritiShikhar:
@pmacik 

I edited postgresql CRD to include annotation `service.binding/tags: path={.spec.tags},elementType=sliceOfStrings` but I corresponding envVar wasnot generated.

Here is the video of what I did to test your PR - https://youtu.be/HZpFbZ2Z7ks

pmacik:
> I edited postgresql CRD to include annotation `service.binding/tags: path={.spec.tags},elementType=sliceOfStrings` but I corresponding envVar wasnot generated.
> 
> Here is the video of what I did to test your PR - https://youtu.be/HZpFbZ2Z7ks

@DhritiShikhar There is a problem currently in `master` with the `sliceOfStrings` (https://github.com/redhat-developer/service-binding-operator/issues/728) and this PR fixes part of it.

DhritiShikhar:
> @DhritiShikhar There is a problem currently in `master` with the `sliceOfStrings` (#728) and this PR suggests a fix for part of it.

Got it.

pmacik:
> @pmacik
> 
> Is there even a use-case for `[]interface{}`?
> I referred this -> https://github.com/golang/go/wiki/InterfaceSlice

To be honest, I'm not sure - @isutton, any thoughts?

pkg/controller/servicebinding/envvars/envvars.go
```diff
@@ -116,6 +119,19 @@ func buildMap(obj map[string]interface{}, path []string) (map[string]string, err
 	return envVars, nil
 }
 
+// buildSliceOfStrings retrurns a slice containing environment variables for
```
DhritiShikhar, line 122:
```suggestion
// buildSliceOfStrings returns a slice containing environment variables for
```

pmacik:
Rebased and squashed commits after update...

pmacik:
> Is there even a use-case for `[]interface{}`?
> I referred this -> https://github.com/golang/go/wiki/InterfaceSlice

Actually `[]string` is [not supported by the k8s apimachinery](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/runtime/converter.go#L468) and fails with panic `cannot deep copy []string`. That is why I needed to update the [sliceOfStringsFromPathDefinition.Apply](https://github.com/redhat-developer/service-binding-operator/pull/729/files#diff-683b572d1128c0146299426b61dba4689c73102cdfb31e0033473c11279ce830R289) method accordingly

Issue #536, opened 2020-07-02
How to install Service Binding Operator on Kubernetes?


dharmit:
I'd like to install SBO on k8s to try linking an application and service on k8s. I'm using minikube

1. I don't find SBO on operatorhub.io website.
2. I tried the step mentioned in [Quick Start section](https://github.com/redhat-developer/service-binding-operator#quick-start) but faced below error:
    ```sh
    $ cat <<EOS |kubectl apply -f -
    ---
    apiVersion: operators.coreos.com/v1
    kind: OperatorSource
    metadata:
      name: redhat-developer-operators
      namespace: openshift-marketplace
    spec:
      type: appregistry
      endpoint: https://quay.io/cnr
      registryNamespace: redhat-developer
    EOS
    error: unable to recognize "STDIN": no matches for kind "OperatorSource" in version "operators.coreos.com/v1"
    ```
Minikube info:
```
$ minikube version
minikube version: v1.11.0
commit: 57e2f55f47effe9ce396cea42a1e0eb4f611ebbd
```

Am I doing something wrong? What's the right way to have SBO on k8s?

sbose78:
> I tried the step mentioned in Quick Start section but faced below error:

Could you try doing so after installing OLM ?

@lanceliuu and @zhangtbj deploys Service Binding on kubernetes clusters in a non-OLM way : could you share your steps?

I would expect it to look like this:
* You could install all the yamls in https://github.com/redhat-developer/service-binding-operator/tree/master/deploy except for the olm-catalog.
* Use quay.io/redhat-developer/app-binding-operator in https://github.com/redhat-developer/service-binding-operator/blob/master/deploy/operator.yaml#L24 

Dharmit, however, if you wish to do an `oc get csv` inside odo, you need it the OLM way :)

greglanthier:
I've been struggling a little to find a way to use OLM to install this operator, too.

I think the confusion with using the OLM way, at least for me, stems from the fact that the `OperatorSource` kind is [deprecated and/or not available](https://github.com/operator-framework/operator-marketplace/pull/284) in new versions of OLM (certainly from [0.15.1](https://github.com/operator-framework/operator-lifecycle-manager/releases/tag/0.15.1) onwards).

I've been looking for ways to add a `CatalogSource` that will let me deploy `service-binding-operator` - something like the equivalent to the following that adds the community-operators catalog to a fresh OLM installation:

```yaml
apiVersion: operators.coreos.com/v1alpha1
[...]

PR #738, opened 2020-10-27
Add installation steps for SBO with OLM on vanilla k8s

Avni-Sharma:
Adding installation steps for running SBO on minikube or any other vanilla k8s.
Fix for https://github.com/redhat-developer/service-binding-operator/issues/536

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,8 +86,17 @@ spec:
 
 ## Quick Start
 
-Clone the repository and run `make local` in an existing `kube:admin` OpenShift
-CLI session. 
+### Running operator locally
+
+Clone the repository and run `make local` in an existing `kube:admin` OpenShift CLI session.
+
+### Installing the operator in a cluster
+
+Start a kubernetes cluster. For this case we have used minikube. Any other Kubernetes cluster can be used as well.
+`minikube start`
+
+Run `chmod +x install.sh` to set execution permission for the install script.
+Now run the install script `install.sh`
 
 
 ## Key Features
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,63 @@
+#!/bin/bash -e
+set -u
+
+#The SBO can also be installed in a vanilla kubernetes cluster. A prerequisite for this would be to add credentials for the registry.redhat.io in the cluster. Steps to be followed to achieve the same are listed below:
+#1) Follow [Red Hat Container Registry Authentication Steps](https://access.redhat.com/RegistryAuthentication)
+#2) Verify that your credentials are correct using docker login -u <your_username> -p <your_passwd> registry.redhat.io
+docker login registry.redhat.io
+
+#3) Start a kubernetes cluster. Any Kubernetes cluster can be used as well. The script assumes that a Kubernetes is up and running and the user has logged into it.
+
+#4) Enable [OLM](https://github.com/operator-framework/operator-lifecycle-manager) in the cluster by running the following command
+curl -sL https://github.com/operator-framework/operator-lifecycle-manager/releases/download/0.16.1/install.sh | bash -s 0.16.1
+#On minikube you can alternatively install OLM by running `minikube addons enable olm`
+
+#5) Create a new image pull secret out of your local .docker/config.json file
+kubectl create secret generic community-operators-secrets -n olm --from-file=.dockerconfigjson=$HOME/.docker/config.json --type=kubernetes.io/dockerconfigjson
+
+#6) Add that pull secret to the default account in olm namespace
+kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "community-operators-secrets"}]}' -n=olm
+
+#7) Install the operator by running the following commands
+#Apply CatalogSource for obtaining catalog of community operators
+kubectl apply -f - << EOD
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: community-operators
+  namespace: olm
+spec:
+  displayName: Community Operators
+  image: registry.redhat.io/redhat/community-operator-index:latest
+  sourceType: grpc
+  publisher: Red Hat
+  updateStrategy:
+    registryPoll:
+      interval: 10m0s
+EOD
+#Apply subscription for subscribing to the beta channel of the service binding operator
+kubectl apply -f - << EOD
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: service-binding-operator
+  namespace: operators
+spec:
+  channel: 'beta'
+  installPlanApproval: Automatic
+  name: service-binding-operator
+  source: community-operators
+  sourceNamespace: olm
+EOD
+#This Operator will be installed in the "operators" namespace and will be usable from all namespaces in the cluster.
+
+retries=50
+until [[ $retries == 0 ]]; do
+  kubectl get deployment/service-binding-operator -n operators >/dev/null 2>&1 && break
+  echo "Waiting for service-binding-operator to be created"
+  sleep 5
+  retries=$(($retries - 1))
+done
+kubectl rollout status -w deployment/service-binding-operator -n operators


sbose78:
Could you improve the docs to have a single line `kubectl apply -f a/url/to/install.yaml` without asking users to clone the repo and run a make command :)

sbose78:
> Could you improve the docs to have a single line kubectl apply -f a/url/to/install.yaml without asking users to clone the repo and run a make command :)

^ could be done separately.

What you have in this PR is extremely helpful!

hack/minikube.sh
```diff
@@ -0,0 +1,61 @@
...
+#set namespace here
+kubectl config set-context --current --namespace=default
+
+kubectl create -f deploy/role.yaml
+kubectl create -f deploy/role_binding.yaml
+kubectl create -f deploy/service_account.yaml
+kubectl create -f deploy/crds/operators.coreos.com_servicebindings_crd.yaml
+
+kubectl create -f - << EOF
+apiVersion: rbac.authorization.k8s.io/v1
```
sbose78, line 16: Might as well store in these yamls in a file so that they can used outside this script too ? :)

hack/minikube.sh
```diff
@@ -0,0 +1,61 @@
...
+- kind: ServiceAccount
+  name: service-binding-operator
+  namespace: default #change the namespace if not using the default namespace
+roleRef:
+  kind: ClusterRole
+  name: sbo-cluster-role
+  apiGroup: rbac.authorization.k8s.io
+EOF
+
+sed -i -e 's,REPLACE_IMAGE,"quay.io/redhat-developer/app-binding-operator:$(GIT_COMMIT_ID)",g' ./deploy/operator.yaml
```
sbose78, line 60: We should ideally remove the placeholder and have  "quay.io/redhat-developer/app-binding-operator:latest" in the `deploy/operator.yaml` . That breaks our yaml directory otherwise :)

Our scripts might need to adapt to it.

sbose78:
@pedjak @Avni-Sharma  Please note, we *must* have a non-OLM way of install the operator documented as well if we were to make this attractive for Kubernetes users. I, as a user, shouldn't have to install OLM only for installing Service Binding Operator, for example. 

For `odo`, it makes sense to use the OLM way of installing, since `odo` already needs OLM  for an 'OLM catalog'.

sbose78:
Also, we must not be using any authenticated registry.redhat.io image urls, there are open options available in quay.io

sbose78:
Found an interesting way ;) 

Added the single install yaml to 
https://github.com/redhat-developer/service-binding-operator/releases/tag/v0.3.0

You can now use 
kubectl apply -f https://github.com/redhat-developer/service-binding-operator/releases/download/v0.3.0/install-fff80ee.yaml

sbose78:
Added the same for https://github.com/redhat-developer/service-binding-operator/releases/tag/v0.1.1-364 for a non-OLM way of installing things.

* Concatenated the yamls in our `deploy/` 
* Replaced image reference
* Added the clusterrole + rolebinding Dharmit discovered 

This should be a release-time activity and could be scripted easily.

Our docs need to have the following ways to install Service Binding:
* Installation on OpenShift/Kubernetes without OLM
* Installation on OpenShift with OLM 
* Installation on Kubernetes with OLM

pedjak:
> Please note, we must have a non-OLM way of installing the operator documented as well if we were to make this attractive for Kubernetes users. I, as a user, shouldn't have to install OLM only for installing Service Binding Operator, for example.

@sbose78 I agree. IMHO, the requirements for this PR and the issue that should be solved (#536) were a bit unspecific (both OLM and non-OLM approached were mentioned). Let's focus the scope of this ticket to OLM approach. In order to move us to upstream OperatorHub we need to verify that first: I would propose the following order:

* Run acceptance tests against plain k8s cluster with installed OLM and SBO
* When green, release to upstream OperatorHub
* Upstream index image requires no credentials, thus a broader community can use them in the script that would install OLM and given SBO version

Afterthat we could repeat the above procedure, but without OLM. IMHO that should happen in a separate PR.

Avni-Sharma:
@pedjak  Yeah we can do that way as well. So we should first test the non OLM way and then give the artifacts yaml as a way to install is what you meant? So this would be done after we have tested with bare artifacts?

pedjak:
@Avni-Sharma we should first test with OLM, because that is our required for us to move to upstream OperatorHub

sbose78:
Agreed Predrag!

docs/installation.md
```diff
@@ -0,0 +1,52 @@
...
+- as Community Operator
+
+
+### Install the Service Binding Operator as Red Hat Operator on OpenShift:
+
+Navigate to the `Operators`->`OperatorHub` in the OpenShift console and in the `Developer Tools` category select the `Service Binding Operator` operator
+
+![Service Binding Operator as shown in OperatorHub](../../assets/operator-hub-sbo-screenshot.png)
+
+and install the `beta` version.
```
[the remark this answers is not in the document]
Avni-Sharma: Done

docs/installation.md
```diff
@@ -0,0 +1,52 @@
...
+
+This makes the `ServiceBinding` custom resource available, that the application developer will use later.
+
+### Install the Service Binding Operator as a Community Operator on vanilla kubernetes:
```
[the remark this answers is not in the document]
Avni-Sharma: Done

docs/installation.md
```diff
@@ -0,0 +1,52 @@
+# Installation Guide
+
+Currently the SBO can be installed in OpenShift 4.5+ clusters via OLM.
+- as Red Hat Operator
+- as Community Operator
+
+
+### Install the Service Binding Operator as Red Hat Operator on OpenShift:
```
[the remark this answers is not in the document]
Avni-Sharma: Done

docs/installation.md
```diff
@@ -0,0 +1,52 @@
...
+
+and install the `beta` version.
+
+This makes the `ServiceBinding` custom resource available, that the application developer will use later.
+
+### Install the Service Binding Operator as a Community Operator on vanilla kubernetes:
+
+The SBO can also be installed in a vanilla kubernetes cluster. A prerequisite for this would be to add credentials for the registry.redhat.io in the cluster. Steps to be followed to achieve the same are listed below:
+
+- Get a Red Hat Login - To use the registry.redhat.io registry, you have to have a Red Hat login
```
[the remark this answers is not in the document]
Avni-Sharma: Done

README.md
```diff
@@ -87,8 +87,15 @@ spec:
 ## Quick Start
 
 Clone the repository and run `make local` in an existing `kube:admin` OpenShift
-CLI session. 
+CLI session
 
+### Running operator locally
+
+Clone the repository and run `make local` in an existing `kube:admin` OpenShift CLI session.
```
pedjak, line 94: This sentence is the same as the one above under "Quick Start" - should we remove the one from above?

install-sbo.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash -e
+set -e
+set -u
+REGISTRY_PASSWORD?=""
+
+#The SBO can also be installed in a vanilla kubernetes cluster. A prerequisite for this would be to add credentials for the registry.redhat.io in the cluster. Steps to be followed to achieve the same are listed below:
+#1) Follow [Red Hat Container Registry Authentication Steps](https://access.redhat.com/RegistryAuthentication)
+#2) Verify that your credentials are correct using docker login -u <your_username> -p <your_passwd> registry.redhat.io
+echo "${REGISTRY_PASSWORD}" | docker login -u "$USERNAME" --password-stdin registry.redhat.io
```
pedjak, line 9: placing a password in env is a security issue... instead, we can use just `docker login registry.redhat.io`. If user did not login before, he/she would be prompted for entering username/password - otherwise, docker would try to use the existing credentials stored in `~/.docker/config.json` and exit with code 0 if everything is ok. We should check the exit code and fail if not zero.

install-sbo.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash -e
```
pedjak, line 1: let's name the script simply `install.sh`

README.md
```diff
@@ -87,8 +87,15 @@ spec:
...
+
+### Installing the operator in a cluster
+
+Follow steps from the [Installation Guide](docs/installation.md) to install the Service Binding Operator in OpenShift 4.5+ or any other Kubernetes cluster.
```
pedjak, line 98: We need a section somewhere about usage of `install.sh` script.

install-sbo.sh
```diff
@@ -0,0 +1,57 @@
...
+set -u
+REGISTRY_PASSWORD?=""
+
+#The SBO can also be installed in a vanilla kubernetes cluster. A prerequisite for this would be to add credentials for the registry.redhat.io in the cluster. Steps to be followed to achieve the same are listed below:
+#1) Follow [Red Hat Container Registry Authentication Steps](https://access.redhat.com/RegistryAuthentication)
+#2) Verify that your credentials are correct using docker login -u <your_username> -p <your_passwd> registry.redhat.io
+echo "${REGISTRY_PASSWORD}" | docker login -u "$USERNAME" --password-stdin registry.redhat.io
+
+#3) Start a kubernetes cluster. For this case we have used minikube. Any other Kubernetes cluster can be used as well.
+minikube start
```
pedjak, line 12: The script should assume that a cluster is up and running - we should not spin up cluster here. This would enable us to use the same script against various clusters.

install-sbo.sh
```diff
@@ -0,0 +1,57 @@
...
+
+#4) Enable [OLM](https://github.com/operator-framework/operator-lifecycle-manager) in the cluster by running the following command
+curl -sL https://github.com/operator-framework/operator-lifecycle-manager/releases/download/0.16.1/install.sh | bash -s 0.16.1
```
pedjak, line 15: we should check if OLM is enabled and if not install it. Extractg OLM version into some variable to make the maintenance easier.

pedjak (changes requested):

install-sbo.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash -e
+set -e
```
pedjak, line 2: `-e` option is already incorporated in the line 1, hence we can remove it here.

install-sbo.sh
Avni-Sharma: If OLM is already installed it would just show as a resource `unchanged` in case of an `apply` API request

README.md
Avni-Sharma: Done

README.md
```diff
@@ -86,9 +86,16 @@ spec:
...
+### Running operator locally
 
+Clone the repository and run `make local` in an existing `kube:admin` OpenShift CLI session.
+
+### Installing the operator in a cluster
+
+Start a kubernetes cluster. For this case we have used minikube. Any other Kubernetes cluster can be used as well.
+`minikube start`
+
+Now run the install script `sh install.sh`
```
pedjak, line 98: `install.sh` should have execution permission set, so that `install.sh` is just enough.

install.sh
```diff
@@ -0,0 +1,61 @@
...
+#1) Follow [Red Hat Container Registry Authentication Steps](https://access.redhat.com/RegistryAuthentication)
+#2) Verify that your credentials are correct using docker login -u <your_username> -p <your_passwd> registry.redhat.io
+docker login registry.redhat.io
+
+EXIT_CODE=$?
+if [ $EXIT_CODE -ne 0 ];
+then
+  echo "docker login was not successful. Please try again"
+  exit 0
+fi
```
pedjak, line 14: in case of unsuccessful login, we are not going to reach this block, because shebang line is `#!/bin/bash -e` - `-e` arg makes that scripts exits on any error immediately.

pedjak (changes requested):

install.sh
Avni-Sharma: In that case we can omit checking the exit code? https://github.com/redhat-developer/service-binding-operator/pull/738#discussion_r517880610

pedjak: right, lines 9-14 will never be executed.

Avni-Sharma: Done

README.md
Avni-Sharma: done

install-sbo.sh
Avni-Sharma: Done

README.md
Avni-Sharma: Done

install-sbo.sh
Avni-Sharma: Done

install.sh
```diff
@@ -0,0 +1,54 @@
...
+  name: service-binding-operator
+  namespace: operators
+spec:
+  channel: 'beta'
+  installPlanApproval: Automatic
+  name: service-binding-operator
+  source: community-operators
+  sourceNamespace: olm
+EOD
+#This Operator will be installed in the "operators" namespace and will be usable from all namespaces in the cluster.
```
pedjak, line 54: just a small UX improvement: it would be great to wait here until service operator is up and ready. We can achieve that with:

```sh
kubectl rollout status -w deployment/service-binding-operator -n operators
```

pedjak (approved):
just one more additional UX improvement, see the comment. You can add it and then squash immediately.

pedjak (approved):

install.sh
```diff
@@ -0,0 +1,55 @@
...
+kubectl rollout status -w deployment/service-binding-operator -n operators
```
pmacik, line 55: It takes a while before the actual deployment of SBO is created by the OLM operator (it took ~30s on my laptop with minikube) and because of that this `rollout status` step (and the entire install script along with it) fails without waiting for that deployment to be created:
```sh
$ kubectl rollout status -w deployment/service-binding-operator -n operators
Error from server (NotFound): deployments.apps "service-binding-operator" not found
```
The script should wait for the `service-binding-operator` deployment to be created for some time before checking the rollout status.

pmacik (changes requested):

install.sh
Avni-Sharma: I checked the rollout status independently on the command line and it waited till the deployment was up and running. :thinking:  Will check it again with the command in the script

pedjak: good catch @pmacik - lets introduce some loop where we check the existence of deployment with something like `kubectl get deployment service-binding-operator -n operators` - looking into exit code should be enough

install.sh
```diff
@@ -0,0 +1,58 @@
...
+  installPlanApproval: Automatic
+  name: service-binding-operator
+  source: community-operators
+  sourceNamespace: olm
+EOD
+#This Operator will be installed in the "operators" namespace and will be usable from all namespaces in the cluster.
+
+while [[ $(kubectl get deployment/service-binding-operator -n operators -o 'jsonpath={..status.conditions[?(@.type=="Available")].status}') != "True" ]]; do
+    echo "waiting for pod" && sleep 5;
+  done
```
pedjak, line 58: we can wait here forever if something goes wrong. How about this:

```bash
retries=100
until [[ $retires == 0 ]]; do
  kubectl get deployment/service-binding-operator -n operators >/dev/null 2>&1 && break
  echo "Waiting for service-binding-operator to be created"
  sleep 5
  retries=$((retries - 1))
done
kubectl rollout status -w deployment/service-binding-operator -n operators
```

install.sh
Avni-Sharma: Done

pedjak:
@pmacik can you verify if the script works for you now correctly?

pmacik:
> @pmacik can you verify if the script works for you now correctly?

@pedjak Yes, I've just verified that the script works correctly, now.

PR #747, opened 2020-11-02
make target for Operator SDK

gorkem:
### Motivation

Make it easier to manage the Operator SDK version

### Changes

adds a new make target for operator SDK installation

### Testing
```
make install-operator-sdk
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -487,3 +487,16 @@ dev-release:
 validate-release: setup-venv
 	$(Q)$(PYTHON_VENV_DIR)/bin/pip install yq==2.10.0
 	BUNDLE_VERSION=$(BASE_BUNDLE_VERSION) CHANNEL="beta" ./hack/validate-release.sh
+
+# Operator-sdk release version
+RELEASE_VERSION ?= v0.16.0
+
+.PHONY: install-operator-sdk
+install-operator-sdk:
+ifeq ($(shell uname -s), Darwin)
+	curl -LO https://github.com/operator-framework/operator-sdk/releases/download/${RELEASE_VERSION}/operator-sdk-${RELEASE_VERSION}-x86_64-apple-darwin
+	chmod +x operator-sdk-${RELEASE_VERSION}-x86_64-apple-darwin && sudo mkdir -p /usr/local/bin/ && sudo cp operator-sdk-${RELEASE_VERSION}-x86_64-apple-darwin /usr/local/bin/operator-sdk && rm operator-sdk-${RELEASE_VERSION}-x86_64-apple-darwin
+else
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -487,3 +487,16 @@ dev-release:
 validate-release: setup-venv
 	$(Q)$(PYTHON_VENV_DIR)/bin/pip install yq==2.10.0
 	BUNDLE_VERSION=$(BASE_BUNDLE_VERSION) CHANNEL="beta" ./hack/validate-release.sh
+
+# Operator-sdk release version
+RELEASE_VERSION ?= v0.16.0
```
pedjak, line 492: please move this var at the top of Makefile, where other vars are - it would make the maintenance easier. Also, please name it like `OPERATOR_SDK_VERSION`

Makefile
```diff
@@ -487,3 +487,16 @@ dev-release:
...
+
+.PHONY: install-operator-sdk
+install-operator-sdk:
+ifeq ($(shell uname -s), Darwin)
+	curl -LO https://github.com/operator-framework/operator-sdk/releases/download/${RELEASE_VERSION}/operator-sdk-${RELEASE_VERSION}-x86_64-apple-darwin
+	chmod +x operator-sdk-${RELEASE_VERSION}-x86_64-apple-darwin && sudo mkdir -p /usr/local/bin/ && sudo cp operator-sdk-${RELEASE_VERSION}-x86_64-apple-darwin /usr/local/bin/operator-sdk && rm operator-sdk-${RELEASE_VERSION}-x86_64-apple-darwin
```
pedjak, line 498: current logic install it under `$GOPATH/bin`, let's keep it like this, instead of installing it globally for the system.

pedjak (changes requested):

Makefile
gorkem: `GOPATH/bin` is not necessarily on my PATH when I am working with the project. Ideally, I would not have to install a specific version and the copy of the SDK I receive from brew would work when I am working locally. But the project has neglected to keep up with the SDK releases and here we are.

pedjak: Ok, in that case, how about that we install operator-sdk in project `out` dir and use that copy throught the project? Or to have Dockerfile crafted with the proper operator-sdk version and spin container whenever we need operator-sdk?

isutton: The assumption was that `operator-sdk` was provided by the environment, not managed by project's build system, which fits @gorkem's requirement on using an `operator-sdk` provided by `brew`.

Unfortunately the presented approach overwrites the host's toolchain, which I consider harmful.

A different approach would be to assert whether the system's `operator-sdk`'s version matches the required version for building the operator in the Makefile and offer the `hack/install-operator-sdk.sh` script to make it easier to install it in the developer's host, leaving the responsibility to choose a specific version to the host administrator, like it currently is.

PR #748, opened 2020-11-03
Add acceptance test to verify `route_k8s_resource` example

pratikjagrut:
### Motivation

JIRA TICKET: https://issues.redhat.com/browse/APPSVC-708

### DIFF
--- a/test/acceptance/features/steps/steps.py
+++ b/test/acceptance/features/steps/steps.py
@@ -25,6 +25,7 @@ from quarkus_s2i_builder_image import QuarkusS2IBuilderImage
 from serverless_operator import ServerlessOperator
 from service_binding import ServiceBinding
 from servicebindingoperator import Servicebindingoperator
+from app import App
 
 
 # STEP
@@ -370,6 +371,7 @@ def check_secret_key_with_ip_value(context, secret_name, secret_key):
 
 
 # STEP
+@given(u'The openshift route is present')
 @given(u'Namespace is present')
 @given(u'Backend service CSV is installed')
 @given(u'The Custom Resource Definition is present')
@@ -473,3 +475,10 @@ def validate_secret_empty(context, secret_name):
                       ignore_exceptions=(json.JSONDecodeError,))
     except polling2.TimeoutException:
         pass
+
+
+@given(u'"{app_name}" is deployed from image "{image_ref}"')
+def create_deployment(context, app_name, image_ref):
+    app = App(app_name, context.namespace.name, image_ref)
+    if not app.is_running():
+        assert app.install() is True, "Failed to create deployment."
--- a/test/acceptance/features/examples.feature
+++ b/test/acceptance/features/examples.feature
@@ -1,0 +1,55 @@
+@examples
+Feature: Verify examples provided in Service Binding Operator github repository
+
+    As a user of Service Binding Operator
+    I want to verify if all the examples given are vouching for the mentioned features
+
+    Background:
+        Given Namespace [TEST_NAMESPACE] is used
+        * Service Binding Operator is running
+
+    # https://github.com/redhat-developer/service-binding-operator/blob/master/examples/route_k8s_resource/README.md
+    Scenario: Bind an application to an openshift route
+        Given The openshift route is present
+            """
+            kind: Route
+            apiVersion: route.openshift.io/v1
+            metadata:
+                name: my-route
+                annotations:
+                    openshift.io/host.generated: 'true'
+                    service.binding/host: path={.spec.host} #annotate here.
+            spec:
+                host: example-sbo.apps.ci-ln-smyggvb-d5d6b.origin-ci-int-aws.dev.rhcloud.com
+                path: /
+                to:
+                    kind: Service
+                    name: my-service
+                    weight: 100
+                port:
+                    targetPort: 80
+            wildcardPolicy: None
+            """
+        * "hello-app" is deployed from image "openshift/hello-openshift"
+        When Service Binding is applied
+            """
+            apiVersion: operators.coreos.com/v1alpha1
+            kind: ServiceBinding
+            metadata:
+                name: sbr-to-bind-hello-app-to-route
+            spec:
+                application:
+                    group: apps
+                    resource: deployments
+                    name: hello-app
+                    version: v1
+                services:
+                  - group: route.openshift.io
+                    version: v1
+                    kind: Route
+                    name: my-route
+            """
+        Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "sbr-to-bind-hello-app-to-route" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "sbr-to-bind-hello-app-to-route" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="Ready").status" of Service Binding "sbr-to-bind-hello-app-to-route" should be changed to "True"
+        And Secret "sbr-to-bind-hello-app-to-route" contains "ROUTE_HOST" key with value "example-sbo.apps.ci-ln-smyggvb-d5d6b.origin-ci-int-aws.dev.rhcloud.com"


test/acceptance/features/bindAppToService.feature
```diff
@@ -723,3 +723,47 @@ Feature: Bind an application to a service
         Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-request-cross-ns-service" should be changed to "True"
         And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-cross-ns-service" should be changed to "True"
         And The application env var "BACKEND_HOST_CROSS_NS_SERVICE" has value "cross.ns.service.stable.example.com"
+
+    Scenario: Bind an application to a openshift route
```
pmacik, line 727: Please place the scenario into a separate feature (`verifyExamples.feature`) and tag it with `@example` tag. Thx

test/acceptance/features/bindAppToService.feature
```diff
@@ -723,3 +723,47 @@ Feature: Bind an application to a service
...
+            """
+            kind: Route
+            apiVersion: route.openshift.io/v1
+            metadata:
+                name: my-route
+                annotations:
+                    openshift.io/host.generated: 'true'
+                    service.binding/host: path={.spec.host} #annotate here.
+            spec:
+                host: example-sbo.apps.ci-ln-smyggvb-d5d6b.origin-ci-int-aws.dev.rhcloud.com
```
pmacik, line 738: Would this kind of host work every cluster? On OpenShift CI? On CRC? On any other ordinary OpenShift cluster?

pratikjagrut: I guess yes. When we expose the service in openshift it has a similar format.

test/acceptance/features/bindAppToService.feature
pratikjagrut: OKAY.

pmacik: You are correct in general but then you need to define what a "feature" actually is. IMHO "bind app to service" is way too broad, look:
```feature
Feature: Bind an application to a service

    As a user of Service Binding Operator
    I want to bind applications to services it depends on
...
```

and it's basically everything what the operator does. Originally, we started with a single feature file for simplicity, but now when we have almost 50 scenarios (and number is growing with almost every PR) we need to start splitting the files logically if possible, because:
1. we need to keep the size of feature files reasonably large to be able to manage them.
2. it's good to have related scenarios grouped - like we have in case of annotations.
I agree that In this particular case "demonstrating and tutoring how to use the operator" (what is the point of the examples) is not a "feature of the operator" per say, but rather we use the concept of feature and use it to group scenarios that are related not by a feature of the operator itself but by purpose (to verify examples).

pratikjagrut: Okay, I'll make the changes.

pratikjagrut: @pmacik I've updated it PTAL.

test/acceptance/features/verifyExamples.feature
```diff
@@ -0,0 +1,54 @@
...
+    As a user of Service Binding Operator
+    I want to verify if all the examples given are vouching for the mentioned features
+
+    Background:
+        Given Namespace [TEST_NAMESPACE] is used
+        * Service Binding Operator is running
+
+    @example
+    # https://github.com/redhat-developer/service-binding-operator/blob/master/examples/route_k8s_resource/README.md
+    Scenario: Bind an application to a openshift route
```
pmacik, line 12: @pratikjagrut Thanks for adding this, however I have some notes.

While this scenario shares the important piece, that is binding the route to an application, It actually does not verify the the steps of the `route_k8s_resource` example. The point of the story is to verify the exact steps (or as close as possible) in the example.

For instance, the example uses the Node.js application to bind a route with, however this scenario uses the generic test application. The scenario needs to follow the steps stated in the example (as close as possible) to really verifying it.

pmacik (changes requested):
By checking the actual scenario in the example I noticed it is messy and has steps that are useless (such as installing the DB operator) and it also uses the nodejs application that needs to be imported via a wizard, which is an overhead for this particular example (that should focus on binding a Route).

So I fixed the example via https://github.com/redhat-developer/service-binding-operator/pull/753. Could you please update the acceptance test scenario to contain the actual steps from that fixed example?

test/acceptance/features/verifyExamples.feature
pratikjagrut: Okay I get it.

test/acceptance/features/verifyExamples.feature
```diff
@@ -0,0 +1,73 @@
+Feature: Verify examples provided in Service Binding Operator github repository
```
pedjak, line 1: I would vote for naming this file simply `examples.feature`

test/acceptance/features/verifyExamples.feature
```diff
@@ -0,0 +1,73 @@
...
+    As a user of Service Binding Operator
+    I want to verify if all the examples given are vouching for the mentioned features
+
+    Background:
+        Given Namespace [TEST_NAMESPACE] is used
+        * Service Binding Operator is running
+
+    @example
+    # https://github.com/redhat-developer/service-binding-operator/blob/master/examples/route_k8s_resource/README.md
+    Scenario: Bind an application to a openshift route
```
pedjak, line 12: nit: an openshift route

pedjak (changes requested):

test/acceptance/features/verifyExamples.feature
pratikjagrut: Done,

test/acceptance/features/verifyExamples.feature
pratikjagrut: Done.

test/acceptance/features/steps/steps.py
```diff
@@ -466,3 +468,10 @@ def validate_secret_empty(context, secret_name):
                       ignore_exceptions=(json.JSONDecodeError,))
     except polling2.TimeoutException:
         pass
+
+
+@given(u'"{image}" image is deployed as "{deployment_name}"')
```
pedjak, line 473: IMHO, it reads better if we declare the step as:

"{app_name} is deployed from image {image_ref}"

pedjak (changes requested):
look good, a small request to rephrase one step and then squash need to happen before merging it.

test/acceptance/features/steps/steps.py
pratikjagrut: okay.

test/acceptance/features/steps/steps.py
```diff
@@ -466,3 +468,10 @@ def validate_secret_empty(context, secret_name):
                       ignore_exceptions=(json.JSONDecodeError,))
     except polling2.TimeoutException:
         pass
+
+
+@given(u'"{app_name}" is deployed from image "{image_ref}"')
+def create_deployment(context, app_name, image_ref):
+    cmd = Command()
+    output, exit_code = cmd.run(f'kubectl create deployment --image={image_ref} {app_name}')
+    assert exit_code == 0, f"Could not create deployment - {output}"
```
pedjak, line 477: We already have `App` class that models any application:

https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/steps/app.py#L6

Let's reuse it.

pedjak (changes requested):

test/acceptance/features/steps/steps.py
pratikjagrut: In this function its also exposing application but in our example which we're verifying here, we're not. Does splitting these two tasks(deploying the app and exposing it) make sense?

pmacik: Exposure of a deployment only creates a Service resource that can be used to access the application. Which is important in case of the generic test `App` to be able to access the API by tests.
In case of the example, only the `Deployment` while no API is accessed, at all. However, the exposure is just additional layer that does not interfere with the example in any matter (since the example only access the intermediate secret). So use of the generic application actually implements the example's step.

pratikjagrut: Okay, I'll update implementation.

pratikjagrut: @pedjak PTAL.

test/acceptance/features/steps/steps.py
```diff
@@ -466,3 +468,12 @@ def validate_secret_empty(context, secret_name):
...
         pass
+
+
+@given(u'"{app_name}" is deployed from image "{image_ref}"')
+def create_deployment(context, app_name, image_ref):
+    app = App(app_name, context.namespace.name, image_ref)
+    if not app.is_running:
+        assert False, "Deployment is already present."
+    else:
+        assert app.install() is True, "Failed to create deployment."
```
pedjak, line 479: We should not fail if app is not running, instead we should try to install it and then check if running, similar to https://github.com/redhat-developer/service-binding-operator/blob/f1c3975476de43b2e305ee03c29632065291392c/test/acceptance/features/steps/generic_testapp.py#L24

pedjak (changes requested):

test/acceptance/features/steps/steps.py
pratikjagrut: Updated.

test/acceptance/features/steps/steps.py
```diff
@@ -466,3 +468,10 @@ def validate_secret_empty(context, secret_name):
                       ignore_exceptions=(json.JSONDecodeError,))
     except polling2.TimeoutException:
         pass
+
+
+@given(u'"{app_name}" is deployed from image "{image_ref}"')
+def create_deployment(context, app_name, image_ref):
+    app = App(app_name, context.namespace.name, image_ref)
+    if not app.is_running():
+        assert app.install() is True, "Failed to create deployment."
```
pedjak, line 477: after installing we should check once again if it is running, but this time we should wait until this happens or the timeout occurs.

pratikjagrut: This case is handled by install function implicitly so we won't need to check it again.

pedjak (approved):

pedjak:
//lgtm
/approve

PR #749, opened 2020-11-03
[acceptance-tests] Make acceptance tests compatible with plain k8s clusters

pmacik:
* `TEST_ACCEPTANCE_CLI` env var controls what CLI is used for comunication with cluster. Default: `oc`
* Expose application endpoints via services and node ports instead of Openshift routes
* Tested against minikube with OLM installed
* Tests requiring 3rd party operators are annotated with `@olm`
* Added additional job to Travis CI conf, enabling acceptanced test running against minikube on each PR

### Testing
1. Set `KUBECONFIG` for both minikube and acceptance tests (it will be generated at minikube's start if it does not exist):
```sh
$ export KUBECONFIG=/tmp/minikubeconfig
```
2. Start minikube (v1.14.2):
```sh
$ minikube start
minikube v1.14.2 on Generic 31
    ▪ KUBECONFIG=/tmp/minikubeconfig
...
🏄  Done! kubectl is now configured to use "minikube" by default
```
3. Enable olm on minikube:
```sh
$ minikube addons enable olm
🌟  The 'olm' addon is enabled
```
4. Execute all acceptance tests expect those tagged with `@knative` using `kubectl` CLI:
```sh
$ TEST_ACCEPTANCE_CLI=kubectl TEST_ACCEPTANCE_TAGS="~@knative,~@examples" make test-acceptance
```


test/acceptance/features/steps/steps.py
```diff
@@ -420,13 +419,17 @@ def invalid_sbr_is_applied(context):
...
 
 
+@then(u'Error message is thrown')
 @then(u'Error message "{err_msg}" is thrown')
-def validate_error(context, err_msg):
-    search = re.search(rf'.*{err_msg}.*', context.expected_error)
-    assert search is not None, f"Actual error: '{context.expected_error}', Expected error: '{err_msg}'"
+def validate_error(context, err_msg=None):
+    if err_msg is None:
+        assert context.expected_error is not None, f"No error message should happen, but was {err_msg}"
```
pmacik, line 429: The assertion message is reverted
```suggestion
        assert context.expected_error is not None, f"Any error message should happen, but was {context.expected_error}"
```

pedjak: good catch.

this is the case where we did not specify the error message, but some error should happen - if assertion fails, then the `expected_error` is `None`, IMHO not need to put that in the assertion error message. How about this:

```
assert context.expected_error is not None, f"An error message should happen"
```

pmacik: I agree, we don't need to mention the `None` in the message :)

pedjak:
/hold cancel

Running acceptance tests again minikube on each PR has been added in this PR

Avni-Sharma:
I tried to run this by following the commands given in description. I think I am missing something. I am facing few errors like
```
      | COMMAND : kubectl wait --for=condition=Available=True deployment/ssd-2 -n test-namespace-4 --timeout=0s
      '---------'-
      ERROR MESSGE: b'Error from server (NotFound): deployments.apps "ssd-2" not found\n'

```

      | COMMAND : kubectl apply -n test-namespace-4 -f /home/avni/go/src/github.com/redhat-developer/service-binding-operator/test/acceptance/resources/backend.operator.manifest.yaml
      '---------'-
      ERROR MESSGE: b'customresourcedefinition.apiextensions.k8s.io/backends.stable.example.com unchanged\nerror: unable to recognize "/home/avni/go/src/github.com/redhat-developer/service-binding-operator/test/acceptance/resources/backend.operator.manifest.yaml": no matches for kind "ClusterServiceVersion" in version "operators.coreos.com/v1alpha1"\n'
      ERROR CODE: 1
```

Also how are we ignoring these for each PR that would be raised
```
      | COMMAND : kubectl apply  -f -
      '---------'-
      ERROR MESSGE: b'error: unable to recognize "STDIN": no matches for kind "CatalogSource" in version "operators.coreos.com/v1alpha1"\n'
      ERROR CODE: 1

```

```
      | COMMAND : kubectl apply -n test-namespace-4 -f -
      '---------'-
      ERROR MESSGE: b'error: unable to recognize "STDIN": no matches for kind "Route" in version "route.openshift.io/v1"\n'
      ERROR CODE: 1
```

File: .travis.yml [hunk 2]
```diff
@@ -55,4 +55,30 @@ jobs:
...
         - cd ..
         - make push-bundle-to-quay
+    - stage: Acceptance tests
+      if: type = pull_request
+      env:
+        - KUBECONFIG=$HOME/miniconfig
+        - MINIKUBE_WANTUPDATENOTIFICATION=false
+        - MINIKUBE_WANTREPORTERRORPROMPT=false
+        # exclude knative and example annotated tests
+        - EXTRA_BEHAVE_ARGS="--tags=~@knative --tags=~@examples"
```
pmacik, line 65: There is no need for extra variable (`EXTRA_BEHAVE_ARGS` for setting tag expression)... you can use existing like `TEST_ACCEPTANCE_TAGS="~@knative,~@examples"`

pedjak: Unfortunately, we need that - because `TEST_ACCEPTANCE_TAGS="~@knative,~@examples"` semantic implies logical OR, but individual `--tags` options are connected with logical AND. See the official help:

```
$ behave --tags-help
Scenarios inherit tags declared on the Feature level. The simplest
TAG_EXPRESSION is simply a tag::

    --tags @dev

You may even leave off the "@" - behave doesn't mind.

When a tag in a tag expression starts with a ~, this represents boolean NOT::

    --tags ~@dev

A tag expression can have several tags separated by a comma, which represents
logical OR::

    --tags @dev,@wip

The --tags option can be specified several times, and this represents logical
AND, for instance this represents the boolean expression
"(@foo or not @bar) and @zap"::

    --tags @foo,~@bar --tags @zap.

Beware that if you want to use several negative tags to exclude several tags
you have to use logical AND::

    --tags ~@fixme --tags ~@buggy.

```

Scenarios that we run on minikube should not have the following tags: `@disabled`, `@knative`, and `@examples` and only way to achieve that on Behave 1.2.6 is to use logical AND, i.e. 

`--tags=~@disabled --tags=~@knative --tags=~@examples`

In Behave 1.2.7 the tag syntax is improved, but that version is not released yet.

pmacik: Ah, I assumed that we would use the "tag expressions" described [here](https://behave.readthedocs.io/en/latest/tag_expressions.html).
In our case that would be: `TEST_ACCEPTANCE_TAGS="not (@knative and @examples)"`.

But didn't realize it wasn't released, yet.

However, having `EXTRA_BEHAVE_ARGS` available is useful for other purposes, anyway. So I agree to use your approach.

PR #758, opened 2020-11-12
[acceptance-tests] Use out/operator binary for local SBO instance instead of operator-sdk

pmacik:
### Motivation

Currently the `deploy-sbo-local.sh` and `make local` use `operator-sdk run --local` to start the local instance of SBO (SDK builds and run the `build/_output/bin/service-binding-operator-local`, however the "production" binary that is part of the shipped image is build by the `make build` target and is actually `out/operator`:

This means that the acceptance tests actually test a different binary, than the one that is a part of releases.

### Changes

This PR:
* Modifies the `local` Makefile target to build and use `out/operator` binary
* Modifies the `test-acceptance-setup` Makefile target to build and use `out/operator` binary
* Updates the `hack/deploy-sbo-local.sh` script to be able to start local instance of SBO either in background or foreground.
* Unifies the above targets to all use `hack/deploy-sbo-local.sh` script approprietaly

### Testing

`make local`
`make test-acceptance-smoke`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,6 +107,7 @@ SBR_MANIFESTS ?= ${PROJECT_DIR}service-binding-operator-manifests
 CGO_ENABLED ?= 0
 GO111MODULE ?= on
 GOCACHE ?= "$(shell echo ${PWD})/out/gocache"
+GOFLAGS ?= -mod=vendor
 OS ?= $(shell go env GOOS)
 ARCH ?= $(shell go env GOARCH)
 
@@ -188,7 +189,7 @@ lint-yaml: ${YAML_FILES}
 lint-go-code: $(GOLANGCI_LINT_BIN)
 	# This is required for OpenShift CI enviroment
 	# Ref: https://github.com/openshift/release/pull/3438#issuecomment-482053250
-	$(Q)GOCACHE=$(GOCACHE) $(OUTPUT_DIR)/golangci-lint ${V_FLAG} run --deadline=30m
+	$(Q)GOFLAGS="$(GOFLAGS)" GOCACHE="$(GOCACHE)" $(OUTPUT_DIR)/golangci-lint ${V_FLAG} run --deadline=30m
 
 $(GOLANGCI_LINT_BIN):
 	$(Q)curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ./out v1.18.0
@@ -240,8 +241,8 @@ test-cleanup: get-test-namespace
 ## Runs the unit tests without code coverage
 test-unit:
 	$(info Running unit test: $@)
-	$(Q)GO111MODULE=$(GO111MODULE) GOCACHE=$(GOCACHE) \
-		go test $(shell GOCACHE="$(GOCACHE)" go list ./...) -v -mod vendor $(TEST_EXTRA_ARGS)
+	$(Q)GO111MODULE=$(GO111MODULE) GOCACHE="$(GOCACHE)" \
+		go test $(shell GOFLAGS="$(GOFLAGS)" GOCACHE="$(GOCACHE)" go list ./...) -v $(GOFLAGS) $(TEST_EXTRA_ARGS)
 
 .PHONY: test-unit-with-coverage
 ## Runs the unit tests with code coverage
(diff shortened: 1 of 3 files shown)


hack/deploy-sbo-local.sh
```diff
@@ -8,6 +8,8 @@ mkdir -p "$OUTPUT"
 
 SBO_LOCAL_LOG="$OUTPUT/sbo-local.log"
 
+export GOCACHE=${GOCACHE:-out/gocache}
```
pedjak, line 11: our project does not use gocache at all, all dependencies are vendored, and hence we should build binary first using vendored dependencies, and then start it. We should avoid using `operator-sdk` on line 25 because under the good invokes `go build..`

pedjak (changes requested):
ins

hack/deploy-sbo-local.sh
pmacik: Oh, I didn't realize that.. I'll update the script to use the proper binary, Thx, @pedjak

pedjak (changes requested):
scope of this PR becomes different, the commit message and PR description need to be updated.

* we need to modify Makefile `local` rule to use build binary. Also that rule should depend on `build` rule
* same goes for `test-acceptance-setup`
* we could stop operator with some `stop-local` rule

Makefile
```diff
@@ -259,7 +259,7 @@ test-unit-with-coverage:
 ## Setup the environment for the acceptance tests
 test-acceptance-setup: setup-venv
 ifeq ($(TEST_ACCEPTANCE_START_SBO), local)
-test-acceptance-setup: test-cleanup create-test-namespace deploy-test-3rd-party-crds set-test-namespace deploy-rbac deploy-crds
+test-acceptance-setup: test-cleanup create-test-namespace deploy-test-3rd-party-crds set-test-namespace deploy-rbac deploy-crds build
```
pedjak, line 262: let's make `build` first in the list so that we can fail fast.

pmacik: good point

Makefile
```diff
@@ -311,7 +311,7 @@ test: test-unit test-acceptance
 build: out/operator
 
 out/operator:
-	$(Q)GOARCH=$(ARCH) GOOS=$(OS) go build ${V_FLAG} -o $(OUTPUT_DIR)/operator cmd/manager/main.go
+	$(Q)GOCACHE=$(GOCACHE) GOARCH=$(ARCH) GOOS=$(OS) go build ${V_FLAG} -o $(OUTPUT_DIR)/operator cmd/manager/main.go
```
pmacik, line 314: Without this, We get the following in CI:
```
failed to initialize build cache at /.cache/go-build: mkdir /.cache: permission denied
failed to initialize build cache at /.cache/go-build: mkdir /.cache: permission denied
GOARCH= GOOS= go build -v -o /go/src/github.com/redhat-developer/service-binding-operator/out/operator cmd/manager/main.go
failed to initialize build cache at /.cache/go-build: mkdir /.cache: permission denied
```

pedjak: hm, but here we also need to use vendoring, but we do not: how about that we set `GOFLAGS="-mod vendor"` instead of COCACHE? Can you please try that?

hack/deploy-sbo-local.sh
```diff
@@ -8,32 +8,30 @@ mkdir -p "$OUTPUT"
...
+    while [ -z "$(grep 'Starting workers' $SBO_LOCAL_LOG)" ]; do
+        if [[ $attempts -ge 0 ]]; then
+            sleep 5
+            attempts=$((attempts-1))
+        else
+            echo "FAILED"
+            kill $SBO_PID
+            exit 1
+        fi
+    done
```
pedjak, line 32: do we really need this, i.e. what we are ensuring with that?

Makefile
```diff
@@ -371,8 +371,13 @@ push-image: build-image
...
 ## Local: Run operator locally
-local: deploy-clean deploy-rbac deploy-crds
-	$(Q)operator-sdk --verbose run --local --operator-flags "$(ZAP_FLAGS)"
+local: stop-local build get-test-namespace deploy-clean deploy-rbac deploy-crds
+	$(Q)OPERATOR_NAMESPACE="$(TEST_NAMESPACE)" ZAP_FLAGS="$(ZAP_FLAGS)" OUTPUT="$(TEST_ACCEPTANCE_OUTPUT_DIR)" ./hack/deploy-sbo-local.sh
+
+.PHONY: stop-local
+## Stop Local: Stop locally running operator
+stop-local:
+	$(Q)-./hack/stop-sbo-local.sh
```
pedjak, line 380: not seeing this file as a part of PR?

hack/deploy-sbo-local.sh
pmacik: We are ensuring that the SBO is actually started and ready for work. That is what is supposed to be meant by `Service Binding Operator is running` step - the part of the step implementation is checking that the reponse is not `FAILED`.
If something goes wrong as part of the "booting", the script will detect that, returns `FAILED` and that will fail the step and fail-fast the whole suite.

Makefile
pmacik: oops... fixed

pmacik: thx

Makefile
pmacik: I'll try, yes

hack/deploy-sbo-local.sh
pedjak: ok, let's then leave as it is... honestly I am not sure if this improves our confidence that the operator is ready, but that is the topic for another PR. We should come up with a mechanism that would work both in remote and local testing scenarios.

PR #761, opened 2020-11-17
Add acceptance test for PodSpec path example

pratikjagrut:
### Motivation

JIRA TICKET: https://issues.redhat.com/browse/APPSVC-707

### DIFF
--- a/test/acceptance/features/examples.feature
+++ b/test/acceptance/features/examples.feature
@@ -9,6 +9,7 @@ Feature: Verify examples provided in Service Binding Operator github repository
         * Service Binding Operator is running
         * PostgreSQL DB operator is installed
 
+    @openshift
     # https://github.com/redhat-developer/service-binding-operator/blob/master/examples/route_k8s_resource/README.md
     Scenario: Bind an application to an openshift route
         Given The openshift route is present
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ jobs:
         - MINIKUBE_WANTUPDATENOTIFICATION=false
         - MINIKUBE_WANTREPORTERRORPROMPT=false
         # exclude knative and example annotated tests
-        - EXTRA_BEHAVE_ARGS="--tags=~@knative --tags=~@examples"
+        - EXTRA_BEHAVE_ARGS="--tags=~@knative --tags=~@openshift"
         - TEST_ACCEPTANCE_CLI="kubectl"
         - TEST_ACCEPTANCE_START_SBO="remote"
         - SBO_NAMESPACE=operators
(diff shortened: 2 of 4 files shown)


test/acceptance/features/examples.feature
```diff
@@ -53,3 +54,72 @@ Feature: Verify examples provided in Service Binding Operator github repository
...
+                  - group: postgresql.baiju.dev
+                    version: v1alpha1
+                    kind: Database
+                    name: db-pod-spec-path-ex
+                    id: zzz
+                    envVarPrefix: qiye
+            """
+        Then Service Binding "binding-request-pod-spec-path-ex" is ready
+        And Secret "binding-request-pod-spec-path-ex" has been injected in to CR "appconfig-pod-spec-path-ex" of kind "AppConfig" at path "{.spec.spec.containers[0].envFrom[0].secretRef.name}"
+        When Service binding "binding-request-pod-spec-path-ex" is deleted
```
pedjak, line 124: each scenario should have only one `When` block, otherwise it means that we are testing multiple things which is not good.

pedjak (commented):
the added scenario looks like duplication of https://github.com/redhat-developer/service-binding-operator/blob/402153f098f369e605cc0ddbd6b7c08221023288/test/acceptance/features/injectSecretAtCustomSchemaPath.feature#L33 ? why do not we tag this properly?

test/acceptance/features/examples.feature
pratikjagrut: In this scenario, we're testing two things, 1. Binding 2. Unbinding, that's why there are two whens.

pratikjagrut:
> the added scenario looks like duplication of

Yes it's half part is same but another part is the deletion of sbr. Instead of amending the available scenario, we thought to keep automation of examples separately. This was decided after having the discussion with @pmacik.
> 
> https://github.com/redhat-developer/service-binding-operator/blob/402153f098f369e605cc0ddbd6b7c08221023288/test/acceptance/features/injectSecretAtCustomSchemaPath.feature#L33
> 
> ? why do not we tag this properly?

I did not understand what you meant here.

test/acceptance/features/examples.feature
pedjak: fine, but example can contain multiple scenario in it.. Hence, it would be good to split this into multiple scenarios.

pratikjagrut: @pmacik WDYT?

pmacik: @pratikjagrut I agree with @pedjak since the example demonstrate 2 things - binding and unbinding - it should have multiple scenarios, one for each thing. Of course that means repeating a set of initial steps.

pedjak (approved):
please squash

PR #764, opened 2020-11-18
Install SBO by curling the install script

Avni-Sharma:
This PR introduces curl based single line command for install.sh so that users don’t need to clone the repo.


sbose78:
Script gets pulled and executed on my Mac!

/lgtm
/approve

PR #767, opened 2020-11-19
Automate building of SBO bundle and index images usable in CatalogSources

pmacik:
### Motivation

Currently there is no direct way to produce SBO bundle and index images.

Jira issue: [APPSVC-766](https://issues.redhat.com/browse/APPSVC-766)

### Changes

This PR:
* Adds Makefile targets for building such operator bundle index image that can be used in `CatalogSources`.
* Adds `release-operator-bundle` Makefile target to run all the necessary steps to do a release to quay such as building operator, operator bundle and bundle index, and pushing it to quay.
* Adds that release target into Travis job that builds operator bundle for each merge-to-master, so that an index image is build for each merge to master, too.

### Testing

1. Set the following env variables:
* `QUAY_TOKEN`
* `QUAY_USERNAME`
* `OPERATOR_GROUP` ... usually same as Quay username
2. Run `make release-operator-bundle`
3. Create a catalog source referencing the above index image
```sh
kubectl apply -f - << EOD
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: sbo-test-catalog
  namespace: openshift-marketplace
spec:
  sourceType: grpc
  image: REPLACE_IMAGE
  displayName: Community Operators
  publisher: Red Hat
EOD
```
4. Install SBO via OperatorHub or Subscription YAML as usual with the appropriate CSV version

PR #769, opened 2020-11-20
Refactor dev-release and validate-release jobs

pmacik:
### Motivation

* [APPSVC-763](https://issues.redhat.com/browse/APPSVC-763)
* Depends on https://github.com/redhat-developer/service-binding-operator/pull/767

### Changes

This PR:

* Replaces `validate-release` and `dev-release` targets with `test-acceptance-with-bundle` that installs SBO from a given index image and then runs acceptance tests with it remotely. By default it uses upstream's `quay.io/redhat-developer/servicebinding-operator:index` index image (build on each merge to master)
* Enables validating any index image including `operatorhub.io`
* Is used by periodic validation job on OpenShift CI https://github.com/openshift/release/pull/13966
* Refactors `install.sh` script to be parametrized and to be usable for various situations
* Introduces detections of namespaces for operators, OLM and SBO itself (if installed in cluster)
* Re-uses existing catalog source for a given index image instead of creating duplicate sources
* Removes the need to specify SBO's namespace for acceptance tests
* Updates Travis CI jobs accordingly

### Testing
* OpenShift
  - `master` bundle:
```$ make test-acceptance-with-bundle```
  - operatorhub-io bundle:
```$ OPERATOR_INDEX_IMAGE_REF=quay.io/operatorhubio/catalog:latest make test-acceptance-with-bundle```
* minikube
  - `master` bundle:
```$ EXTRA_BEHAVE_ARGS="--tags=~@knative --tags=~@examples" TEST_ACCEPTANCE_CLI=kubectl make test-acceptance-with-bundle```
  - operatorhub-io bundle:
```$ EXTRA_BEHAVE_ARGS="--tags=~@knative --tags=~@examples" TEST_ACCEPTANCE_CLI=kubectl OPERATOR_INDEX_IMAGE_REF=quay.io/operatorhubio/catalog:latest make test-acceptance-with-bundle```

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -499,7 +499,7 @@ push-bundle-to-quay:
 .PHONY: dev-release
 ## validating the operator by installing new quay releases
 dev-release:
-	BUNDLE_VERSION=$(BUNDLE_VERSION) ./hack/dev-release.sh
+	INDEX_IMAGE="quay.io/$(OPERATOR_GROUP)/$(GO_PACKAGE_REPO_NAME)-index:$(BUNDLE_VERSION)" BUNDLE_VERSION=$(BUNDLE_VERSION) ./hack/dev-release.sh
```
pedjak, line 502: Where do we build the index image? Do we really need to have index image per SBO version? I could imaging one index version that gets rebuild with every build to master and its purpose is to keep the last master build.

hack/dev-release.sh
```diff
@@ -3,14 +3,28 @@
 set -e
 set -u
 
-OPERATOR_SOURCE="./test/operator-hub/operator_source.yaml"
+INDEX_IMAGE="${INDEX_IMAGE:-quay.io/$OPERATOR_GROUP/$GO_PACKAGE_REPO_NAME-index:$BUNDLE_VERSION}"
```
pedjak, line 6: Do we ever plan to run this script standalone. i.e. without Makefile?

hack/dev-release.sh
```diff
@@ -3,14 +3,28 @@
...
+  name: example-operators
+  namespace: openshift-marketplace
+spec:
+  sourceType: grpc
+  image: $INDEX_IMAGE
+  displayName: Community Operators
+  publisher: Red Hat
+EOD
+
+./hack/check-catalogsource-pods.sh
```
pedjak, line 27: Since we are reworking things where, what is the purpose of this check? Do we really need it? IMHO, once we add the catalog source, we can immediately apply the subscription, wait for the operator to become available, and then eventually apply either ALM example or some other smoke test.

While we are here - how about an even more radical approach: `dev-release.sh` with an acceptance test that installs SBO and perform some binding? That would reduce the maintenance complexity a lot IMHO.

Makefile
pmacik: The index image would be build by Travis CI per merge-to-master. Once https://github.com/redhat-developer/service-binding-operator/pull/767 is merged

hack/dev-release.sh
pmacik: probably not...

Makefile
pmacik: Let's move the discussion about the index image to https://github.com/redhat-developer/service-binding-operator/pull/767

install.sh
```diff
@@ -1,63 +1,86 @@
...
 
-#3) Start a kubernetes cluster. Any Kubernetes cluster can be used as well. The script assumes that a Kubernetes is up and running and the user has logged into it.
+if [ "$OPENSHIFT" = "true" ]; then
+  INDEX_IMAGE=${INDEX_IMAGE:-}
+  if [ -n "$INDEX_IMAGE" ]; then
+    CATSRC_NAME=${CATSRC_NAME:-$(echo -n "$INDEX_IMAGE" | sed -e 's,[/:\.],-,g')}
+  else
+    CATSRC_NAME=${CATSRC_NAME:-community-operators}
+  fi
+  CATSRC_NAMESPACE="${CATSRC_NAMESPACE:-openshift-marketplace}"
```
pedjak, line 13: I do not think it is important to know if we are running against openshift or not - instead we should detect if OLM is present or not, and install it. This can be done by performing following step:

* detect if `csv/packageserver` is available and finding out the namespace
* if available, we can detect the namespace where CatalogSources live
* detect if `openshift-operators` or `operators` namespace exit

If OLM is installed, we could detect if CatalogSource with the given index image is already installed and skip its installation if it is.

We should allow non-interactive registry logins and we need to install credentials to the cluster and add to the proper service accounts

install.sh
```diff
@@ -1,63 +1,86 @@
...
+  displayName: SBO Test
   publisher: Red Hat
-  updateStrategy:
-    registryPoll:
-      interval: 10m0s
 EOD
-#Apply subscription for subscribing to the beta channel of the service binding operator
+fi
+
+CSV_VERSION=$(./hack/get-current-csv.sh)
```
pedjak, line 62: we do not need this because we always want to install latest version available on the channel.

install.sh
```diff
@@ -1,63 +1,86 @@
...
 spec:
-  channel: 'beta'
+  channel: beta
   installPlanApproval: Automatic
   name: service-binding-operator
-  source: community-operators
-  sourceNamespace: olm
+  source: $CATSRC_NAME
+  sourceNamespace: $CATSRC_NAMESPACE
+  startingCSV: $CSV_VERSION
```
pedjak, line 76: see above, it does not have any effect since `installPlanApproval` is set to `Automatic`.

hack/validate-community-release.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -e
```
pedjak, line 1: do we really need this file anymore?

pedjak (changes requested):

install.sh
pmacik: In order to install the operator via subscription, CSV version is needed. This script looks into the package manifest for a given catalog source and it's channel and extract that version from it (that is the "latest version available on the channel" you mentined)... the above line is missing the arguments for catalog source and channel which I have fixed in the recent commit.

hack/validate-community-release.sh
pmacik: No, I've removed it

install.sh
pmacik: Right, I'll update it

install.sh
pmacik: Take it back... you are correct, we do not need CSV in subscription in our case.

install.sh
pmacik: +1

Avni-Sharma:
Hi @pmacik  Would this PR aim on pushing the operator artifacts to https://github.com/redhat-developer/service-binding-operator-manifests in new bundle format ? i.e having metadata/ and manifests/ under each version/sha directory .

install.sh
```diff
@@ -7,41 +7,63 @@ OPERATOR_CHANNEL=${OPERATOR_CHANNEL:-beta}
...
-  #6) Add that pull secret to the default account in olm namespace
-  kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "sbo-operators-secrets"}]}' -n=olm
+    #6) Add that pull secret to the default account in olm namespace
+    kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "sbo-operators-secrets"}]}' -n=olm
+  fi
+}
+
+if [ "$(kubectl get crd clusterserviceversions.operators.coreos.com &>1 || echo false)" == "false" ]; then
+  echo no csv
+  INSTALL_OLM="true"
```
pedjak, line 46: it is not used anywhere.

install.sh
```diff
@@ -7,41 +7,63 @@ OPERATOR_CHANNEL=${OPERATOR_CHANNEL:-beta}
...
+    kubectl create secret generic sbo-operators-secrets -n olm --from-file=.dockerconfigjson=$DOCKER_CFG --type=kubernetes.io/dockerconfigjson
 
-  #6) Add that pull secret to the default account in olm namespace
-  kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "sbo-operators-secrets"}]}' -n=olm
+    #6) Add that pull secret to the default account in olm namespace
+    kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "sbo-operators-secrets"}]}' -n=olm
+  fi
+}
+
+if [ "$(kubectl get crd clusterserviceversions.operators.coreos.com &>1 || echo false)" == "false" ]; then
```
pedjak, line 44: it can be simplified as:

```
kubectl get crd clusterserviceversions || install_olm
```

install.sh
```diff
@@ -7,41 +7,63 @@ OPERATOR_CHANNEL=${OPERATOR_CHANNEL:-beta}
...
+  echo no csv
+  INSTALL_OLM="true"
+  install_olm
 fi
 
+PACKAGESERVER_INFO=($(kubectl get clusterserviceversions.operators.coreos.com --all-namespaces| grep packageserver))
+echo PS_NAMESPACE=${PACKAGESERVER_INFO[0]}
+
+CATSRC_NAMESPACE="${CATSRC_NAMESPACE:-$(kubectl get catalogsources.operators.coreos.com --all-namespaces -o jsonpath='{.items[0].metadata.namespace}' --ignore-not-found=true)}"
```
pedjak, line 53: I would say that catalogsource live in the same as package server?

install.sh
```diff
@@ -7,41 +7,63 @@ OPERATOR_CHANNEL=${OPERATOR_CHANNEL:-beta}
...
+OPERATOR_NAMESPACE="${OPERATOR_NAMESPACE:-$(./hack/detect-operators-namespace.sh)}"
```
pedjak, line 54: it would be great that `install.sh` script does not depend on any other script in the repo, because potentially we would like to publish this script on our relase pages.

test/acceptance/features/environment.py
```diff
@@ -39,9 +39,12 @@ def before_all(_context):
...
         assert not os.getenv("TEST_ACCEPTANCE_SBO_STARTED").startswith("FAILED"), "TEST_ACCEPTANCE_SBO_STARTED shoud not be FAILED."
     elif start_sbo == "remote":
-        sbo_namespace = os.getenv("SBO_NAMESPACE")
-        assert (sbo_namespace is not None) and (sbo_namespace != ""), f"SBO_NAMESPACE is required but it is not set or it is empty: {sbo_namespace}"
-        _context.sbo_namespace = sbo_namespace
+        output, code = cmd.run(
+            f"{ctx.cli} get deployment --all-namespaces -o json"
+            + " | jq -rc '.items[] | select(.metadata.name == \"service-binding-operator\").metadata.namespace'")
+        assert code == 0, f"Non-zero return code while trying to detect namespace for SBO: {output}"
+        assert (output is not None) and (output != ""), "Unable to find SBO's deployment in any namespace."
```
pedjak, line 46: can `output` really be `None` here, given that we execute external command? Do we wanto trim the output perhaps?

test/acceptance/features/steps/openshift.py
```diff
@@ -397,6 +397,14 @@ def search_resource_lst_in_namespace(self, resource_plural, name_pattern, namesp
         print('Resource list is empty under namespace - {}'.format(namespace))
         return None
 
+    def search_namespace_for_resource(self, resource_plural, name):
+        output, code = self.cmd.run(
+            f"{ctx.cli} get {resource_plural} --all-namespaces -o json"
+            + f" | jq -rc '.items[] | select(.metadata.name == \"{name}\").metadata.namespace'")
+        assert code == 0, f"Non-zero return code while trying to detect namespace for {resource_plural} '{name}': {output}"
+        assert (output is not None) and (output != ""), f"Unable to find {resource_plural} '{name}' in any namespace."
```
pedjak, line 405: same question as above

test/acceptance/features/steps/openshift.py
```diff
@@ -397,6 +397,14 @@ def search_resource_lst_in_namespace(self, resource_plural, name_pattern, namesp
         print('Resource list is empty under namespace - {}'.format(namespace))
         return None
 
+    def search_namespace_for_resource(self, resource_plural, name):
```
pedjak, line 400: perhaps to name the method as `lookup_namespace_for_resource`?

test/acceptance/features/steps/servicebindingoperator.py
```diff
@@ -14,34 +14,27 @@ def __init__(self,  name="service-binding-operator", namespace="openshift-operat
...
-        assert output is not None, f"Unable to find role with name that maches '{name_pattern}' in namespace '{self.namespace}': {output}"
-
-        output = self.openshift.search_resource_in_namespace("serviceaccounts", name_pattern, self.namespace)
-        assert output is not None,  f"Unable to find service account with name that maches '{name_pattern}' in namespace '{self.namespace}': {output}"
-
         return True
 
+    def check_deployment(self):
+        sbo_namespace = self.openshift.search_namespace_for_resource("deployments", "service-binding-operator")
+        return sbo_namespace is not None and sbo_namespace != ""
```
pedjak, line 24: `search_namespace_for_resource` return only non-empty string, otherwise, assertion happens.. therefore we do not need to have this check here, i.e. the result is always `true`.

pedjak (commented):
look good, just a few questions.

test/acceptance/features/steps/openshift.py
pmacik: +1

test/acceptance/features/environment.py
pmacik: +1

test/acceptance/features/steps/servicebindingoperator.py
pmacik: It should return name of namespace or `None`, I fixed that...

install.sh
```diff
@@ -1,47 +1,49 @@
...
-#On minikube you can alternatively install OLM by running `minikube addons enable olm`
+OPNS="$(kubectl get namespaces -o name | grep -E '.*/(.+-)?operators$' || test $? = 1)"
+OPERATOR_NAMESPACE=${OPNS#"namespace/"}
 
-#5) Create a new image pull secret out of your local .docker/config.json file
-if [ -r "$DOCKER_CFG" ]; then
-  kubectl create secret generic sbo-operators-secrets -n olm --from-file=.dockerconfigjson=$DOCKER_CFG --type=kubernetes.io/dockerconfigjson
+if [ -n "$OPERATOR_INDEX_IMAGE" ]; then
+  if [ -r "$DOCKER_CFG" ]; then
+    kubectl create secret generic sbo-operators-secrets -n $CATSRC_NAMESPACE --from-file=.dockerconfigjson=$DOCKER_CFG --type=kubernetes.io/dockerconfigjson --dry-run=client -o json | kubectl apply -f -
```
pedjak, line 19: why do we need to use `--dry-run` option now and pass the result to another `kubectl apply`?

install.sh
```diff
@@ -51,28 +53,31 @@ spec:
...
   name: service-binding-operator
-  namespace: operators
+  namespace: $OPERATOR_NAMESPACE
 spec:
   channel: $OPERATOR_CHANNEL
   installPlanApproval: Automatic
   name: service-binding-operator
-  source: sbo-operators
-  sourceNamespace: olm
+  source: $CATSRC_NAME
```
pedjak, line 71: what is the default behaviour when the script is invoked as

```
./install.sh
```

would it work at all?

pmacik: it does, now

install.sh
```diff
@@ -51,28 +53,42 @@ spec:
     registryPoll:
       interval: 10m0s
 EOD
+  fi
+else
+  CATSRC_CANDIDATES=(
+    operatorhubio-catalog
+    community-operators
```
pedjak, line 60: we should not look into `community-operators` anymore, but instead in case of openshift we should look in `redhat-operators`

pedjak (changes requested):

install.sh
pmacik: Indeed, thx.

PR #774, opened 2020-11-26
Increase Travis waiting to 30mins

pedjak:
By default Travis cancel the job if no new output is produced within 10 minutes. Some of our acceptance test steps can be slower on some infrastructures/VMs. Hence, let's increase the timeout to prevent premature job cancellation.


DhritiShikhar:
@pedjak Why was this change done?
It's good to capture this information in PR description.

pedjak:
> It's good to capture this information in PR description.

Thanks @DhritiShikhar - I have added the description.

Issue #646, opened 2020-08-27
Troubles getting binding:env:object:secret annotation to work


durera:
CRD currently looks something like this (experimenting with a minimal CR that holds mongodb connection properties that I'm trying to use servicebindingrequest to get into a simple demo app):

in namespace 1 we have the CR, and a secret:

The CRD:
```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  ...
  annotations:
    servicebindingoperator.redhat.io/status.configDb: binding:env:attribute
    servicebindingoperator.redhat.io/status.credentials-password: binding:env:object:secret
    servicebindingoperator.redhat.io/status.credentials-username: binding:env:object:secret
    servicebindingoperator.redhat.io/status.hosts: binding:env:attribute
  ...
```

The status object of the CR looks like this:
```
status:
  conditions:
  - ansibleResult:
      changed: 2
      completion: 2020-08-27T17:54:00.33111
      failures: 0
      ok: 8
      skipped: 0
    lastTransitionTime: 2020-08-27T17:53:36Z
    message: Awaiting next reconciliation
    reason: Successful
    status: "True"
    type: Running
  configDb: admin
  credentials: secretname
  hosts:
  - host: <removed>
    port: <removed>
  - host: <removed>
    port: <removed>
```

Where `secretname` is a secret that contains the name of a secret that hold username/password in it's data.  

Over in namespace 2, all of the `binding:env:attribute` stuff "just works" -- the servicebindingrequest created and injected the data into my deployment using the intermediary secret:

```
apiVersion: v1
data:
  MONGODB_CONFIGDB: <removed>
  MONGODB_HOSTS_0_HOST: <removed>
  MONGODB_HOSTS_0_PORT: <removed>
  MONGODB_HOSTS_1_HOST: <removed>
  MONGODB_HOSTS_1_PORT: <removed>
kind: Secret
...
```

But I cannot for the life of me get the username and password pulled out of the referenced secret.  

[...]

PR #787, opened 2020-12-02
Add acceptance test: Verify pulling data from a secret living in another namespace

pratikjagrut:
### Motivation
Verifies GH issue #646 
JIRA TICKET: https://issues.redhat.com/browse/APPSVC-762

### DIFF
--- a/test/acceptance/features/bindAppToServiceUsingSecret.feature
+++ b/test/acceptance/features/bindAppToServiceUsingSecret.feature
@@ -395,11 +395,8 @@ Feature: Bind values from a secret referred in backing service resource
         And The application env var "BACKEND_USERNAME" has value "AzureDiamond"
         And The application env var "BACKEND_PASSWORD" has value "hunter2"
 
-    Scenario: Verify pulling data from a secret living in another namespace
-        Binding definition is declared on service CRD.
-        Given OLM Operator "backend" is running
-        * Generic test application "ssa-3" is running
-        * The Custom Resource Definition is present
+    Scenario: Inject into app all keys from a secret existing in a namespace different from the service resource
+        Given The Custom Resource Definition is present
             """
             apiVersion: apiextensions.k8s.io/v1beta1
             kind: CustomResourceDefinition
@@ -449,6 +446,7 @@ Feature: Bind values from a secret referred in backing service resource
             status:
                 credentials: ssa-3-secret
             """
+        * Generic test application "ssa-3" is running
         When Service Binding is applied
             """
             apiVersion: operators.coreos.com/v1alpha1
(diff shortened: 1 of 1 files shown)


test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -394,3 +394,80 @@ Feature: Bind values from a config map referred in backing service resource
         Then Service Binding "ssd-2" is ready
         And The application env var "BACKEND_USERNAME" has value "AzureDiamond"
         And The application env var "BACKEND_PASSWORD" has value "hunter2"
+
+    Scenario: Verify pulling data from a secret living in another namespace
```
pedjak, line 398: the ticket is about verifying, but the scenario name itself should describe the intention, i.e. how system should work from user perspective. Thus:

```
Inject into app all keys from a secret existing in a namespace different from the service resource
```

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -394,3 +394,80 @@ Feature: Bind values from a config map referred in backing service resource
...
+        Binding definition is declared on service CRD.
```
pedjak, line 399: I think this line could be removed.

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -394,3 +394,80 @@ Feature: Bind values from a config map referred in backing service resource
...
+        Given OLM Operator "backend" is running
```
pedjak, line 400: If you use this step, then you should not install CustomResourceDefinition under, or vice versa. I think we need later.

pedjak (changes requested):

test/acceptance/features/bindAppToServiceUsingSecret.feature
pratikjagrut: Done.

test/acceptance/features/bindAppToServiceUsingSecret.feature
pratikjagrut: Removed OLM step.

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -394,3 +394,78 @@ Feature: Bind values from a config map referred in backing service resource
...
+                username: AzureDiamond
+                password: hunter2
+            """
+        * The Custom Resource is present
+            """
+            apiVersion: stable.example.com/v1
+            kind: Backend
+            metadata:
+                name: ssa-3-service
+                namespace: backend-services
```
DhritiShikhar, line 445: I can see that service `ssa-3-service` lives in `backend-services` namespace
and secret `ssa-3-secret` lives in `backend-services` namespace.

Both service and secret are in the same namespace... but the title of the PR says that we want to verify a situation where service and secret live in different namespace.

pratikjagrut: Sorry, @DhritiShikhar I'll update the title service and secret should be in same ns. This test is based on issue #646.

pedjak (approved):

PR #789, opened 2020-12-03
add a parameter max-concurrent-reconciles to set the reconile workers

qibobo:


isutton:
Hi @qibobo, thanks for this PR!

We've discussed internally about this matter very recently, so it was a good timing.

What do you think of changing the implementation to use `pflag`, which already provides the command line flags? 

We could then use the environment variable to set this flag (or perhaps `pflag` already does the hard work of getting flags from environment variables?).

baijum:
OLM subscription has support for environment variables: https://github.com/operator-framework/operator-lifecycle-https://github.com/operator-framework/operator-lifecycle-manager/blob/bf9eeb14ae0f930fdd2a0a91c7769569efed0410/deploy/upstream/manifests/0.17.0/0000_50_olm_00-subscriptions.crd.yaml#L74

qibobo:
@isutton @baijum @pedjak 
Either pflag or env var is OK for us. Which option do you prefer?

isutton:
> @isutton @baijum @pedjak
> Either pflag or env var is OK for us. Which option do you prefer?

`pflag`.

baijum:
> > @isutton @baijum @pedjak
> > Either pflag or env var is OK for us. Which option do you prefer?
> 
> `pflag`.

How a command-line flag is going to be used for an operator? I don't see support in OLM subscription.

isutton:
> How a command-line flag is going to be used for an operator? I don't see support in OLM subscription.

If you see in our `main` function, there is already support for `pflag` which is used to switch the behavior of the operator; what I am suggesting is to use that already in place mechanism instead of peeking directly in the environment variables, since we might want to override this when running locally to test, debug, and whatnot.

Another point is that `pflag` underneath already uses `cobra` it seems, and it wouldn't surprise me if `viper` is being used as well (which gives the bridge between environment variables and the program flags): this means that by adding the flag it might get the environment variable handling *for free*.

pedjak:
Making the number of workers configurable is a good idea - both having env var and command line arg. However, we should give an opportunity to user to configure that in OLM way as well. `Subscription` talks about possibility to pass env vars to deployments:

https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/subscription_types.go#L42

@qibobo can you please investigate if this really works?

qibobo:
@pedjak 
I use service-binding-operator without OLM, so I do not have a way to verify whether it works with OLM. Could you do it?

isutton (commented):
That looks fine, thanks for the changes!

To make sure we're on the same page: to modify the number of workers the cluster admin would change add a `--max-concurrent-reconciles=$WORKERS` switch to the SBO deployment comment; do you agree?

Another improvement would be having the number of workers being the number of CPUs available to the container instead of 1; this would enable a decent threadiness by default, but I'm unsure if this PR would be the right place to do so. What do you think?

pkg/controller/servicebinding/flag.go
```diff
@@ -0,0 +1,17 @@
+package servicebinding
```
pedjak, line 1: I would not introduce the package just for this.. please move this file under `cmd/manager`.

pedjak:
> Another improvement would be having the number of workers being the number of CPUs available to the container instead of 1; this would enable a decent threadiness by default, but I'm unsure if this PR would be the right place to do so. What do you think?

I would keep this for another PR.

pkg/controller/servicebinding/flag.go
```diff
@@ -0,0 +1,17 @@
...
+import "github.com/spf13/pflag"
+
+var (
+	sboFlagSet              *pflag.FlagSet
+	maxConcurrentReconciles int
+)
+
+func init() {
+	sboFlagSet = pflag.NewFlagSet("sbo", pflag.ExitOnError)
+	sboFlagSet.IntVar(&maxConcurrentReconciles, "max-concurrent-reconciles", 1, "max-concurrent-reconciles is the maximum number of concurrent Reconciles which can be run. Defaults to 1.")
```
pedjak, line 12: we should read the default value from an env var here, and assume 1 if the var is not found.

pedjak (changes requested):
apart from having the commnad line arg, we should have ability to configure max workers via env var as well.

pkg/controller/servicebinding/flag.go
isutton: > apart from having the commnad line arg, we should have ability to configure max workers via env var as well.

Sure, but this would mean change `flag` if it doesn't already support it, which we do not want to do in this work item; this leaves us with the option of introducing a broader support for communicating environment variables in another PR.

pedjak: Instead of fixing the value to `1`, we should try to read it from an env var. This PR should be about ability to configure max number of workers, not about additing just command line arg for that.

isutton: In my perspective, if I had to decide what goes on, from a performance point of view it is much better to have the default value being the number of CPUs, since it will reflect any changes in the deployment automatically leaving the environment variable or command line switch as a resource for the developer when debugging or stressing the software.

Other than that, it seems to be trivial to extract the number of CPUs available: http://play.golang.org/p/cuaf2ZHLIx.

isutton: Here there is an example on how to extend `pflag` with environment variables: https://scene-si.org/2020/04/28/extending-pflag-with-environment-variables/.

isutton: Let's consider the number of workers being CPU cores + 1 the default, and extend pflag with environment variables in another changelist.

isutton:
I think this is good enough for what it is, so I'm approving to build on further enhancements.

/lgtm
/approve

PR #794, opened 2020-12-08
Add `prepare-operatorhub-pr` target

pedjak:
The target extracts files from published bundle image (default: for the current HEAD commit)
and place them into `out/operatorhub-pr-files/service-binding-operator` folder so that they
could be easily copied into local clone of `github.com/operator-framework/community-operators`,
creating PR afterwards.

### Motivation

The added target makes the operator publishing to the upstream easier:

* Use already bundles published from master, no need to build any images
* The existing bundles are already verified, i.e. acceptance tests passed on
  them
* The bundles refer images using shas, so the published upstream operator
has already support for air-gapped installations


sbose78:
Neat!

PR #802, opened 2020-12-14
Rename customEnvVar to mappings

baijum:
Also rename envVarPrefix fields as well to namePrefix

Ref. https://issues.redhat.com/browse/APPSVC-538


examples/knative_postgresql_customvar/README.md
```diff
@@ -183,7 +183,7 @@ There are 3 parts in the request:
 
 * `application` - used to search for the application based on the name that we set earlier and the `group`, `version` and `resource` of the application to be a knative `Service`.
 * `services` - used to find the backing service - our operator-backed DB instance called `db-demo`.
-* `customEnvVar` - used to create custom environment variables constructed using a templating engine from out-of-the-box bound information.
+* `mappings` - used to create custom environment variables constructed using a templating engine from out-of-the-box bound information.
```
Avni-Sharma, line 186: only environment vars?

sbose78: Good point, they do land up as "files" too, I think?

sbose78:
Please do let CodeEngine folks know :)

pkg/apis/operators/v1alpha1/servicebinding_types.go
```diff
@@ -19,13 +19,13 @@ type ServiceBindingSpec struct {
...
+	// NamePrefix is the prefix for environment variables or file name
 	// +optional
-	EnvVarPrefix string `json:"envVarPrefix,omitempty"`
+	NamePrefix string `json:"namePrefix,omitempty"`
 
-	// Custom env variables
+	// Custom mappings
 	// +optional
-	CustomEnvVar []corev1.EnvVar `json:"customEnvVar,omitempty"`
+	Mappings []corev1.EnvVar `json:"mappings,omitempty"`
```
pedjak, line 28: We should change the type here: `corev1.EnvVar` should not be the type we use - it should be something similar to https://github.com/k8s-service-bindings/spec/blob/master/internal/service.binding/v1alpha2/service_binding.go#L69

pedjak (changes requested):

pkg/apis/operators/v1alpha1/servicebinding_types.go
sbose78: +1

baijum: done

examples/knative_postgresql_customvar/README.md
baijum: done

pkg/apis/operators/v1alpha1/servicebinding_types.go
```diff
@@ -47,6 +47,14 @@ type ServiceBindingSpec struct {
 	BindAsFiles bool `json:"bindAsFiles,omitempty"`
 }
 
+// ServiceBindingMapping defines a mapping from the existing collection of Secret values to a new Secret entry.
```
pedjak, line 50: Can we remove `ServiceBinding` prefix from the struct name, it is redundant. Also, can we tailor the description to user facing abstraction, i.e. binding:

```
defines a new binding from set of existing bindings
```

pkg/apis/operators/v1alpha1/servicebinding_types.go
```diff
@@ -47,6 +47,14 @@ type ServiceBindingSpec struct {
...
+type ServiceBindingMapping struct {
+	// Name is the name of the mapped Secret entry
```
pedjak, line 52: How about: `Name is the name of new binding`

pkg/apis/operators/v1alpha1/servicebinding_types.go
```diff
@@ -47,6 +47,14 @@ type ServiceBindingSpec struct {
...
+	Name string `json:"name"`
+	// Value is the value of the new Secret entry.
```
pedjak, line 54: We should state here that the value is actually template and rendered value gets injected into application.

deploy/crds/operators.coreos.com_servicebindings_crd.yaml
```diff
@@ -123,12 +123,21 @@ spec:
...
+                variables from different subresources owned by backing operator CR.
+              type: boolean
+            namePrefix:
+              description: NamePrefix is the prefix for environment variables
+              type: string
+            mappings:
+              description: Custom mappings
               items:
-                description: EnvVar represents an environment variable present in
+                description: Mappings represents an environment variable or file name present in
```
pedjak, line 140: CRD file need to regenerated - it does not match the latest state in `servicebinding_types.go`

pedjak (changes requested):

Issue #799, opened 2020-12-10
Non-OLM cluster support


pedjak:
Service binding operator should work even on plain k8s clusters having no OLM installed.

* It should be possible to install the operator with kubectl apply -f release.yaml where release.yaml contains all needed resources.
*  Such release.yaml file should be generated in automated way
* It should be documented
* Acceptance tests should be run against such cluster

pedjak:
Done in https://github.com/redhat-developer/service-binding-operator/pull/806

PR #806, opened 2020-12-16
Generating release.yaml

Avni-Sharma:
Fixes - https://github.com/redhat-developer/service-binding-operator/issues/799
https://issues.redhat.com/browse/APPSVC-769

The release manifest file can be created by running the make target `release-manifests` . On running the make release-manifests target a release.yaml file is produced in the /out dir. This release.yaml can now be used to install the SBO on a vanilla k8s.
A GitHub action job name `acceptance_without_olm` is configured to test the release manifest file on a vanilla k8s cluster.

### DIFF
--- a/.github/actions/setup-cli/action.yaml
+++ b/.github/actions/setup-cli/action.yaml
@@ -1,0 +1,35 @@
+# action.yml
+name: 'Setup Tools'
+description: 'Setup CLI configuration with kubectl operator-sdk opm minikube, setup path, kubeconfig, minikube and docker info '
+runs:
+  using: "composite"
+  steps:
+    - id: setup-path
+      run: |
+        mkdir -p $GITHUB_WORKSPACE/bin/
+        echo "PATH=$PATH:$GITHUB_WORKSPACE/bin/" >> $GITHUB_ENV
+      shell: bash
+
+    - run: $GITHUB_ACTION_PATH/setup_tools.sh
+      shell: bash
+
+    - id: set-kubeconfig
+      run: |
+        KUBECONFIG=$GITHUB_WORKSPACE/miniconfig
+        echo "KUBECONFIG=$KUBECONFIG" >> $GITHUB_ENV
+        touch $KUBECONFIG
+      shell: bash
+
+    - id: setup-minikube
+      run: |
+        ./hack/start-minikube.sh start --kubernetes-version=v${K8S_VERSION} --driver=$CONTAINER_RUNTIME --cpus $(nproc) --memory 5g
+      shell: bash
+
+    - id: docker-info
+      run: |
+        eval $(minikube docker-env)
+        docker ps
+        kubectl get nodes -o yaml
+        kubectl cluster-info
+        docker info
+      shell: bash
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -7,11 +7,19 @@ on:
 
 env:
   GO111MODULE: on
+  SDK_VERSION: "1.3.0"
+  OPM_VERSION: "1.15.2"
+  MINIKUBE_WANTUPDATENOTIFICATION: false
+  MINIKUBE_WANTREPORTERRORPROMPT: false
+  K8S_VERSION: "1.19.2"
+  MINIKUBE_VERSION: "1.15.1"
+  TEST_ACCEPTANCE_CLI: "kubectl"
+  CONTAINER_RUNTIME: "docker"
 
 jobs:
   lint:
     name: Code Quality
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
       - name: Set up Go
@@ -33,7 +41,7 @@ jobs:
 
   unit:
     name: Unit Tests with Code coverage
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
       - name: Set up Go
@@ -56,28 +64,49 @@ jobs:
           fail_ci_if_error: true
 
   acceptance:
-    name: Acceptance Tests with Kubernetes
-    runs-on: ubuntu-latest
+    name: Acceptance Tests with Kubernetes and using OLM
+    runs-on: ubuntu-20.04
 
     env:
-      SDK_VERSION: "1.3.0"
-      OPM_VERSION: "1.15.2"
-      MINIKUBE_WANTUPDATENOTIFICATION: false
-      MINIKUBE_WANTREPORTERRORPROMPT: false
-      K8S_VERSION: "1.19.2"
-      MINIKUBE_VERSION: "1.15.1"
       EXTRA_BEHAVE_ARGS: "--tags=~@knative --tags=~@openshift"
-      TEST_ACCEPTANCE_CLI: "kubectl"
-      CONTAINER_RUNTIME: "docker"
 
     steps:
       - name: Checkout Git Repository
         uses: actions/checkout@v2
 
-      - name: Set up PATH
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: "^1.15.6"
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.7"
+          architecture: "x64"
+
+      - name: Setup-cli
+        uses: ./.github/actions/setup-cli
+
+      - name: Acceptance tests
+        timeout-minutes: 60
         run: |
-          mkdir -p $GITHUB_WORKSPACE/bin/
-          echo "PATH=$PATH:$GITHUB_WORKSPACE/bin/" >> $GITHUB_ENV
+          eval $(minikube docker-env)
+          make OPERATOR_REPO_REF=$(minikube ip):5000/sbo SKIP_REGISTRY_LOGIN=true release-operator -o registry-login test-acceptance-with-bundle
+
+
+  acceptance_without_olm:
+    name: Acceptance tests running on Kubernetes without using OLM
+    runs-on: ubuntu-20.04
+
+    env:
+      EXTRA_BEHAVE_ARGS: "--tags=~@knative --tags=~@openshift --tags=~@olm"
+      UMOCI_VERSION: "0.4.5"
+      OPERATOR_VERSION: "0.4.0"
+
+    steps:
+      - name: Checkout Git Repository
+        uses: actions/checkout@v2
 
       - name: Set up Go
         uses: actions/setup-go@v2
@@ -91,43 +120,18 @@ jobs:
           architecture: "x64"
 
       - name: Set up CLI
+        uses: ./.github/actions/setup-cli
+
+      - name: Setup umoci cli
         run: |
-          curl -Lo operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v${SDK_VERSION}/operator-sdk_linux_amd64
-          chmod +x operator-sdk
-          mv -v operator-sdk $GITHUB_WORKSPACE/bin/
+          curl -Lo umoci https://github.com/opencontainers/umoci/releases/download/v${UMOCI_VERSION}/umoci.amd64
+          chmod +x umoci
+          mv -v umoci $GITHUB_WORKSPACE/bin/
 
-          curl -Lo opm https://github.com/operator-framework/operator-registry/releases/download/v${OPM_VERSION}/linux-amd64-opm
-          chmod +x opm
-          mv -v opm $GITHUB_WORKSPACE/bin/
-
-          curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v${K8S_VERSION}/bin/linux/amd64/kubectl
-          chmod +x kubectl
-          mv -v kubectl $GITHUB_WORKSPACE/bin/
-
-          curl -Lo minikube https://storage.googleapis.com/minikube/releases/v${MINIKUBE_VERSION}/minikube-linux-amd64
-          chmod +x minikube
-          mv -v minikube $GITHUB_WORKSPACE/bin/
-
-      - name: Set KUBECONFIG
-        run: |
-          KUBECONFIG=$GITHUB_WORKSPACE/miniconfig
-          echo "KUBECONFIG=$KUBECONFIG" >> $GITHUB_ENV
-          touch $KUBECONFIG
-
-      - name: Set up minikube
-        run: |
-          ./hack/start-minikube.sh start --kubernetes-version=v${K8S_VERSION} --driver=$CONTAINER_RUNTIME --cpus $(nproc) --memory 5g
-
-      - name: Docker Info
-        run: |
-          eval $(minikube docker-env)
-          docker ps
-          kubectl get nodes -o yaml
-          kubectl cluster-info
-          docker info
-
-      - name: Acceptance tests
+      - name: Acceptance tests against vanilla k8s without OLM
         timeout-minutes: 60
         run: |
           eval $(minikube docker-env)
-          make OPERATOR_REPO_REF=$(minikube ip):5000/sbo SKIP_REGISTRY_LOGIN=true release-operator -o registry-login test-acceptance-with-bundle
+          make OPERATOR_REPO_REF=$(minikube ip):5000/sbo SKIP_REGISTRY_LOGIN=true release-operator -o registry-login release-manifests
+          kubectl apply -f out/release.yaml
+          make test-acceptance
--- a/Makefile
+++ b/Makefile
@@ -263,7 +263,7 @@ release-operator: push-image push-bundle-image push-index-image
 ## prepare files for OperatorHub PR
 ## use this target when the operator needs to be released as upstream operator
 prepare-operatorhub-pr:
-	./hack/prepare-operatorhub-pr.sh $(OPERATOR_VERSION) $(OPERATOR_BUNDLE_IMAGE_REF)
+	./hack/prepare-operatorhub-pr.sh $(VERSION) $(OPERATOR_BUNDLE_IMAGE_REF)
 
 .PHONY: deploy-from-index-image
 ## deploy the operator from a given index image
@@ -301,3 +301,9 @@ test-acceptance-generate-report:
 ## Serves acceptance tests report at http://localhost:8088
 test-acceptance-serve-report:
 	$(Q)CONTAINER_RUNTIME=$(CONTAINER_RUNTIME) $(HACK_DIR)/allure-report.sh serve
+
+.PHONY: release-manifests
+## prepare a manifest file for releasing operator on vanilla k8s cluster
+release-manifests: setup-venv prepare-operatorhub-pr
+	$(Q)$(PYTHON_VENV_DIR)/bin/pip install -q -r hack/check-python/requirements.txt
+	$(Q)$(PYTHON_VENV_DIR)/bin/python3 ./hack/release-manifest.py  $(OUTPUT_DIR)/operatorhub-pr-files/service-binding-operator/$(VERSION)/manifests/
--- a/hack/prepare-operatorhub-pr.sh
+++ b/hack/prepare-operatorhub-pr.sh
@@ -9,7 +9,7 @@
 
 TMP_OCI_PATH=$(mktemp -d out/sbo-bundle-oci.XXX)
 
-skopeo copy docker://$2 oci:$TMP_OCI_PATH:bundle
+skopeo copy --src-tls-verify=false docker://$2 oci:$TMP_OCI_PATH:bundle
 umoci unpack --image $TMP_OCI_PATH:bundle --rootless ${TMP_OCI_PATH}-unpacked
 rm -rf out/operatorhub-pr-files
 mkdir out/operatorhub-pr-files/service-binding-operator -p
--- a/hack/release-manifest.py
+++ b/hack/release-manifest.py
@@ -1,0 +1,113 @@
+import os
+import yaml
+import sys
+from pathlib import Path
+
+MANIFEST_DIR = str(sys.argv[1])
+path = os.getcwd()
+release_yaml = path + "/out/release.yaml"
+Path(release_yaml).touch()
+
+def release_manifest():
+    res = list()
+    ns = create_service_binding_operator_ns()
+    res.append(ns)
+    for filename in os.listdir(MANIFEST_DIR):
+        with open(os.path.join(MANIFEST_DIR, filename), 'r') as f:
+            data = yaml.load(f)
+        if data["kind"] == "ClusterServiceVersion":
+            cluster_role = create_cluster_role_yaml(os.path.join(MANIFEST_DIR, filename))
+            res.append(cluster_role)
+            depl = create_operator_deployment(os.path.join(MANIFEST_DIR, filename))
+            res.append(depl)
+        elif data["kind"] == "ConfigMap" or data["kind"] == "Service":
+            metadata  = data["metadata"]
+            metadata.update({"namespace" : 'service-binding-operator'})
+            res.append(data)
+        else:
+            with open(os.path.join(MANIFEST_DIR, filename), "r") as stream:
+                res.extend(list(yaml.safe_load_all(stream)))
+    clusterrolebinding = create_clusterrolebinding_yaml()
+    res.append(clusterrolebinding)
+    service_account = create_service_account()
+    res.append(service_account)
+    with open(release_yaml, "a") as stream:
+        yaml.dump_all(
+            res,
+            stream,
+            default_flow_style=False
+        )  
+
+def create_cluster_role_yaml(csv):
+    role_data = {
+    "apiVersion" : 'rbac.authorization.k8s.io/v1',
+    "kind" : 'ClusterRole',
+    "metadata" : {
+        "name" : 'service-binding-operator',
+    },
+    }
+    with open(csv, 'r') as f:
+        data = yaml.load(f)
+    roles = data["spec"]["install"]["spec"]["clusterPermissions"][0]
+    roles.pop('serviceAccountName')
+    role_data.update(roles)
+    return role_data
+
+def create_clusterrolebinding_yaml():
+    # namespace = str(sys.argv[3])
+    role_binding = {
+    "apiVersion" : 'rbac.authorization.k8s.io/v1',
+    "kind" : 'ClusterRoleBinding',
+    "metadata" : {
+        "name" : 'service-binding-operator',
+    },
+    "roleRef" : {
+        "apiGroup" : 'rbac.authorization.k8s.io',
+        "kind" : 'ClusterRole',
+        "name" : 'service-binding-operator',
+    },
+    "subjects":
+        [{"kind" : 'ServiceAccount', "name": 'service-binding-operator', "namespace": 'service-binding-operator'},],
+}
+    return role_binding
+
+def create_operator_deployment(csv):
+    depl = {
+    "apiVersion" : 'apps/v1',
+    "kind" : 'Deployment',
+    "metadata" : {
+        "name" : 'service-binding-operator',
+        "annotations": {"olm.targetNamespaces" : ""},
+        "namespace" : 'service-binding-operator',
+    },
+}
+    with open(csv, 'r') as f:
+        data = yaml.load(f)
+    roles = data["spec"]["install"]["spec"]["deployments"][0]
+    del roles['name']
+    depl.update(roles)
+    return depl
+
+def create_service_account():
+    service_account = {
+    "apiVersion" : 'v1',
+    "kind" : 'ServiceAccount',
+    "metadata" : {
+        "name" : 'service-binding-operator',
+        "namespace" : 'service-binding-operator'
+    },
+}
+    return service_account
+
+def create_service_binding_operator_ns():
+    namespace = {
+    "apiVersion" : 'v1',
+    "kind" : 'Namespace',
+    "metadata" : {
+        "name" : 'service-binding-operator'
+    },
+}
+    return namespace
+
+## prepares release.yaml manifest to be added with GitHub releases/tags
+release_manifest()
--- a/test/acceptance/features/bindAppToServiceAnnotations.feature
+++ b/test/acceptance/features/bindAppToServiceAnnotations.feature
@@ -135,7 +135,6 @@ Feature: Bind an application to a service using annotations
         And The application env var "BACKEND_SPEC_IMAGENAME" has value "postgres"
         And The application env var "BACKEND_SPEC_DBNAME" has value "db-demo"
 
-
     Scenario: Each value in referred slice of strings from service resource gets injected into app as separate env variable
         Given Generic test application "slos-app" is running
         And The Custom Resource Definition is present
@@ -325,6 +324,7 @@ Feature: Bind an application to a service using annotations
         And The application env var "BACKEND_WEBARROWS_SECONDARY" has value "secondary.example.com"
         And The application env var "BACKEND_WEBARROWS_404" has value "black-hole.example.com"
 
+    @olm
     Scenario: Backend Service metadata annotations update for service bindings gets propagated to the binding secret
         Given OLM Operator "backend" is running
         * The Custom Resource is present
@@ -379,6 +379,7 @@ Feature: Bind an application to a service using annotations
 
 
     @negative
+    @olm
     Scenario: Backend Service metadata annotations update not specific to service bindings does not get propagated to the binding secret
         Given OLM Operator "backend" is running
         * The Custom Resource is present
--- a/test/acceptance/features/bindAppToServiceUsingConfigMap.feature
+++ b/test/acceptance/features/bindAppToServiceUsingConfigMap.feature
@@ -7,6 +7,7 @@ Feature: Bind values from a config map referred in backing service resource
         Given Namespace [TEST_NAMESPACE] is used
         * Service Binding Operator is running
 
+    @olm
     Scenario: Inject into app a key from a config map referred within service resource
         Binding definition is declared on service CRD.
 
@@ -78,6 +79,7 @@ Feature: Bind values from a config map referred in backing service resource
         Then Service Binding "cmsa-1" is ready
         And The application env var "BACKEND_CERTIFICATE" has value "certificate value"
 
+    @olm
     Scenario: Inject into app all keys from a config map referred within service resource
         Binding definition is declared on service CRD.
 
--- a/test/acceptance/features/bindAppToServiceUsingSecret.feature
+++ b/test/acceptance/features/bindAppToServiceUsingSecret.feature
@@ -7,6 +7,7 @@ Feature: Bind values from a secret referred in backing service resource
         Given Namespace [TEST_NAMESPACE] is used
         * Service Binding Operator is running
 
+    @olm
     Scenario: Inject into app a key from a secret referred within service resource
         Binding definition is declared on service CRD.
         Given OLM Operator "backend" is running
@@ -77,6 +78,7 @@ Feature: Bind values from a secret referred in backing service resource
         Then Service Binding "ssa-1" is ready
         And The application env var "BACKEND_USERNAME" has value "AzureDiamond"
 
+    @olm
     Scenario: Inject into app all keys from a secret referred within service resource
 
         Given OLM Operator "backend" is running
@@ -149,6 +151,7 @@ Feature: Bind values from a secret referred in backing service resource
         And The application env var "BACKEND_USERNAME" has value "AzureDiamond"
         And The application env var "BACKEND_PASSWORD" has value "hunter2"
 
+    @olm
     Scenario: Inject into app a key from a secret referred within service resource Binding definition is declared via OLM descriptor.
 
         Given Generic test application "ssd-1" is running
@@ -259,6 +262,7 @@ Feature: Bind values from a secret referred in backing service resource
         And The application env var "BACKEND_USERNAME" has value "AzureDiamond"
         And The application env var "BACKEND_HOST" has value "example.com"
 
+    @olm
     Scenario: Inject into app all keys from a secret referred within service resource Binding definition is declared via OLM descriptor.
 
         Given Generic test application "ssd-2" is running
--- a/test/acceptance/features/bindWithCustomNamingStrategies.feature
+++ b/test/acceptance/features/bindWithCustomNamingStrategies.feature
@@ -1,3 +1,4 @@
+@olm
 Feature: Bind an application to a service using custom naming strategies
 
     As a user of Service Binding Operator
@@ -42,7 +43,6 @@ Feature: Bind an application to a service using custom naming strategies
             """
         Then Service Binding "binding-request-no-naming" is ready
         And The application env var "BACKEND_HOST_CROSS_NS_SERVICE" has value "cross.ns.service.stable.example.com"
-
 
     Scenario: Bind an application to a service with naming strategy none
         * OLM Operator "backend" is running
--- a/test/acceptance/features/customEnvVar.feature
+++ b/test/acceptance/features/customEnvVar.feature
@@ -8,6 +8,7 @@ Feature: Inject custom env variable into application
         Given Namespace [TEST_NAMESPACE] is used
         * Service Binding Operator is running
 
+    @olm
     Scenario: Sequence from service resource is injected into application using custom env variables without specifying annotations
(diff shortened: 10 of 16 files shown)


Makefile
```diff
@@ -495,4 +496,26 @@ validate-release: setup-venv
...
 prepare-operatorhub-pr:
-	./hack/prepare-operatorhub-pr.sh $(OPERATOR_VERSION) $(OPERATOR_BUNDLE_IMAGE_REF)
\ No newline at end of file
+	./hack/prepare-operatorhub-pr.sh $(OPERATOR_VERSION) $(OPERATOR_BUNDLE_IMAGE_REF)
+
+.PHONY: prepare-release-manifests
+## prepare file for release
+## use this target when the operator needs to be released as a single manifest with project release tags
+prepare-release-manifests: setup-venv
+	./hack/prepare-operatorhub-pr.sh $(OPERATOR_VERSION) $(OPERATOR_BUNDLE_IMAGE_REF)
```
pedjak, line 505: we have already target for this and we should declare it as dependency to `prepare-release-manifests`

Makefile
```diff
@@ -495,4 +496,26 @@ validate-release: setup-venv
...
 ## use this target when the operator needs to be released as upstream operator
 prepare-operatorhub-pr:
-	./hack/prepare-operatorhub-pr.sh $(OPERATOR_VERSION) $(OPERATOR_BUNDLE_IMAGE_REF)
\ No newline at end of file
+	./hack/prepare-operatorhub-pr.sh $(OPERATOR_VERSION) $(OPERATOR_BUNDLE_IMAGE_REF)
+
+.PHONY: prepare-release-manifests
+## prepare file for release
+## use this target when the operator needs to be released as a single manifest with project release tags
+prepare-release-manifests: setup-venv
```
pedjak, line 504: please short it to `release-manifests`.

Makefile
```diff
@@ -495,4 +496,26 @@ validate-release: setup-venv
...
+	printf '\n---\n' >> $(RELEASE_MANIFEST)/release.yaml
+	cat ./deploy/operator.yaml >> $(RELEASE_MANIFEST)/release.yaml
+	printf '\n---\n' >> $(RELEASE_MANIFEST)/release.yaml
+	sed -i -e 's,REPLACE_IMAGE,$(CONTAINER_IMAGE),g' $(RELEASE_MANIFEST)/release.yaml
+	cat ./deploy/role_binding.yaml >> $(RELEASE_MANIFEST)/release.yaml
+	printf '\n---\n' >> $(RELEASE_MANIFEST)/release.yaml
+	cat ./deploy/role.yaml >> $(RELEASE_MANIFEST)/release.yaml
+	printf '\n---\n' >> $(RELEASE_MANIFEST)/release.yaml
+	cat ./deploy/service_account.yaml >> $(RELEASE_MANIFEST)/release.yaml
+	printf '\n---\n' >> $(RELEASE_MANIFEST)/release.yaml
```
pedjak, line 521: * Please move it into a separate script, for easier maintainability. 
* Consider using Python for that task
* Extract from CSV file full resources, try to copy from `deploy` folder only things that cannot be found in CSV (e.g. service accounts)

pedjak (changes requested):

Makefile
Avni-Sharma: role bindings as well.

Makefile
```diff
@@ -496,4 +496,9 @@ deploy-from-index-image:
...
 ## Run acceptance tests with the operator installed from a given index image and channel
 test-acceptance-with-bundle: deploy-from-index-image
-	$(Q)TEST_ACCEPTANCE_START_SBO=remote $(MAKE) test-acceptance
\ No newline at end of file
+	$(Q)TEST_ACCEPTANCE_START_SBO=remote $(MAKE) test-acceptance
+
+release-manifests: get-test-namespace prepare-operatorhub-pr
+	mkdir -p ./out/releases/$(OPERATOR_VERSION)
+	touch ./out/releases/$(OPERATOR_VERSION)/release.yaml
+	python3 ./hack/release-manifest.py $(TEST_NAMESPACE)
```
pedjak, line 504: the script should be run using previously created virtualenv

hack/release-manifest.py
```diff
@@ -0,0 +1,103 @@
+import os
+import yaml
+import sys
+
+
+OPERATOR_VERSION = os.getenv('OPERATOR_VERSION')
+GIT_COMMIT_ID = os.getenv('GIT_COMMIT_ID')
+release_yaml = "./out/releases/" + OPERATOR_VERSION + "/release.yaml"
+
+def release_manifest():
```
pedjak, line 10: IMHO, the script can be simplified significantly, if we use `yaml` library more.

https://stackoverflow.com/questions/53150743/add-multiple-docs-in-yaml-file-pyyaml

and for example `glob` package:

https://pymotw.com/2/glob/

we need to perform the following:
* iterate over all `*.yaml` files in the passed manifest folder
* load `deploy/service_account.yaml` and add it to the list
* if the filename does not end with `clusterserviceversion.yaml`, read it using `yaml` library, store in the list
* for CSV file, read it using `yaml` library and extract from there (1) `spec.install.spec.deployments in standalone deployments, and (2) `spec.install.permissions.rules` into standalone cluster role
* create cluster role binding using `deploy/role_binding.yaml`

We should not use `deploy/operator.yaml` - the real deployment with real image ref is already part of CSV, we should extract it from there.

pedjak (changes requested):
This PR need to add additional GtiHub workflow/job that runs acceptances tests against minukube cluster where the operator is installed used the produced `release.yaml` file. Only with tests passing we can be sure that the generated `release.yaml` is correct.

Finally, PR summary and description need to be updated to reflect what is done.

Makefile
```diff
@@ -122,9 +122,9 @@ GOCOV_DIR ?= $(ARTIFACT_DIR)/test-coverage
 GOCOV_FILE_TEMPL ?= $(GOCOV_DIR)/REPLACE_TEST.txt
 GOCOV ?= "-covermode=atomic -coverprofile REPLACE_FILE"
 
-GIT_COMMIT_ID ?= $(shell git rev-parse --short=8 HEAD)
+export GIT_COMMIT_ID ?= $(shell git rev-parse --short=8 HEAD)
```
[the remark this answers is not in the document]
Avni-Sharma: for it to be available as an env var, which then can be used in the python script

Makefile
```diff
@@ -496,4 +496,9 @@ deploy-from-index-image:
 .PHONY: test-acceptance-with-bundle
 ## Run acceptance tests with the operator installed from a given index image and channel
 test-acceptance-with-bundle: deploy-from-index-image
-	$(Q)TEST_ACCEPTANCE_START_SBO=remote $(MAKE) test-acceptance
\ No newline at end of file
+	$(Q)TEST_ACCEPTANCE_START_SBO=remote $(MAKE) test-acceptance
+
+release-manifests: get-test-namespace prepare-operatorhub-pr
```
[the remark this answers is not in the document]
Avni-Sharma: the variable is passed as an argument to the script

Makefile
```diff
@@ -122,9 +122,9 @@ GOCOV_DIR ?= $(ARTIFACT_DIR)/test-coverage
 GOCOV_FILE_TEMPL ?= $(GOCOV_DIR)/REPLACE_TEST.txt
 GOCOV ?= "-covermode=atomic -coverprofile REPLACE_FILE"
 
-GIT_COMMIT_ID ?= $(shell git rev-parse --short=8 HEAD)
+export GIT_COMMIT_ID ?= $(shell git rev-parse --short=8 HEAD)
 
-OPERATOR_VERSION ?= 0.4.0
+export OPERATOR_VERSION ?= 0.4.0
```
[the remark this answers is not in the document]
Avni-Sharma: Removed this for latestcommit.Passing as arg to script now

Makefile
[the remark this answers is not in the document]
Avni-Sharma: Removed this for latestcommit.Passing as arg to script now

hack/release-manifest.py
Avni-Sharma: I tried the yaml.dump_all function and it gives a stream which can be hard to read for a user. It gives something like
```
      example-servicebinding\"\n    },\n    \"spec\": {\n      \"application\": {\n\
      \        \"group\": \"apps\",\n        \"name\": \"nodejs-rest-http-crud\",\n\
      \        \"resource\": \"deployments\",\n        \"version\": \"v1\"\n     \
      \ },\n      \"services\": [\n        {\n          \"group\": \"postgresql.example.dev\"\
      ,\n          \"kind\": \"Database\",\n          \"name\": \"pg-instance\",\n\
      \          \"version\": \"v1alpha1\"\n        }\n      ]\n    }\n  }\n]"
      ```

Comment (Avni-Sharma) [reply hack/release-manifest.py:14]: This would work but the user who might take a look at the release.yaml at a glance, this might be not so readable

Comment (pedjak) [reply hack/release-manifest.py:15]: > I tried the yaml.dump_all function and it gives a stream which can be hard to read for a user. It gives something like

What you posted is not yaml, it looks like json.

I have just tried to follow stackoverflow link shared above and I was able to get proper yaml files. It would be interesting to see the code snippet that generates the snippet you posted.

File: hack/prepare-operatorhub-pr.sh [hunk 9]
```diff
@@ -1,19 +1,19 @@
-#!/usr/bin/env -S bash -e
+#!/bin/bash -e
```
pedjak, line 1: why is this change necessary?

hack/prepare-operatorhub-pr.sh
```diff
@@ -1,19 +1,19 @@
...
 
 TMP_OCI_PATH=$(mktemp -d out/sbo-bundle-oci.XXX)
 
-skopeo copy docker://$2 oci:$TMP_OCI_PATH:bundle
+skopeo copy --src-tls-verify=$3 --dest-tls-verify=$3 docker://$2 oci:$TMP_OCI_PATH:bundle
 umoci unpack --image $TMP_OCI_PATH:bundle --rootless ${TMP_OCI_PATH}-unpacked
 rm -rf out/operatorhub-pr-files
 mkdir out/operatorhub-pr-files/service-binding-operator -p
-mv ${TMP_OCI_PATH}-unpacked/rootfs out/operatorhub-pr-files/service-binding-operator/$1
+sudo mv ${TMP_OCI_PATH}-unpacked/rootfs out/operatorhub-pr-files/service-binding-operator/$1
```
pedjak, line 16: no need for sudo.. we are moving files inside the same project root.

hack/prepare-operatorhub-pr.sh
```diff
@@ -1,19 +1,19 @@
...
 # - https://github.com/containers/skopeo
 
 # Usage:
-# prepare-operatorhu-pr.sh <version> <bundle-image-ref>
+# prepare-operatorhu-pr.sh <version> <bundle-image-ref> <tls-verification-requirement>
 
 TMP_OCI_PATH=$(mktemp -d out/sbo-bundle-oci.XXX)
 
-skopeo copy docker://$2 oci:$TMP_OCI_PATH:bundle
+skopeo copy --src-tls-verify=$3 --dest-tls-verify=$3 docker://$2 oci:$TMP_OCI_PATH:bundle
```
pedjak, line 12: `--dest-tls-verify` option is not needed, because the destination is the local folder.

Did you try to turn off tls verification for all cases? That should work and then we do not need to pass extra argument to the script, nor declare the var in Makefile.

Makefile
```diff
@@ -497,4 +498,10 @@ deploy-from-index-image:
 .PHONY: test-acceptance-with-bundle
 ## Run acceptance tests with the operator installed from a given index image and channel
 test-acceptance-with-bundle: deploy-from-index-image
-	$(Q)TEST_ACCEPTANCE_START_SBO=remote $(MAKE) test-acceptance
\ No newline at end of file
+	$(Q)TEST_ACCEPTANCE_START_SBO=remote $(MAKE) test-acceptance
+
+.PHONY: release-manifests
+## prepare file for Release
+release-manifests: setup-venv release-operator prepare-operatorhub-pr get-test-namespace
```
pedjak, line 505: is `get-test-namespace` really required for this target?

Makefile
```diff
@@ -497,4 +498,10 @@ deploy-from-index-image:
 .PHONY: test-acceptance-with-bundle
 ## Run acceptance tests with the operator installed from a given index image and channel
 test-acceptance-with-bundle: deploy-from-index-image
-	$(Q)TEST_ACCEPTANCE_START_SBO=remote $(MAKE) test-acceptance
\ No newline at end of file
+	$(Q)TEST_ACCEPTANCE_START_SBO=remote $(MAKE) test-acceptance
+
+.PHONY: release-manifests
+## prepare file for Release
```
pedjak, line 504: Can you provide a better help message?

Makefile
```diff
@@ -497,4 +498,10 @@ deploy-from-index-image:
...
+release-manifests: setup-venv release-operator prepare-operatorhub-pr get-test-namespace
+	$(Q)$(PYTHON_VENV_DIR)/bin/pip install pyaml
```
pedjak, line 506: this should happen as a part of `setup-env`. Add `pyaml` to `requirements.txt`

hack/release-manifest.py
```diff
@@ -0,0 +1,96 @@
...
+import yaml
+import sys
+from pathlib import Path
+
+OPERATOR_VERSION = str(sys.argv[2])
+GIT_COMMIT_ID = str(sys.argv[3])
+path = os.getcwd()
+release_path = os.path.join(path, "out", "releases", OPERATOR_VERSION)
+os.makedirs(release_path)
+release_yaml = path + "/out/releases/" + OPERATOR_VERSION + "/release.yaml"
```
pedjak, line 11: pass the full path to `release.yaml` as the script arg.

hack/release-manifest.py
pedjak: This conversation is marked as resolved, but unfortunately the script still need to evolve to address the previous comments, so that readability and maintainability can reach acceptable level:

* we should iterate over all found yaml files, not only fixed one.
* we should not copy files line be line
* ClusterRole should be constructed from CSV, no need to keep it in `deploy` folder.
* there is no need to use temp file `.data.yaml`

Avni-Sharma:
For https://github.com/redhat-developer/service-binding-operator/pull/806#discussion_r561919357 currently, ClusterRole is not there in the CSV. So it needs to be under the deploy/ as discussed earlier. Only role is in CSV.

pedjak:
> For #806 (comment) currently, ClusterRole is not there in the CSV. So it needs to be under the deploy/ as discussed earlier. Only role is in CSV.

@Avni-Sharma - in CSV one can find rules under `spec.install.permissions and they are the same either if one wants to construct `Role` or `ClusterRole` object - what objects will be created depends on the decision if the operator is cluster-scoped, or namespace-scoped. Some resources like CSVs, CRD are cluster-scoped, and we need `ClusterRole` for them. Anyhow, it can constructed out of CSV.

hack/prepare-operatorhub-pr.sh
Avni-Sharma: The CI didnt have -S and the command was failing ..even after excluding S flag it was failing on CI

pedjak: Can you paste here the error? If needed we should update CI configuration - `using /usr/bin/env` has it advantages over `/bin/bash` drectly, because on some systems `bash` can be installed on a different location.

.github/workflows/pr-checks.yaml
```diff
@@ -77,7 +77,7 @@ jobs:
       - name: Set up PATH
         run: |
           mkdir -p $GITHUB_WORKSPACE/bin/
-          echo "PATH=$PATH:$GITHUB_WORKSPACE/bin/" >> $GITHUB_ENV
+          echo "PATH=$GITHUB_WORKSPACE/bin/:$PATH" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
Avni-Sharma: The $PATH is searched from beginning to end, with the first matching executable being run. So directories at the beginning of $PATH take precedence over those that come later. https://superuser.com/questions/238987/how-does-unix-search-for-executable-files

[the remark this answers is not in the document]
pedjak: Sure, but why do we need to change the order in this PR? `$GITHUB_WORKSPACE/bin` does not contain any binaries that might be found in `$PATH`. What we are trying to fix here?

sbose78:
Are we publishing or generating here?

pedjak:
> Are we publishing or generating here?

Generating first ofcourse. @Avni-Sharma PR description should be clearer.

.github/workflows/pr-checks.yaml
```diff
@@ -131,3 +131,90 @@ jobs:
...
+      OPM_VERSION: "1.15.2"
+      MINIKUBE_WANTUPDATENOTIFICATION: false
+      MINIKUBE_WANTREPORTERRORPROMPT: false
+      K8S_VERSION: "1.19.2"
+      MINIKUBE_VERSION: "1.15.1"
+      EXTRA_BEHAVE_ARGS: "--tags=@non-olm"
+      TEST_ACCEPTANCE_CLI: "kubectl"
+      CONTAINER_RUNTIME: "docker"
+      UMOCI_VERSION: "0.4.5"
+      OPERATOR_VERSION: "0.4.0"
```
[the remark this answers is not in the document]
Avni-Sharma: Done https://github.com/redhat-developer/service-binding-operator/pull/806/commits/163308e04286c26213502395c3d54caa7b87d823

.github/workflows/pr-checks.yaml
```diff
@@ -131,3 +131,90 @@ jobs:
...
+      - name: Docker Info
+        run: |
+          eval $(minikube docker-env)
+          docker ps
+          kubectl get nodes -o yaml
+          kubectl cluster-info
+          docker info
+
+      - name: Acceptance tests with release artifacts
+        timeout-minutes: 60
```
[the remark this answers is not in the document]
Avni-Sharma: Done https://github.com/redhat-developer/service-binding-operator/pull/806/commits/163308e04286c26213502395c3d54caa7b87d823

Makefile
Avni-Sharma: For testing in the CI, it is required

hack/release-manifest.py
Avni-Sharma: PTAL https://github.com/redhat-developer/service-binding-operator/pull/806/commits/10c7cabca396b5e1f0435bb5f25f5115db807cda

.github/workflows/pr-checks.yaml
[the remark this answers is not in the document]
Avni-Sharma: reverted https://github.com/redhat-developer/service-binding-operator/pull/806/commits/10c7cabca396b5e1f0435bb5f25f5115db807cda

deploy/cluster_role.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: sbo-cluster-role
```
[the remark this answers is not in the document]
Avni-Sharma: Done

Makefile
Avni-Sharma: Done https://github.com/redhat-developer/service-binding-operator/pull/806/commits/10c7cabca396b5e1f0435bb5f25f5115db807cda

hack/release-manifest.py
Avni-Sharma: Any particular reason for that? I thought of passing min args to the script

.github/workflows/pr-checks.yaml
```diff
@@ -131,3 +131,90 @@ jobs:
...
+        run: |
+          eval $(minikube docker-env)
+          GIT_COMMIT_ID=$(git rev-parse --short=8 HEAD)
```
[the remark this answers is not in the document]
Avni-Sharma: Done https://github.com/redhat-developer/service-binding-operator/pull/806/commits/96e7530d81013ab81905fb05c1315a484532ee8b

Makefile
Avni-Sharma: This requirements.txt https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/requirements.txt ? Btw this requirements.txt is setup during test-acceptance-setup target, but I just want to  use setup-venv. I am using setup-venv target since the release-manifests target is not solely for acceptance tests

.github/workflows/pr-checks.yaml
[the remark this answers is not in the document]
Avni-Sharma: On the hindsight I think it is necessary at GIT_COMMIT_ID is not being passed by Makefile(unless we mention export). Why CI is failing is due to that, `./hack/prepare-operatorhub-pr.sh 0.4.0 192.168.49.2:5000/sbo:bundle-0.4.0- false` 
after the version commit isn't there

pedjak (changes requested):

hack/release-manifest.py
```diff
@@ -0,0 +1,100 @@
...
+        yaml.dump(role_data, outfile, default_flow_style=False)
+    with open(csv, 'r') as f:
+        data = yaml.load(f)
+    roles = data["spec"]["install"]["spec"]["clusterPermissions"]
+    with open('.data.yml', 'a') as outfile:
+        roles[0].pop('serviceAccountName')
+        yaml.dump(roles[0],outfile)
+    copy_lines_to_file("./.data.yml")
+    if os.path.exists("./.data.yml"):
+        os.remove("./.data.yml")
```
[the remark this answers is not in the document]
Avni-Sharma: https://github.com/redhat-developer/service-binding-operator/pull/806/commits/2be9948b83d3de4f7ae5894abe9c5ee7bde47121

hack/prepare-operatorhub-pr.sh
Avni-Sharma: I tried running it with the suggested usage nd the script was not able to execute, It threw the following error 
`Error: The action has timed out`

pedjak: that is not coming from the script, it is coming from GitHub actions - what was the run, can you share the link?

Avni-Sharma: https://github.com/redhat-developer/service-binding-operator/runs/1908927102?check_suite_focus=true

Avni-Sharma: Why I felt this commit triggered the timeout err is because only on this specific commit change I got the timeout issue, never before and not in the later commits as well.

pedjak: The script was stuck somewhere, hence the timeout. It would be good to add maybe some debugging infos to that script, so that you can see where it stucks. Please make shebang line `#!/usr/bin/env -S bash -e` and try again.

Avni-Sharma: Had experienced this earlier, getting the err again `/usr/bin/env: invalid option -- 'S'`

.github/actions/setup-cli/action.yaml
```diff
@@ -0,0 +1,35 @@
+# action.yml
+name: 'Setup Tools'
```
pedjak, line 2: there even greater common denominator - checking out code, setting up go, etc... we should extract all this in this common action.

.github/workflows/pr-checks.yaml
```diff
@@ -90,44 +85,57 @@ jobs:
...
+          mv -v umoci $GITHUB_WORKSPACE/bin/
 
-      - name: Acceptance tests
+      - name: Acceptance tests with release artifacts
         timeout-minutes: 60
         run: |
           eval $(minikube docker-env)
-          make OPERATOR_REPO_REF=$(minikube ip):5000/sbo SKIP_REGISTRY_LOGIN=true release-operator -o registry-login test-acceptance-with-bundle
\ No newline at end of file
+          GIT_COMMIT_ID=$(git rev-parse --short=8 HEAD)
```
pedjak, line 135: you do not need to set this here - it is already set in `Makefile`.

hack/release-manifest.py
```diff
@@ -0,0 +1,93 @@
...
+    create_clusterrolebinding_yaml()
+    create_service_account()
+    with open(release_yaml, "a") as stream:
+        yaml.dump_all(
+            res,
+            stream,
+            default_flow_style=False
+        )  
+
+def copy_lines_to_file(filename):
```
pedjak, line 28: this function does not what its name says. please rename.

hack/release-manifest.py
```diff
@@ -0,0 +1,93 @@
+import os
+import yaml
+import sys
+from pathlib import Path
+
+MANIFEST_DIR = str(sys.argv[1])
+path = os.getcwd()
+release_yaml = path + "/out/release.yaml"
+Path(release_yaml).touch()
+res = list()
```
pedjak, line 10: define this is var inside `release_manifest()`, avoid that functions are accessing global vars.

test/acceptance/features/bindAppToServiceAnnotations.feature
```diff
@@ -65,6 +65,7 @@ Feature: Bind an application to a service using annotations
         And Secret "binding-request-backend-a" contains "BACKEND_READY" key with value "true"
         And Secret "binding-request-backend-a" contains "BACKEND_HOST" key with value "example.common"
 
+
```
pedjak, line 68: whitespace changes, please remove.

pedjak (changes requested):

Avni-Sharma:
@pedjak Should I revert the changes with the @non-olm tag? Something like we previously had?

.github/workflows/pr-checks.yaml
```diff
@@ -90,44 +85,57 @@ jobs:
...
+          make OPERATOR_REPO_REF=$(minikube ip):5000/sbo SKIP_REGISTRY_LOGIN=true release-operator -o registry-login
+          make TLS_VERIFY=false OPERATOR_REPO_REF=$(minikube ip):5000/sbo OPERATOR_BUNDLE_IMAGE_REF=$(minikube ip):5000/sbo:bundle-${OPERATOR_VERSION}-$GIT_COMMIT_ID SKIP_REGISTRY_LOGIN=true release-manifests -o registry-login
+          namesp=$(kubectl config view --minify --output 'jsonpath={..namespace}')
+          sed -i "s/REPLACE_NAMESPACE/$namesp/g" out/release.yaml
```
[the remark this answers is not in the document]
Avni-Sharma: The actual release.yaml would have REPLACE_NAMESPACE. For testing purposes we are just renaming in the CI and not in the actual one.

.github/workflows/pr-checks.yaml
Avni-Sharma: I tried it without the GIT_COMMIT_ID mentioned here and the CI is not picking up from the Makefile

.github/workflows/pr-checks.yaml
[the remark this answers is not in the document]
pedjak: We should not have it. Any reasons for that?

.github/workflows/pr-checks.yaml
pedjak: Can you be more specific? Can you share the job run?

.github/workflows/pr-checks.yaml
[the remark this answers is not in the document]
Avni-Sharma: Also we cant omit to have REPLACE_NAMESPACE because if we do we encounter this error
```
Error from server (Invalid): error when creating "out/release.yaml": ClusterRoleBinding.rbac.authorization.k8s.io "service-binding-operator" is invalid: subjects[0].namespace
```

if we have a subject we need a ns for it to be applied successfully

.github/workflows/pr-checks.yaml
Avni-Sharma: ./hack/prepare-operatorhub-pr.sh 0.4.0 192.168.49.2:5000/sbo:bundle-0.4.0- 
time="2021-02-16T16:30:13Z" level=fatal msg="Error initializing source docker://192.168.49.2:5000/sbo:bundle-0.4.0-: Error reading manifest bundle-0.4.0- in 192.168.49.2:5000/sbo: manifest unknown: manifest unknown"

Avni-Sharma: https://github.com/redhat-developer/service-binding-operator/pull/806/checks?check_run_id=1912133716

pedjak: You should have the line like this:

https://github.com/redhat-developer/service-binding-operator/pull/806/files#diff-7a6a4578fe04c4fc86b28e2dc9672e48d1a9f721315e540b90bac0ef12263053R95

in order to things to work.

hack/release-manifest.py
Avni-Sharma: We need it as global as it is being used in another function copy_lines_to_file , (needs renaming)

pedjak: Good practice in any language is to avoid having functions with side effects: instead of adding something `rest` withing the called function (and not passing `res` as param), return to the caller the object that will be added to `res` by the caller.

Avni-Sharma:
In composite actions we can reference scripts or inline commands,(reference https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsrun) not quite sure how to reference a predefined action from a composite action. (Had mentioned this in a comment probably it got lost due to squashing)

pedjak:
> In composite actions we can reference scripts or inline commands,(reference https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsrun) not quite sure how to reference a predefined action from a composite action. (Had mentioned this in a comment probably it got lost due to squashing)

Ah, thanks for investigating. You are right, this is still not supported, here is the issue that tracks this: https://github.com/actions/runner/issues/646

.github/workflows/pr-checks.yaml
```diff
@@ -90,44 +85,55 @@ jobs:
...
+          make OPERATOR_REPO_REF=$(minikube ip):5000/sbo SKIP_REGISTRY_LOGIN=true release-operator -o registry-login test-acceptance-with-bundle
 
-          curl -Lo opm https://github.com/operator-framework/operator-registry/releases/download/v${OPM_VERSION}/linux-amd64-opm
-          chmod +x opm
-          mv -v opm $GITHUB_WORKSPACE/bin/
 
-          curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v${K8S_VERSION}/bin/linux/amd64/kubectl
-          chmod +x kubectl
-          mv -v kubectl $GITHUB_WORKSPACE/bin/
+  acceptance_non_olm:
```
pedjak, line 98: I would call it `acceptance_without_olm`

.github/workflows/pr-checks.yaml
```diff
@@ -90,44 +85,55 @@ jobs:
...
+    name: Non OLM Acceptance Tests with Kubernetes
```
pedjak, line 99: I would call it `Acceptance tests running on Kubernetes without installed OLM`

.github/workflows/pr-checks.yaml
```diff
@@ -90,44 +85,55 @@ jobs:
...
+    runs-on: ubuntu-20.04
 
-          curl -Lo minikube https://storage.googleapis.com/minikube/releases/v${MINIKUBE_VERSION}/minikube-linux-amd64
-          chmod +x minikube
-          mv -v minikube $GITHUB_WORKSPACE/bin/
+    env:
+      EXTRA_BEHAVE_ARGS: "--tags=~@knative --tags=~@openshift --tags=non-olm --tags=~@olm"
```
pedjak, line 103: we do not need to introduce a new `non-olm` tag, all other tags are sufficient: we want to run scenarios that do not have neither of `@knative`, `@openshift` and `@olm` tags.

Avni-Sharma: Tried it without the nonolm tag and somehow the features and scenarios (all) were being skipped

Avni-Sharma: Tried it with ` --tags=~@knative --tags=~@openshift --tags=~@olm --tags=annotations` as well :/

Avni-Sharma:
Passing scenarios(5) feature(1) for `non-olm` tag https://github.com/redhat-developer/service-binding-operator/pull/806/checks?check_run_id=1918461787#step:7:3731

Skipped tests with `~olm` tag https://github.com/redhat-developer/service-binding-operator/runs/1910014910?check_suite_focus=true#step:7:3712 

How to incorporate ~olm tags passing the CI. cc @pmacik @pratikjagrut 

We are trying to avoid the entry of a new `non-olm` tag and instead using `~olm` for such scenarios.

.github/workflows/pr-checks.yaml
```diff
@@ -90,44 +85,55 @@ jobs:
...
-          docker ps
-          kubectl get nodes -o yaml
-          kubectl cluster-info
-          docker info
+          curl -Lo umoci https://github.com/opencontainers/umoci/releases/download/v${UMOCI_VERSION}/umoci.amd64
+          chmod +x umoci
+          mv -v umoci $GITHUB_WORKSPACE/bin/
 
-      - name: Acceptance tests
+      - name: Acceptance tests with release artifacts
```
pedjak, line 131: How about "Acceptance tests against vanilla k8s cluster"?

Makefile
```diff
@@ -301,3 +301,9 @@ test-acceptance-generate-report:
 ## Serves acceptance tests report at http://localhost:8088
 test-acceptance-serve-report:
 	$(Q)CONTAINER_RUNTIME=$(CONTAINER_RUNTIME) $(HACK_DIR)/allure-report.sh serve
+
+.PHONY: release-manifests
+## prepare a manifest file for Release which installs the sbo operator on vanilla k8s
```
pedjak, line 306: how about: `prepare a manifest file for releasing operator on vanilla k8s cluster`?

.github/workflows/pr-checks.yaml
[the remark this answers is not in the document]
pedjak: The more I think about this, the more I would like to see `release.yaml` to be fully ready for user consumption, i.e. it should be possible to perform the following:

```
kubectl apply -f release.yaml
```

As the result, the operator will be installed in a predefined documented namespace. That should be default and would lower the barrier a lot. If somebody wants to drift from that, he already knows how to do it. The target namespace could be passed as arg to ` hack/release-manifest.py` script.

Therefore, let's create `release.yaml` file that creates all resources in `service-binding-operator` namespace, add the proper `Namespace` object to the file as well as the first entry.

pedjak (changes requested):
Please create a good PR title/description what this PR all addressed and how.

.github/workflows/pr-checks.yaml
Avni-Sharma: I think the previous job is also running on vanilla k8s but with OLM installed as an addon. So this job is not exclusively just vanilla k8s. Probably Acceptance tests against vanilla k8s without OLM installed?

pedjak: Sounds good.. feel free to rename the other job as well to describe better what it does.

Avni-Sharma:
@pedjak PTAL

test/acceptance/features/examples.feature
```diff
@@ -1,4 +1,5 @@
 @examples
+@openshift
```
pedjak, line 2: why this change?

Avni-Sharma: Both scenarios are openshift dependent and last scenario was missing the tag openshift. so instead of adding on two scenarios i added it to the feature file itself

.github/workflows/pr-checks.yaml
```diff
@@ -90,44 +85,53 @@ jobs:
...
 
-      - name: Acceptance tests
+      - name: Acceptance tests against vanilla k8s without OLM
         timeout-minutes: 60
         run: |
           eval $(minikube docker-env)
-          make OPERATOR_REPO_REF=$(minikube ip):5000/sbo SKIP_REGISTRY_LOGIN=true release-operator -o registry-login test-acceptance-with-bundle
\ No newline at end of file
+          make OPERATOR_REPO_REF=$(minikube ip):5000/sbo SKIP_REGISTRY_LOGIN=true release-operator -o registry-login release-manifests
+          kubectl apply -f out/release.yaml
```
pedjak, line 136: all resources that are not cluster-scoped should be deployed within `service-binding-operator` namespace. Hence, `release.yaml` needs to contain `Namespace` object and all other namespace-scoped objectes should have `namespace` seto `service-binding-operator`. By inspecting the python script, I can see this is not the case.

hack/release-manifest.py
```diff
@@ -0,0 +1,107 @@
...
+        with open(os.path.join(MANIFEST_DIR, filename), 'r') as f:
+            data = yaml.load(f)
+        if data["kind"] == "ClusterServiceVersion":
+            cluster_role = create_cluster_role_yaml(os.path.join(MANIFEST_DIR, filename))
+            res.append(cluster_role)
+            depl = create_operator_deployment(os.path.join(MANIFEST_DIR, filename))
+            res.append(depl)
+        else:
+            with open(os.path.join(MANIFEST_DIR, filename), "r") as stream:
+                res.extend(list(yaml.safe_load_all(stream)))
```
pedjak, line 23: resources loaded here should be enriched with namespace.

test/acceptance/features/bindAppToServiceUsingConfigMap.feature
```diff
@@ -1,3 +1,4 @@
+@olm
```
pedjak, line 1: I would rather not annotation the whole feature with `olm` - it would mean that it fully depend on existence of OLM, but that is not the case. Hence, please annotation individual scnearios where appropriate. Another drawback of annotating the whole feature  is that any new scenarios will be implicitly tagged as well, although perhaps this is not really required.

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -1,3 +1,4 @@
+@olm
```
pedjak, line 1: see above

test/acceptance/features/bindWithCustomNamingStrategies.feature
```diff
@@ -1,3 +1,4 @@
+@olm
```
pedjak, line 1: see above

test/acceptance/features/customEnvVar.feature
```diff
@@ -1,3 +1,4 @@
+@olm
```
pedjak, line 1: see above

test/acceptance/features/deleteBackendServiceAndRecreate.feature
```diff
@@ -1,3 +1,4 @@
+@olm
```
pedjak, line 1: see above

test/acceptance/features/examples.feature
pedjak: see above

test/acceptance/features/bindAppToServiceUsingConfigMap.feature
Avni-Sharma: Took inspiration from https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToMultipleServices.feature#L1 https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToService.feature#L1 https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindKnativeService.feature#L2

Avni-Sharma: Can we document this in Good practices also? Probably it would be helpful for future PRs?

pedjak: sure, feel free to raise PR for that.

Avni-Sharma: For those having 
```
    Background:
        * PostgreSQL DB operator is installed
```

Backgroud as Postgres DB operator installed, that means that is the background for the feature file, in that case also should we tag scenarios instead of the feature file?

pedjak: that background applies to all scenarios in feature file, so then we can just tag `@olm` the whole feature... not the ideal, but we should fix this later.

cc @pratikjagrut

pedjak (approved):

PR #807, opened 2020-12-16
Fix Docker image pull limits 

pedjak:
Due to recently introduced image pull limits, we see occasionally job failures at Travis.

Hence, openshift/origin-release:golang-1.15 image got mirrored to quay
and that image is now our base build image.


pmacik:
The acceptance fails because `etcd` operator vanished from `community-operators` catalog source... This PR https://github.com/redhat-developer/service-binding-operator/pull/812 changes the catalog source to `operatorhub.io` where `etcd` operator is anyway.

Issue #520, opened 2020-06-24
Update status with a proper message when two similar services have been referred in 2 SBR for the same app


Avni-Sharma:
An informational condition could be created to inform the user there's a duplicated entry in the services field, but it shouldn't prevent the well functioning of the operator because of that fact.

Refer https://github.com/redhat-developer/service-binding-operator/pull/514#issuecomment-645809793

Avni-Sharma:
An update in the status would only  happen effectively if the similar services are in the same SBR, however, with 2 different sbrs and a similar service in both the sbrs, won't be able to update the status of their respective sbrs, with the knowledge of having distinct services.

pedjak:
@Avni-Sharma please provide acceptance tests demonstrating the expected behaviour.

Avni-Sharma:
Currently, we avoid redundancy of binding 2 services of a particular db instance with an app. ie if db-1 instance is mentioned in the list of the services more than once in a single sbr, then we take the count as 1 and avoid duplicates or more copies. But if the same instance of the service is in a separate sbr, then the redundancy is not avoided. That os covered as an acceptance test in #825

Avni-Sharma:
We need to discuss whether we should propagate a message as a status, mentioning the redundant service? or we want to completely avoid the injection in case of a second/third/n th sbr with a redundant service?

Avni-Sharma:
As illustrated in PR #825 . The user experience for the application author is not broken, i.e the author can continue to work with multiple instances of SBR but with a similar spec. 
- As a follow up we can look at how this turns out to be on the OpenShift Console.
- Add a status/message suggesting the redundant bindings(multiple sbrs with the same spec)

Right now the issue does not seem to be critical as the user experience is not broken. As discussed in the Cabal I would be removing the 0.5 release tag for this.

Also, we would like to explore any use cases wherein there is a need by the app author to have multiple sbrs with the same spec.

Avni-Sharma:
On OpenShift DevConsole when we apply the second SBR with the same spec (so we can test the behavior of the console with duplicate bindings) and it only rendered a single arrow (for the first binding). When deleted the binding (via right-clicking on the arrow) the arrow remained to represent the second binding. So It Doesnt seem to be a critical issue on the DevConsole UI as well. So this can be closed IMO

PR #750, opened 2020-11-04
Immutable Binding Secrets

DhritiShikhar:
### Motivation

This PR:

1. Makes binding secrets immutable.
- Each time binding data is updated, a new binding secret is created with the name format `<name of ServiceBinding> + "-" + <hash of the binding data>`
- Example: `sb-1-72ddc0c540ab3a290e138726940591debf14c581` where `sb-1` is the name of ServiceBinding and `72ddc0c540ab3a290e138726940591debf14c581` is the hash of the binding data.

2. Fixes #630 

### Changes

- [x] Removes `changeTriggerEnv` which was used earlier to force redeployment of the application.
- [x] Removes SBR annotations from the application
- [x] Adds code to sort the secret keys and create data hash
- [x] Adds an acceptance test for the scenario https://github.com/redhat-developer/service-binding-operator/issues/630
- [x] Updates existing unit and acceptance test

### Steps to test this PR

1. Use one of the examples given [here](https://github.com/redhat-developer/service-binding-operator/tree/master/examples) to create a service, an application and a ServiceBinding. Ensure binding is successful by checking the status of the ServiceBinding.

2. Execute `oc get deployment <application_name> -o json | jq '.spec.template.spec.containers` and notice binding secret name is present

3. Update binding data in the service.

4. Again, Execute `oc get deployment <application_name> -o json | jq '.spec.template.spec.containers` and notice new secret name is present.

5. Create another ServiceBinding refering to the same application.

6. Check the application pods `oc get pods` and notice constant creation and termination of pods has stopped.

7. Also, check Service Binding Operator logs to see that indefinite reconciliation has stopped on applying two ServiceBindings referring to the same application.

### Acceptance Test

[1] To run the acceptance tests, execute `make test-acceptance`

PR #825, opened 2020-12-30
Bind a single app and a particular db instance with multiple sbrs

Avni-Sharma:
Motivation

#520

This PR is not a fix to #520 , but it rather adds an acceptance test which is currently not included in the master and we would have to fix thereafter.

### DIFF
--- a/test/acceptance/features/bindAppToMultipleSbrs.feature
+++ b/test/acceptance/features/bindAppToMultipleSbrs.feature
@@ -1,0 +1,54 @@
+# @olm
+Feature: Bind a single application to multiple SBRs
+
+    As a user of Service Binding operator
+    I want to bind a single application to multiple SBRs
+
+    Background:
+        Given Namespace [TEST_NAMESPACE] is used
+        * Service Binding Operator is running
+        * PostgreSQL DB operator is installed
+
+    Scenario: Bind two backend services by creating 2 SBRs to a single application
+        Given Imported Nodejs application "nodejs-app" is running
+        * DB "db-demo-1" is running
+        * Service Binding is applied
+            """
+            apiVersion: operators.coreos.com/v1alpha1
+            kind: ServiceBinding
+            metadata:
+                name: binding-request-1
+            spec:
+                application:
+                    name: nodejs-app
+                    group: apps
+                    version: v1
+                    resource: deployments
+                services:
+                -   group: postgresql.baiju.dev
+                    version: v1alpha1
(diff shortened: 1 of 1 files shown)


DhritiShikhar:
@Avni-Sharma My PR #750 already adds an acceptance test for this scenario -> https://github.com/redhat-developer/service-binding-operator/pull/750/files#diff-2f593af0f1f9a93dcfe23785e784dd8e05c285c3e8f4d362fc7d23547c3c0c6c

Please review it.

Avni-Sharma:
@DhritiShikhar the acceptance test here is referring to a particular database instance, db-demo in both the sbr being connected to a particular app and then sbr2 does the same. The intention is to omit redundancy so this acceptance test currently shows that the redundancy is there, i.e if we have the same service in 2 sbrs then there is a redundant connection. In your acceptance test, the database instances are different.

test/acceptance/features/bindAppToMultipleSbrs.feature
```diff
@@ -0,0 +1,54 @@
+@olm
+Feature: Bind a single application to multiple SBRs
+
+    As a user of Service Binding operator
+    I want to bind a single application to multiple SBRs
```
pedjak, line 5: This is not a new feature, it is either binding application to single or multiple services.

test/acceptance/features/bindAppToMultipleSbrs.feature
```diff
@@ -0,0 +1,54 @@
...
+
+    Background:
+        Given Namespace [TEST_NAMESPACE] is used
+        * Service Binding Operator is running
+        * PostgreSQL DB operator is installed
+
+    Scenario: Bind a single db instance by creating 2 SBRs to a single application
```
pedjak, line 12: So, we speak about binding application to single service? Thus, this scenario should be moved to `bindAppToService.feature`

Also, let's describe better how this scenario is different from others we have already. From the applied steps, it looks like that we apply to the cluster two identical service bindings. Such action should be idempotent - the outcome should be equal to the scenario when just single service binding is applied.

Also, please consider replacing DB service with mocked one, and using generic test application, to make the test lighter. That would allow us to proper assert expectations, i.e. how application context looks like.

pedjak (changes requested):

Avni-Sharma:
https://github.com/redhat-developer/service-binding-operator/issues/520 is not relevant now, hence closing the PR

PR #827, opened 2021-01-04
Update contributing guide with code of coduct

baijum:
Ref. https://issues.redhat.com/browse/APPSVC-787

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,11 +1,99 @@
 # How to Contribute
 
-#### Table Of Contents
-* [Getting Started](#getting-started)
-* [How Can I Contribute?](#how-can-i-contribute)
-  * [Reporting Bugs](#reporting-bugs)
-  * [Suggesting Enhancements](#suggesting-enhancements)
-  * [Pull Requests](#pull-requests)
+:+1::tada: First off, thanks for taking the time to contribute!
+
+You can look at the issues [with help wanted label][help-wanted] for items that
+you can work on.
+
+If you need help, please feel free to reach out to our [slack channel](https://app.slack.com/client/T09NY5SBT/C019LQYGC5C)!
+
+When contributing to this repository, please first discuss the change you wish
+to make via issue, email, or any other method with the owners of this repository
+before making a change.  Small pull requests are easy to review and merge.  So,
+please send small pull requests.
+
+Please note we have a [code of conduct][conduct], please follow it in all your
+interactions with the project.
+
+Contributions to this project should conform to the [Developer Certificate of
+Origin][dco].  See the [next section](#sign-your-work) for more details.
+
+## Sign Your Work
+
+Contributions to this project should conform to the [Developer Certificate of
+Origin][dco].  You need to sign-off your git commits before sending the pull
+requests.  The sign-off is a single line of text at the end of the commit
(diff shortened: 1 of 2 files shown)


sbose78:
Thank you Baiju.

CONTRIBUTING.md
```diff
@@ -7,6 +7,101 @@
...
+Note: If your git config information is set properly then viewing the `git log`
+information for your commit will look something like this:
+
+```
+Author: Joe Smith <joe.smith@example.com>
+Date:   Thu Feb 2 11:41:15 2018 -0800
+
+    Update README
+
+    Signed-off-by: Joe Smith <joe.smith@example.com>
```
pedjak, line 98: we should add CI job checking this, preferably in this PR.

CONTRIBUTING.md
```diff
@@ -7,6 +7,101 @@
   * [Suggesting Enhancements](#suggesting-enhancements)
   * [Pull Requests](#pull-requests)
```
pedjak, line 8: some links are dead, please verify - should we update TOC, or remove it?

CONTRIBUTING.md
baijum: Let's see, if we can enable this: https://probot.github.io/apps/dco/

pedjak: it looks it is configured now?

baijum: Yes. See this new PR #834

CONTRIBUTING.md
baijum: I would suggest removing TOC. It's not really required for a doc like this.

CONTRIBUTING.md
```diff
@@ -7,6 +7,101 @@
...
 
+:+1::tada: First off, thanks for taking the time to contribute!
+
+You can look at the issues [with help wanted label][help-wanted] for items that
+you can work on.
+
+If you need help, please feel free to reach out to our [slack channel](https://app.slack.com/client/T09NY5SBT/C019LQYGC5C)!
```
DhritiShikhar, line 15: This link is not correct. It should point to kubernetes workspace

Screenshot from 2021-01-07 17-27-03

CONTRIBUTING.md
```diff
@@ -7,6 +7,101 @@
...
+    it.
+
+(d) I understand and agree that this project and the contribution
+    are public and that a record of the contribution (including all
+    personal information I submit with it, including my sign-off) is
+    maintained indefinitely and may be redistributed consistent with
+    this project or the open source license(s) involved.
+```
+
+Then you just add a line to every git commit message:
```
DhritiShikhar, line 80: This should also be mentioned:

`git commit --signoff`

CONTRIBUTING.md
baijum: If the person already logged into Kubernetes slack. That link will open the channel.

CONTRIBUTING.md
baijum: That will require more explanation about the Git configuration file also. We can assume the developer knows Git.

CONTRIBUTING.md
baijum: removed TOC

Issue #389, opened 2020-03-20
Adding support to use a resource without spec or status as a backing service


navidsh:
Currently, it is only possible to bind to a backing service that has `spec`/`status`. However, resources such as Secrets and ConfigMaps do not have `spec` and `status`.

This is to track the support following the discussion [here](https://github.com/redhat-developer/service-binding-operator/pull/367#issuecomment-596547734).

sbose78:
Thanks!

Avni-Sharma:
Currently, a workaround is there

That is to use mappings, initially known as customEnvVar.

Steps to reproduce:

Create a configMap:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: example
data:
  extra: "hello"
```
Create a sbr yaml as follows:
```
apiVersion: operators.coreos.com/v1alpha1
kind: ServiceBinding
metadata:
  name: binding-request
spec:
  application:
    group: apps
    version: v1
    resource: deployments
    name: nodejs-rest-http-crud
  services:
  - group: ""
    version: v1
    kind: ConfigMap
    name: example
    id: s
  customEnvVar:
  - name: HELLO
    value: '{{ .s.data.extra }}'
```

now check the intermediate secret that is created. the secret that the sbr created will have data as HELLO: hello from the configmap

Will be adding an acceptance test for this

pedjak:
Fixed by #984

PR #522, opened 2020-06-26
Revert annotation handling mechanism to not use the original config map or secret resource name location

isutton:
### Motivation

After #407 an undocumented behavior regarding annotations meant to extract
config map and secret values has been unwillingly changed and needs to be
reverted.

The annotations handling mechanism after #407 extracts the whole contents of
the config map or secret to the same path where the config map or secret name
has been stored (resulting in, for example, .status.secretName.password if
.data.password is present in the associated resource) to workaround
situations where two services might have a password field within different
paths and might get shadowed, whereas before #407 it used to extract the
contents to the root of the resulting object (for example, .password),
imposing no issue whatsoever because there were only one service being
declared in the binding.

Closes #485 

### Changes

This change reverts the behavior where the source path isn't expected in the
resulting environment variable name; for example, the excerpt below shows
annotations present in a resource CRD, as a JSON object:

```json
{
  "servicebindingoperator.redhat.io/status.connectionArguments.databaseName": "binding:env:attribute",
  "servicebindingoperator.redhat.io/status.connectionArguments.host": "binding:env:attribute",
  "servicebindingoperator.redhat.io/status.connectionArguments.port": "binding:env:attribute",
  "servicebindingoperator.redhat.io/status.connectionCredentials-password": "binding:env:object:secret",
  "servicebindingoperator.redhat.io/status.connectionCredentials-username": "binding:env:object:secret",
  "servicebindingoperator.redhat.io/status.externalConnectionArguments-databaseName": "binding:env:object:configmap",
  "servicebindingoperator.redhat.io/status.externalConnectionArguments-host": "binding:env:object:configmap",
  "servicebindingoperator.redhat.io/status.externalConnectionArguments-port": "binding:env:object:configmap"
}
```

Before the changes introduced in this PR, the following environment variables
are generated and encoded in the service binding intermediate secret, where
the issue can be seen where the string `STATUS_EXTERNALCONNECTIONARGUMENTS`
and `STATUS_CONNECTIONCREDENTIALS` can be read:

```
NAME            TYPE    KEY                                                                     VALUE
service-binding Opaque  DATABASE_CONFIGMAP_STATUS_EXTERNALCONNECTIONARGUMENTS_DATABASENAME      "db-name"
service-binding Opaque  DATABASE_CONFIGMAP_STATUS_EXTERNALCONNECTIONARGUMENTS_HOST              "database.prod.example.com"
[...]

PR #407, opened 2020-03-31
Streamline data retrieval based on annotations

isutton:
### Motivation

- Have a test baseline covering existing service binding annotation semantics; and
- Provide configuration values from all declared services to the custom environment variable templates.

### Changes

Some existing components have been simplified and several others have been removed in order to reduce the number of indirections in the system.

The `annotations` package has been created to group all annotation related symbols. `annotations.Handler` is an interface specifying the annotation handler contract, which is responsible for processing a single annotation key and value pairs as for example `.../status.dbConnectionIP` and `binding:env:attribute`; an instance can be obtained through `annotations.BuildHandler()`. 

There are currently two annotation handlers: `annotations.AttributeHandler` and `annotations.ResourceHandler`. The handlers return a `Result` containing the data produced by the annotation in `Data`, which is stored in a deep structure with the root path found in the `Path` field. Additionally the binding type (whether volume mounts or environment variables should be used) is present in the `Type` field.

The `AttributeHandler` is used to collect a value from the object where the annotation is declared; for example, in the example `.../status.dbConnectionIP` and `binding:env:attribute` annotation pair the data found in the simple JSONPath `status.dbConnectionIP` will be extracted from the object. On the other hand, the `ResourceHandler` is used to collect information present in another external resource, for example a config map or a secret; in the `.../status.dbCredentials-user` and `binding:env:object:configmap` annotation pairs example, the `data.user` field from the config map resource which name is present in `status.dbCredentials` will be extracted.

The `envvars` package contains tooling to construct `map[string][]byte` data-structures from an arbitrarily nested `map[string]interface{}` value.

The `nested` package contains utilities to extract the arbitrarily deep data-structures in the annotation handlers.

Since the whole data extraction process has been modified to provide the testing baseline, this change also includes support for custom environment variable templates such as `{{ index "v1alpha1" "postgresql.baiju.dev" "Database" "db-testing" "metadata" "name" }}` and `{{ .v1alpha1.postgresql_baiju_dev.Database.db_testing.metadata.name }}`.

Test coverage should be improved, and internals should keep to be simplified in subsequent work units, such as:

- Review whether `ServiceBinder` and `Binder` can be further simplified.
- Review whether the data being propagated through function parameters in the reconciliation can be re-grouped now that some indirections have been removed.

### Testing

Currently the only tests being amended are unit tests. Once integration is more mature, e2e tests will be used to guarantee the behavior hasn't been changed unwillingly.

For further more details refer the CONTRIBUTING.md

PR #831, opened 2021-01-05
Add acceptance test with configMap as a service

Avni-Sharma:
Motivation : #389


test/acceptance/features/bindAppToService.feature
```diff
@@ -756,3 +756,36 @@ Feature: Bind an application to a service
             """
         Then Service Binding "binding-request-cross-ns-service" is ready
         And The application env var "BACKEND_HOST_CROSS_NS_SERVICE" has value "cross.ns.service.stable.example.com"
+
+    @negative
+    Scenario: Service binding with a missing app and configMap as a service
+        Given ConfigMap is applied
```
pedjak, line 762: we can use existing step called `The ConfigMap is present`

test/acceptance/features/bindAppToService.feature
```diff
@@ -756,3 +756,36 @@ Feature: Bind an application to a service
             """
         Then Service Binding "binding-request-cross-ns-service" is ready
         And The application env var "BACKEND_HOST_CROSS_NS_SERVICE" has value "cross.ns.service.stable.example.com"
+
+    @negative
+    Scenario: Service binding with a missing app and configMap as a service
```
pedjak, line 761: We should focus on scenarios with app involved, in our case that could be generic generic test app. This particular case does not have much sense without the app, because a user could ask - what is the point of applying service binding against a config map, to get the secret containing the same content. Please rephrase the scenario stating what are the expectations, e.g. `Inject all configmap keys into application`

pedjak (changes requested):

test/acceptance/features/bindAppToService.feature
```diff
@@ -756,3 +756,36 @@ Feature: Bind an application to a service
...
+        Given ConfigMap is applied
+            """
+            apiVersion: v1
+            kind: ConfigMap
+            metadata:
+                name: example
+                annotations:
+                    service.binding: path={.data},elementType=map
```
pedjak, line 769: `elementType=map` is not a part of [the spec](https://github.com/k8s-service-bindings/spec#binding-secret-generation-strategies). If we support it, we should document it.

test/acceptance/features/bindAppToService.feature
```diff
@@ -756,3 +756,36 @@ Feature: Bind an application to a service
             """
         Then Service Binding "binding-request-cross-ns-service" is ready
         And The application env var "BACKEND_HOST_CROSS_NS_SERVICE" has value "cross.ns.service.stable.example.com"
+
+    @negative
```
pedjak, line 760: why is this considered as negative scenario?

test/acceptance/features/bindAppToService.feature
pedjak, line 761: Would it be possible to binding just a single key from config map? We should have a test for that. https://github.com/redhat-developer/service-binding-operator/issues/389 mentions Secrets as well.. we should check if they could be used as service as well.

test/acceptance/features/bindAppToService.feature
```diff
@@ -756,3 +756,36 @@ Feature: Bind an application to a service
...
+                -   group: ""
+                    version: v1
+                    kind: ConfigMap
+                    name: example
+            """
+        Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "binding-request-configmap" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "binding-request-configmap" should be changed to "False"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").reason" of Service Binding "binding-request-configmap" should be changed to "EmptyApplication"
+        And jq ".status.conditions[] | select(.type=="Ready").status" of Service Binding "binding-request-configmap" should be changed to "True"
+        And Secret "binding-request-configmap" contains "CONFIGMAP_DATA_WORD" key with value "hello"
```
pedjak, line 790: is there are a way to inject bindings using the original keys from configmap/secret. For example, to inject env var `WORD` with value `hello`?

test/acceptance/features/bindAppToService.feature
isutton: @pedjak I believe any annotation syntax that might have been implemented should be, in addition of being documented, ironed out by the spec proposers.

test/acceptance/features/bindAppToService.feature
isutton: This has been implemented in https://github.com/redhat-developer/service-binding-operator/pull/407 and https://github.com/redhat-developer/service-binding-operator/pull/522.

pedjak: @isutton what should be the syntax to achieve this?

isutton: I was referring to including the kind in the name for disambiguation purposes. It has appeared in both PRs and was pointing out so you can check what has been implemented.

Avni-Sharma:
Hi @pedjak  I have added documentation for annotation case but in descriptors we have status x desc and spec x desc parts so I am not sure if the elementType=`map` when path ={.data} is supported under x descriptors. Thoughts? cc @isutton

pedjak (approved):

PR #834, opened 2021-01-07
Fix ServiceBinding removal when service not exist

qibobo:
I will add the UT and acceptance tests later

### DIFF
--- a/controllers/reconciler.go
+++ b/controllers/reconciler.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"errors"
+
 	"k8s.io/apimachinery/pkg/api/meta"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


DhritiShikhar:
@qibobo You would have to signoff your commit with `git commit --amend --signoff` for DCO PR check to pass.

qibobo:
@DhritiShikhar @isutton 
Could you help review this PR?

pkg/controller/servicebinding/reconciler.go
```diff
@@ -113,6 +113,34 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+			return requeueError(err)
+		}
+		return done()
+	}
+	if sbr.GetDeletionTimestamp() != nil {
+		logger := logger.WithName("unbind")
+		logger.Info("Executing unbinding steps...")
+		// here just build a servicebinder with necessary fields to do unbind
+		ensureDefaults(sbr.Spec.Application)
+		sb := &serviceBinder{
```
pedjak, line 130: could we use `buildServiceBinder()` here? How is this instance different from the existing instance, i.e. why `unbind` cannot be performed?

qibobo: @pedjak 
In current reconcile process, to do the unbind, it needs to `buildServiceContexts` and `buildBinding`, and `buildServiceBinder`. When the backing service has been deleted and it tries to delete the SBR, `buildServiceContexts` fail and requeue, then fail and requeue..... The SBR can not be deleted.

The reason here I did not use the `buildServiceBinder` is that it needs the result `buildServiceContexts` and `buildBinding`, which will fall as I mentioned above. And for unbind, I think there is no need to `buildServiceContexts` and `buildBinding` because the results of them are not necessary to do the unbind.

pedjak: @qibobo thanks for explanation. If I understand correctly, the issue here is that `buildServiceContexts()`is not aware of the fact that service binding is about to be deleted. However, it would be good that the change inside `Reconcile()` function is as small as possible. Hence, how about this: instead of creating `servicebinder` manually, how about that we use `buildServiceBinder` as before, but we creating empty service context if we recognize that deletion is in progress? Something like:

```golang
serviceCtxs, err := buildServiceContexts(...)
if err != nil {
   if sbr.GetDeletionTimestamp() != nil {
      serviceCtxs = serviceContextList{}
   } else {
	//handle service not found error
	if k8serrors.IsNotFound(err) {
        ...
        }
```

I believe this is the only change we need to perform in the reconcile function.

pkg/controller/servicebinding/reconciler.go
```diff
@@ -165,29 +165,34 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-				},
-				conditionsv1.Condition{
-					Type:   BindingReady,
-					Status: corev1.ConditionFalse,
-				},
-			)
-			if err != nil {
-				logger.Error(err, "Failed to update SBR conditions", "sbr", sbr)
+		if sbr.GetDeletionTimestamp() != nil {
+			serviceCtxs = serviceContextList{}
```
isutton, line 169: Wouldn't make sense to execute `buildServiceContexts` only if the `GetDeletionTimestamp() != nil` check fails?

isutton (changes requested):

pkg/controller/servicebinding/reconciler_test.go
```diff
@@ -648,3 +649,98 @@ func TestBindTwoSbrsWithSingleApplication(t *testing.T) {
 		require.Equal(t, sbrName2, dep.Spec.Template.Spec.Containers[0].EnvFrom[1].SecretRef.LocalObjectReference.Name)
 	})
 }
+
+// TestApplicationByName tests discovery of application by name
+func TestUnBind(t *testing.T) {
```
pedjak, line 654: what is scenario we test here, can you reflect with the function name? If the scenario is the same as the new introduced acceptance test, we can then avoid introducing this test  fully.

qibobo:
@isutton @pedjak Fixed.

pedjak (approved):
lgtm, please squash and restructure a bit the commit message, something like:

`Fix ServiceBinding removal when service not exist`

qibobo:
@pedjak 
done

qibobo:
@pedjak 
rebased and need lgtm again.

Issue #838, opened 2021-01-13
Service binding is creating a secret with wrong key when sourceKey is mentioned in annotation


pratikjagrut:
I was trying to bind a backend service to an application using annotation in CRD
```
service.binding/certificate: path={.status.data.dbConfiguration},objectType=ConfigMap,sourceKey=certificate
```

This is CR:
```
apiVersion: stable.example.com/v1
kind: Backend
metadata:
    name: cmsa-1-service
spec:
    image: docker.io/postgres
    imageName: postgres
    dbName: db-demo
status:
    data:
        dbConfiguration: cmsa-1-configmap    # ConfigMap
```
This is CM:
```
apiVersion: v1
kind: ConfigMap
metadata:
    name: cmsa-1-configmap
data:
    certificate: "certificate value"
```
This is sb cr:
```
apiVersion: operators.coreos.com/v1alpha1
kind: ServiceBinding
metadata:
    name: cmsa-1
spec:
    services:
      - group: stable.example.com
        version: v1
        kind: Backend
        name: cmsa-1-service
    application:
        name: cmsa-1
        group: apps
        version: v1
        resource: deployments
```

Expected intermediate secret data:

```
BACKEND_CERTIFICATE: Y2VydGlmaWNhdGUgdmFsdWU=
```

Actual intermediate secret data:
```
BACKEND: Y2VydGlmaWNhdGUgdmFsdWU=
```

Acceptance test to reproduce: https://github.com/redhat-developer/service-binding-operator/pull/842

pratikjagrut:
Same problem if we use this annotation
```
service.binding: path={.status.data.dbConfiguration},objectType=ConfigMap,sourceKey=certificate
```

pmacik:
@pratikjagrut Could you please create a new acceptance tests that would serve as a reproducer (it fails now and passes when the issues is fixed) and also to be used to verify that the issue/bug was fixed?

pmacik:
And of course, since it is supposed to fail until fixed, it should be pushed as `@disabled` to be enabled along with the PR to fix the bug

isutton:
`sourceKey` is used in other scenarios, what should be used there is `sourceValue` instead. See:

https://github.com/redhat-developer/service-binding-operator/blob/52c0f396240175bb3abe7e6809007081551649c9/pkg/controller/servicebinding/binding/spec_test.go#L172-L182

wtrocki:
[...]

PR #842, opened 2021-01-15
Add acceptance test to verify key in a secret

pratikjagrut:
### Motivation

This PR is to verify the scenario in #838.


test/acceptance/features/steps/steps.py
```diff
@@ -510,3 +510,14 @@ def create_deployment(context, app_name, image_ref):
...
         assert app.install() is True, "Failed to create deployment."
+
+
+@then(u'Secret "{secret_name}" does not have key "{secret_key}"')
+def check_key_in_secret(context, secret_name, secret_key):
+    time.sleep(10)
+    openshift = Openshift()
+    json_path = f'{{.data.{secret_key}}}'
+    if not openshift.get_resource_info_by_jsonpath("secrets", secret_name, context.namespace.name, json_path):
+        assert True
```
pmacik, line 521: You don't need to `assert True` that is like a `nop` or effectively empty line... The assertion here should be only for the `False` case, also the failure message should explain more precisely, what went wrong - something like:
```python
assert openshift.get_resource_info_by_jsonpath("secrets", secret_name, context.namespace.name, json_path) is None, f"Secret '{secret_name}' must not have key '{secret_key}'"
```

pmacik:
@isutton could you please confirm, that this (the acceptance scenario) is expected behavior? Thanks

test/acceptance/features/steps/steps.py
pratikjagrut: Updated, PTAL.

pmacik: Thx, but actually there is already a step for secret not having a secret... `@given(u'Secret "{secret_name}" does not contain "{key}"')` - it is only for `@given` but it can be changed to `@step` to cover the `Then` as well...

test/acceptance/features/bindAppToServiceAnnotations.feature
```diff
@@ -427,3 +427,73 @@ Feature: Bind an application to a service using annotations
                 ready: true
             """
         Then Secret "binding-request-backend-ann" is empty
+
+
+    @disabled
```
pedjak, line 432: can we enable to test or it is not working?

test/acceptance/features/bindAppToServiceAnnotations.feature
```diff
@@ -427,3 +427,73 @@ Feature: Bind an application to a service using annotations
...
+    # Verify https://github.com/redhat-developer/service-binding-operator/issues/838
+    Scenario: Verify data of secret created by service binding
```
pedjak, line 434: This scenario is referring a key from a config map and for that we have dedicated feature file: `bindAppToServiceUsingConfigMap.feature`. Even more, we have scenario there that should actually cover it:

https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToServiceUsingConfigMap.feature#L10

but it looks that it uses a wrong annotation. Hence that scenario should be updated, instead of adding a new one. Similar scenario exists in 

https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToServiceUsingSecret.feature#L10

That should be also updated, because the spec defines that `sourceKey` should be used:

https://github.com/k8s-service-bindings/spec#annotation-examples

```
“service.binding/certificate”:
  "path={.status.data.dbConfiguration},objectType=Secret,sourceKey=certificate"
```

pedjak (changes requested):

test/acceptance/features/bindAppToServiceAnnotations.feature
pratikjagrut: Okay, I'll enable it to see failure then will post the CI output and will disable the test again.

test/acceptance/features/bindAppToServiceAnnotations.feature
pratikjagrut: > This scenario is referring a key from a config map and for that we have dedicated feature file: `bindAppToServiceUsingConfigMap.feature`. Even more, we have scenario there that should actually cover it:
> 

Actually, this behaviour was due to annotation which is same if I even use secrets so I put this test in the annotation file.

> but it looks that it uses a wrong annotation. Hence that scenario should be updated, instead of adding a new one. Similar scenario exists in

I will update other scenarios as well but this test will serve as a regression test for  #838. So I think this test should be in test suites.

pedjak: > I will update other scenarios as well but this test will serve as a regression test for #838. So I think this test should be in the test suite.

When you update these scenarios, there is no need to add additional scenario, because they are covering the same issue. or are they different and how?

pratikjagrut:
Expected failure scenarios:
```
Failing scenarios:
  test/acceptance/features/bindAppToServiceAnnotations.feature:434  Verify data of secret created by service binding
  test/acceptance/features/bindAppToServiceUsingConfigMap.feature:10  Inject into app a key from a config map referred within service resource
  test/acceptance/features/bindAppToServiceUsingSecret.feature:10  Inject into app a key from a secret referred within service resource
```
One of the failure scenarios log.
```
  Scenario: Verify data of secret created by service binding                                                                         # test/acceptance/features/bindAppToServiceAnnotations.feature:434
    Given Namespace [TEST_NAMESPACE] is used                                                                                         # test/acceptance/features/steps/steps.py:43
    * Service Binding Operator is running                                                                                            # test/acceptance/features/steps/steps.py:83
    Given OLM Operator "backend" is running                                                                                          # test/acceptance/features/steps/steps.py:330
    And The Custom Resource Definition is present                                                                                    # test/acceptance/features/steps/steps.py:369
      """
      apiVersion: apiextensions.k8s.io/v1beta1
      metadata:
          name: backends.stable.example.com
          annotations:
              service.binding/certificate: path={.status.data.dbConfiguration},objectType=ConfigMap,sourceKey=certificate
      spec:
          group: stable.example.com
          versions:
            - name: v1
              served: true
              storage: true
          scope: Namespaced
          names:
              plural: backends
              singular: backend
              kind: Backend
              shortNames:
                - bk
      """
    And The ConfigMap is present                                                                                                     # test/acceptance/features/steps/steps.py:369
      """
      apiVersion: v1
      kind: ConfigMap
      metadata:
          name: cmsa-3-configmap
      data:
          certificate: "certificate value"
      """
    And The Custom Resource is present                                                                                               # test/acceptance/features/steps/steps.py:369
      """
      apiVersion: stable.example.com/v1
      kind: Backend
      metadata:
          name: cmsa-3-service
      spec:
          image: docker.io/postgres
          imageName: postgres
          dbName: db-demo
      status:
          data:
              dbConfiguration: cmsa-3-configmap
      """
    When Service Binding is applied                                                                                                  # test/acceptance/features/steps/steps.py:180
      """
      kind: ServiceBinding
      metadata:
          name: cmsa-3
      spec:
          services:
            - group: stable.example.com
              version: v1
              kind: Backend
              name: cmsa-3-service
      """
    Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "cmsa-3" should be changed to "True" # test/acceptance/features/steps/steps.py:245
    And Secret "cmsa-3" does not have key "BACKEND"                                                                                  # test/acceptance/features/steps/steps.py:515
      Assertion Failed: Secret 'cmsa-3' must not have key 'BACKEND'
      Captured stdout:
      TEST_NAMESPACE = test-namespace-2e3bcf857c95
      ,---------,-
      | COMMAND : oc get ns test-namespace-2e3bcf857c95
      '---------'-
      Namespace test-namespace-2e3bcf857c95 is created!!!
      ,---------,-
      | COMMAND : oc get servicebinding
      '---------'-
      Service binding operator is running!!!
      ,---------,-
      | COMMAND : oc apply -n test-namespace-2e3bcf857c95 -f /go/src/github.com/redhat-developer/service-binding-operator/test/acceptance/resources/backend.operator.manifest.yaml
      '---------'-
      ,---------,-
      | COMMAND : oc apply -n test-namespace-2e3bcf857c95 -f -
      '---------'-
      ,---------,-
      | COMMAND : oc apply -n test-namespace-2e3bcf857c95 -f -
      '---------'-
      ,---------,-
      | COMMAND : oc apply -n test-namespace-2e3bcf857c95 -f -
      '---------'-
      ,---------,-
      | COMMAND : oc apply -n test-namespace-2e3bcf857c95 -f -
      '---------'-
      ,---------,-
      | COMMAND : oc get servicebinding cmsa-3 -n test-namespace-2e3bcf857c95 -o json | jq  '.status.conditions[] | select(.type=="CollectionReady").status'
      '---------'-
      ERROR MESSGE: b'jq: error (at <stdin>:47): Cannot iterate over null (null)\n'
      ERROR CODE: 5
      ,---------,-
      | COMMAND : oc get servicebinding cmsa-3 -n test-namespace-2e3bcf857c95 -o json | jq  '.status.conditions[] | select(.type=="CollectionReady").status'
      '---------'-
      ERROR MESSGE: b'jq: error (at <stdin>:47): Cannot iterate over null (null)\n'
      ERROR CODE: 5
      ,---------,-
      | COMMAND : oc get servicebinding cmsa-3 -n test-namespace-2e3bcf857c95 -o json | jq  '.status.conditions[] | select(.type=="CollectionReady").status'
      '---------'-
      ,---------,-
      | COMMAND : oc get secrets cmsa-3 -n test-namespace-2e3bcf857c95 -o "jsonpath={.data.BACKEND}"
      '---------'-

    And Secret "cmsa-3" contains "BACKEND_CERTIFICATE" key with value "certificate value"                                            # None
```

test/acceptance/features/bindAppToServiceAnnotations.feature
pratikjagrut: In this scenario, only one step is more which is testing regression, https://github.com/redhat-developer/service-binding-operator/pull/842/files#diff-4af525dbccc1a8950ce632ed0a0372567bd9159532d9a3a04349a1ee5ffe8170R495.

pratikjagrut: But I'll update the existing scenario with this step.

pedjak: You do not need to update the existing scenario at all. The step

```
Secret "cmsa-1" does not have key "BACKEND"
```

is not relevant, since the only thing what counts is what the application sees. The content of the secret is an implementatino detail, and not something what user should be aware of.

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -7,6 +7,7 @@ Feature: Bind values from a secret referred in backing service resource
         Given Namespace [TEST_NAMESPACE] is used
         * Service Binding Operator is running
 
+    @disabled
```
pedjak, line 10: why do we want to disable this scenario now?

test/acceptance/features/bindAppToServiceUsingConfigMap.feature
```diff
@@ -19,7 +20,7 @@ Feature: Bind values from a config map referred in backing service resource
             metadata:
                 name: backends.stable.example.com
                 annotations:
-                    service.binding/username: path={.status.data.dbConfiguration},objectType=ConfigMap,valueKey=username
+                    service.binding/username: path={.status.data.dbConfiguration},objectType=ConfigMap,sourceKey=username
```
pedjak, line 23: we should expect `BACKEND_USERNAME` env var in the application context, not `BACKEND_CERTIFICATE`, right?

pedjak (changes requested):

test/acceptance/features/bindAppToServiceUsingSecret.feature
pratikjagrut: Because #838 is not fixed yet so this scenario will fail.

test/acceptance/features/bindAppToServiceUsingConfigMap.feature
pratikjagrut: Sorry it's `BACKEND_CERTIFICATE` I need to rename the username to certificate.

test/acceptance/features/bindAppToServiceUsingSecret.feature
akashshinde: @pratikjagrut are you sure this scenario will fail ? since it did not @disabled tag, This tests runs for each CI test and we haven't seen any failure.

test/acceptance/features/bindAppToServiceUsingConfigMap.feature
```diff
@@ -76,6 +77,7 @@ Feature: Bind values from a config map referred in backing service resource
                     resource: deployments
             """
         Then Service Binding "cmsa-1" is ready
+        And Secret "cmsa-1" does not have key "BACKEND"
```
akashshinde, line 80: @pratikjagrut why is this required ?

test/acceptance/features/bindAppToServiceUsingSecret.feature
pratikjagrut: It's not failing currently in CI but the test is modified little bit now to do regression check so until that bug is this regression will fail.

test/acceptance/features/bindAppToServiceUsingConfigMap.feature
pratikjagrut: This is the step where we're confirming that the operator is not creating data in secret with just KIND as a key which is what happens in the bug #838

test/acceptance/features/bindAppToServiceUsingSecret.feature
akashshinde: got it

test/acceptance/features/bindAppToServiceUsingConfigMap.feature
pedjak: And this is really not needed to check - only what counts is what application sees. These are acceptance tests and users should be not aware of such implementations details. Hence, please remove, and if needed, check that application does not sees that variable.

pedjak:
embedded as the part of https://github.com/redhat-developer/service-binding-operator/pull/875

PR #850, opened 2021-01-27
Fix regression: previously create test namespace is used across test runs

pedjak:
This fixes regression introduced in PR #848 - we should reuse created test namespace
across multiple runs of acceptance tests

### Testing

Two consecutive runs of `make test-acceptance` should create just single
test namespace.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,7 @@ TEST_ACCEPTANCE_START_SBO ?= local
 TEST_ACCEPTANCE_OUTPUT_DIR ?= $(OUTPUT_DIR)/acceptance-tests
 TEST_ACCEPTANCE_REPORT_DIR ?= $(OUTPUT_DIR)/acceptance-tests-report
 TEST_ACCEPTANCE_ARTIFACTS ?= /tmp/artifacts
+TEST_NAMESPACE = $(shell $(HACK_DIR)/get-test-namespace $(OUTPUT_DIR))
 
 TEST_ACCEPTANCE_TAGS ?=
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -286,8 +280,8 @@ deploy-from-index-image:
 
 .PHONY: test-acceptance-with-bundle
 ## Run acceptance tests with the operator installed from a given index image and channel
-test-acceptance-with-bundle: deploy-from-index-image
-	$(Q)TEST_ACCEPTANCE_START_SBO=remote $(MAKE) test-acceptance
+test-acceptance-with-bundle: TEST_ACCEPTANCE_START_SBO=remote
```
pmacik, line 283: @pedjak This sets only the local variable `TEST_ACCEPTANCE_START_SBO` that is only valid for the given target `test-acceptance-with-bundle` but it is not propagated to the dependent `test-acceptance-setup` one.

Thus, down-the-line the `test-acceptance-setup` target then uses the global variable `TEST_ACCEPTANCE_START_SBO` which is by default set to the value of `local`.

As consequence first the SBO is deployed from the given index image and then started as a local instance.

See the GHA logs between the following lines where the `test-acceptance-setup` is executed with the `local` option:
* https://github.com/redhat-developer/service-binding-operator/pull/850/checks?check_run_id=1778481606#step:10:350
and
* https://github.com/redhat-developer/service-binding-operator/pull/850/checks?check_run_id=1778481606#step:10:412

Using a sub-make was to ensure that the variable is set properly to `remote` for the `test-acceptance` target as a global variable so it is set so for the whole execution of that target (including dependencies).

pedjak: Good catch, thanks! I hope it is fixed now.

PR #865, opened 2021-02-05
Update API group to binding.operators.coreos.com

### LINKED ISSUE #857: API group name need update
TL;DR: ServiceBinding API group needs to be changed - OLM does not allow operators to define webhooks for API group `operators.coreos.com` and that is by OLM design.

As a part of the work on https://github.com/redhat-developer/service-binding-operator/pull/784 we need to define a mutating web hook, so that we can gather information about requester. Webhooks are also the k8s way to go in cases when we need to validate our ServiceBindings (we have already a few issues related to that)

Unfortunately, I have stumbled upon a blocker, and the issue can be solved only by changing ServiceBinding API group. Currently, it is set to 'operators.coreos.com' and that group is particularly protected by OLM, even OpenShift documentation mentions it:

* https://docs.openshift.com/container-platform/4.6/operators/operator_sdk/osdk-generating-csvs.html#olm-webhook-considerations_osdk-generating-csvs - look for "Admission webhook rules constraints"

* the OLM code ref: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/install/webhook.go#L29

As a result, ServiceBinding CSV gets marked as "Failed" and the operator cannot be installed.

It is a good thing that we still did not reach GA, and our api is still marked as v1alpha1. Group change can certainly be disruptive for some, but that is not the hard part.

pedjak:
The update is necessary and dictated by the fact that OLM does not allow
deploying webhook for `operator.coreos.com` API group name:

* https://docs.openshift.com/container-platform/4.6/operators/operator_sdk/osdk-generating-csvs.html#olm-webhook-considerations_osdk-generating-csvs
* https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/install/webhook.go#L29

Fixes #857

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -1,10 +1,10 @@
-domain: coreos.com
+domain: operators.coreos.com
 layout: go.kubebuilder.io/v3
 projectName: service-binding-operator
 repo: github.com/redhat-developer/service-binding-operator
 resources:
 - crdVersion: v1
-  group: operators
+  group: binding
   kind: ServiceBinding
   version: v1alpha1
 version: 3-alpha
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ connect to a backing service (for example, a database):
 #### Binding a Java Application with a Database
 
 ``` yaml
-apiVersion: operators.coreos.com/v1alpha1
+apiVersion: binding.operators.coreos.com/v1alpha1
 kind: ServiceBinding
 metadata:
   name: binding-request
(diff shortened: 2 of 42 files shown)


PROJECT
```diff
@@ -1,10 +1,10 @@
-domain: coreos.com
+domain: operators.coreos.com
 layout: go.kubebuilder.io/v3
 projectName: service-binding-operator
 repo: github.com/redhat-developer/service-binding-operator
 resources:
 - crdVersion: v1
-  group: operators
+  group: binding
```
DhritiShikhar, line 7: Previously the group name was plural `operators`, shouldn't the new group name be plural as well? 
`bindings` instead of `binding`.

pedjak: The previous group name was set so that it matches the group used by OLM. Best practice is to keep group name as singular, because they are very much equivalent to package names:

* https://cluster-api.sigs.k8s.io/developer/providers/implementers-guide/naming.html
* https://softwareengineering.stackexchange.com/questions/75919/should-package-names-be-singular-or-plural

PR #873, opened 2021-02-15
Service and application referenceable both using GVK and GVR

### LINKED ISSUE #710: Making service API to use GVR instead of GVK
Currently, we have GVK being referred in for the service, and GVR for application. Not quite sure on the impetus of this decision, but we should move towards a more consistent usage across our Spec which will also be less confusing (https://github.com/redhat-developer/service-binding-operator/issues/709). 

Motivation for GVK to GVR
- Since a restMapper is propagated all way from the controller manager only to transform from K to R.
-  Helps to think that services and applications are Resources in a K8s cluster.
-  Currently, we use gvr, _ := meta.UnsafeGuessKindToResource(gvk) and according to the documentations 
```
// UnsafeGuessKindToResource converts Kind to a resource name.
// Broken. This method only "sort of" works when used outside of this package.  It assumes that Kinds and Resources match
// and they aren't guaranteed to do so.
```

pedjak:
GroupVersionKind and GroupVersionResource can be used interchangeable now
for specifying application and/or services.

* `Service` and `Application` type do not embed anymore LocalObjectReference (as recommended by the doc)
Instead, `Ref` type got introduced and embedded into `Service`, `Application` - `BundledApplication` got replaced by `Ref`
* RESTMapper usage replaced by usage of the introduced `K8STypeLookup` interface.

Fixes #710


api/v1alpha1/servicebinding_types.go
```diff
@@ -112,34 +113,56 @@ type ServiceBindingStatus struct {
...
+	// Version of the referent.
+	Version string `json:"version"`
 
+	// Kind of the referent.
+	// +optional
+	Kind string `json:"kind,omitempty"`
+
+	// Resource of the referent.
+	// +optional
+	Resource string `json:"resource,omitempty"`
```
sbose78, line 138: Do you want to gradually deprecate `Resource` ?

.github/workflows/pr-checks.yaml
```diff
@@ -127,7 +127,7 @@ jobs:
           docker info
 
       - name: Acceptance tests
-        timeout-minutes: 60
+        timeout-minutes: 120
```
Avni-Sharma, line 130: Service and application referenceable both using GVK and GVR introduces changes which would require a change in timeout?

api/v1alpha1/servicebinding_types.go
Avni-Sharma: https://github.com/redhat-developer/service-binding-operator/issues/710 hints at shifting from gvk to gvr. So I think we should consider deprecating Kind, Thoughts?

test/acceptance/features/bindAppToServiceAnnotations.feature
```diff
@@ -427,3 +427,74 @@ Feature: Bind an application to a service using annotations
                 ready: true
             """
         Then Secret "binding-request-backend-ann" is empty
+
+
+    Scenario: Binding refer service using group version resource
```
Avni-Sharma, line 432: @olm tag should be added if we are using OLM Operator backend. I guess we can use https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToServiceAnnotations.feature#L13 by applying crd of the service and then the instance without having to rely on olm backend service

api/v1alpha1/servicebinding_types.go
pedjak: None of the above - I would rather allow developers to use both of them, whatever suites them better. The spec is GVK focused, so we cannot deprecate the usage of GVK. Deprecating Resource would introduce an unnecessary API break.

sbose78: Makes sense.
The downside of that is we have to leave both as _optional_

test/acceptance/features/bindAppToServiceAnnotations.feature
Avni-Sharma: oops, sorry @olm

Avni-Sharma: I mean the tag `@olm` .. :see_no_evil:

pedjak: tag were added.

PR #880, opened 2021-02-22
Remove dependency of acceptance tests to other operators

pratikjagrut:
### Motivation

Currently, there is a number of scenarios that depends on other operators (such as the PostgreSQL, etcd or knative operators). That makes the acceptance tests heavy, dependent on those operators and more difficult to be executed without a cluster-admin role.

Also, the features are organized in a way that in some cases a scenario depends on the PostgreSQL operator, even if it does not actually use it - the dependency is in the {{Background: }} section of a feature file where the above-mentioned scenarios are placed.

### Changes

Segregating the external dependent operator scenarios and annotating them with @examples tag.


Avni-Sharma:
For the Remove dependency of acceptance tests to other operators how are we actually checking whether the time elapsed would be less in this case than what actually exists in the master? Are we checking it in the CI? How are we confirming that removing all dependencies will indeed result in a speedy test pass/fail scenario? As mentioned in https://issues.redhat.com/browse/APPSVC-812?focusedCommentId=15736777&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15736777 the expectation is also to see at the end of the work less time spent in developers's inner loop tests , in addition to ofcourse the already existing acceptance criteria.

test/acceptance/features/customEnvVar.feature
```diff
@@ -7,11 +7,11 @@ Feature: Inject custom env variable into application
     Background:
         Given Namespace [TEST_NAMESPACE] is used
         * Service Binding Operator is running
+        * OLM Operator "backend" is running
```
Avni-Sharma, line 10: If we have OLM operator as backend in the Background then like previous feature files we can technically have olm as a tag to the entire feature file itself.

test/acceptance/features/deleteBackendServiceAndRecreate.feature
```diff
@@ -6,11 +6,11 @@ Feature: Reconcile when BackingService CR got deleted and recreated
     Background:
         Given Namespace [TEST_NAMESPACE] is used
         * Service Binding Operator is running
+        * OLM Operator "backend" is running
```
Avni-Sharma, line 9: Refer https://github.com/redhat-developer/service-binding-operator/pull/880#discussion_r584524746

test/acceptance/features/bindAppToService.feature
```diff
@@ -7,183 +7,256 @@ Feature: Bind an application to a service
...
+                data:
+                    dbCredentials: backend-secret
+            """
         When Service Binding is applied
             """
             apiVersion: binding.operators.coreos.com/v1alpha1
             kind: ServiceBinding
             metadata:
-                name: binding-request-a-d-s
+                name: service-binding-a-b-s
```
pedjak, line 62: is this change really needed in this PR?

test/acceptance/features/bindAppToService.feature
```diff
@@ -7,183 +7,256 @@ Feature: Bind an application to a service
...
+        * The Secret is present
+            """
+            apiVersion: v1
+            kind: Secret
+            metadata:
+                name: backend-secret
+            stringData:
+                username: AzureDiamond
+                password: hunter2
+            """
```
pedjak, line 42: does each scenario in this feature require existence of this secret?

test/acceptance/features/bindAppToService.feature
```diff
@@ -7,183 +7,256 @@ Feature: Bind an application to a service
...
+
+    Scenario:  Bind an application to backend service in the following order: Application, Service Binding and backend service
+        Given Generic test application "gen-app-a-s-b" is running
+        And Service Binding is applied
             """
             apiVersion: binding.operators.coreos.com/v1alpha1
             kind: ServiceBinding
             metadata:
-                name: binding-request-a-s-d
+                name: service-binding-a-s-b
```
pedjak, line 86: same as above

test/acceptance/features/postgresqlOperatorExamples.feature
```diff
@@ -0,0 +1,331 @@
+@examples
+Feature: Bind and application to postgresql database
```
pedjak, line 2: type: `and`

test/acceptance/features/bindAppToMultipleServices.feature
```diff
@@ -7,9 +7,10 @@ Feature: Bind a single application to multiple services
     Background:
         Given Namespace [TEST_NAMESPACE] is used
         * Service Binding Operator is running
+        * OLM Operator "backend" is running
```
pedjak, line 10: is this fake operator really required for the scenarios in this feature file to pass? My personal feeling is no - what most of the scenarios need is just existence of appropriate CRD so that service instances can be created. We should really identify scenarios where operators ClusterServiceVersion is needed and only then use this step. For all other, we should define another step, e.g. `Operator X is running` where we just deploy needed CRDs.

test/acceptance/features/bindAppToService.feature
```diff
@@ -7,183 +7,256 @@ Feature: Bind an application to a service
     Background:
         Given Namespace [TEST_NAMESPACE] is used
         * Service Binding Operator is running
-        * PostgreSQL DB operator is installed
+        * OLM Operator "backend" is running
```
pedjak, line 10: is this step needed at given that in the next step we apply right CRD?

pedjak (changes requested):
a number of `@olm` tags can be removed now.

test/acceptance/features/bindAppToService.feature
pratikjagrut: `a-d-s` stands for `app-databse-servicebinding` but now I replaced databse with `backend-service` so I changed it to `a-b-s`.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Same explanation as above.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Not every but most of the scenarios need secrets so instead of executing every time I added it here.

test/acceptance/features/bindAppToService.feature
pmacik: In the examples have permutations of App, DB and ServiceBinding - therefore `a-d-s`. However, here we have App, Service and the Binding, so what about using `s` instead of `d` and `b` instead of `s` so we would have permutations of `a-s-b`? 

I like the idea of naming things properly, consistently and uniquely to avoid collisions.

test/acceptance/features/bindAppToMultipleServices.feature
pmacik: I agree with @pedjak, we should depend on CSV only if that is essential for the scenario itself

test/acceptance/features/bindAppToService.feature
pedjak: "Bacground" should contain steps that are needed by all scenarios.

test/acceptance/features/bindAppToMultipleServices.feature
pratikjagrut: Okay, I removed `Operator X is running` step from various scenarios PTAL.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Removed it.

test/acceptance/features/bindAppToService.feature
pratikjagrut: Done.

test/acceptance/features/bindAppToService.feature
```diff
@@ -344,7 +473,7 @@ Feature: Bind an application to a service
             """
             true
             """
-
+    @olm
```
pedjak, line 476: does this scenario depends on the existence of a CSV? If not, we should not introduce this tag

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -10,8 +10,7 @@ Feature: Bind values from a secret referred in backing service resource
     @olm
     Scenario: Inject into app a key from a secret referred within service resource
         Binding definition is declared on service CRD.
-        Given OLM Operator "backend" is running
```
pedjak, line 13: if removed, then we should remove `@olm` tag as well?

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -80,9 +79,7 @@ Feature: Bind values from a secret referred in backing service resource
 
     @olm
     Scenario: Inject into app all keys from a secret referred within service resource
-
-        Given OLM Operator "backend" is running
```
pedjak, line 84: same question as above.

test/acceptance/features/bindWithCustomNamingStrategies.feature
```diff
@@ -7,11 +7,29 @@ Feature: Bind an application to a service using custom naming strategies
...
+                    served: true
+                    storage: true
+                scope: Namespaced
+                names:
+                    plural: backends
+                    singular: backend
+                    kind: Backend
+                    shortNames:
+                      - bk
+            """
```
pedjak, line 29: Let's follow DRY and avoid copying - wrap this into a step like `CustomResourceDefinition backends.stable.example.com is available`

test/acceptance/features/postgresqlOperatorExamples.feature
```diff
@@ -0,0 +1,331 @@
...
+                    kind: Database
+                    name: db-demo-2
+            """
+        Then Service Binding "binding-request-1" is ready
+        And Service Binding "binding-request-2" is ready
+        And "nodejs-app" deployment must contain SBR name "binding-request-1"
+        And "nodejs-app" deployment must contain SBR name "binding-request-2"
+
+    # https://github.com/redhat-developer/service-binding-operator/tree/master/examples/pod_spec_path
+    Scenario: Specify container's path in Service Binding
```
pedjak, line 253: do we have now the same scenario but without database available in some other feature file?

test/acceptance/features/postgresqlOperatorExamples.feature
```diff
@@ -0,0 +1,331 @@
...
+                  - group: postgresql.baiju.dev
+                    version: v1alpha1
+                    kind: Database
+                    name: db-demo-csp
+                    id: zzz
+            """
+        Then Service Binding "binding-request-csp" is ready
+        And Secret "binding-request-csp" has been injected in to CR "demo-appconfig-csp" of kind "AppConfig" at path "{.spec.spec.containers[0].envFrom[0].secretRef.name}"
+
+    Scenario: Specify secret's path in Service Binding
```
pedjak, line 297: same as above.

pedjak (changes requested):

test/acceptance/features/bindAppToService.feature
pratikjagrut: I guess it depends on CSV, CSV file define some descriptors. It is using `OLM Operator "backend-new-spec" is running` step.

test/acceptance/features/bindAppToServiceUsingSecret.feature
pratikjagrut: Yes.

test/acceptance/features/bindAppToServiceUsingSecret.feature
pratikjagrut: Yes.

test/acceptance/features/postgresqlOperatorExamples.feature
pratikjagrut: Yes we've at https://github.com/redhat-developer/service-binding-operator/pull/880/files#diff-ca848b998fbd78a0989748822c09065858004a5e79b457ae9b57ad5ed9dcd25cR117

test/acceptance/features/postgresqlOperatorExamples.feature
pratikjagrut: Yes we've at https://github.com/redhat-developer/service-binding-operator/pull/880/files#diff-ca848b998fbd78a0989748822c09065858004a5e79b457ae9b57ad5ed9dcd25cR117

test/acceptance/features/bindWithCustomNamingStrategies.feature
pratikjagrut: PTAL.

test/acceptance/features/bindAppToService.feature
pedjak: thanks for checking.

test/acceptance/features/steps/steps.py
```diff
@@ -557,3 +557,30 @@ def update_binding_secret(context):
     output = openshift.apply(yaml.dump(secret_yaml), context.namespace.name)
     result = re.search(rf'.*{context.sb_secret}.*(created|unchanged|configured)', output)
     assert result is not None, f"Unable to apply YAML for binding secret '{context.sb_secret}': {output}"
+
+
+@given(u'CustomResourceDefinition backends.stable.example.com is available')
+def backends_crd(context):
```
pmacik, line 563: Could you please move the CRD YAML into the separate file under `test/acceptance/resources` and load it here from it, so it is not hard-coded?

pmacik: I can see that the actual CRD is already there as part of `test/acceptance/resources/backend-new-spec.operator.manifest.yaml`.

test/acceptance/features/postgresqlOperatorExamples.feature
```diff
@@ -0,0 +1,331 @@
+@examples
```
pedjak, line 1: are we sure that `@examples` scenarios are passing? How about that we create additional job that can run examples on demand/manually? https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/ That should be a part of the current PR.

test/acceptance/features/steps/steps.py
```diff
@@ -348,7 +348,12 @@ def operator_manifest_installed(context, backend_service):
         ns = context.namespace.name
     else:
         ns = None
-    _ = openshift.apply_yaml_file(os.path.join(os.getcwd(), "test/acceptance/resources/", backend_service + ".operator.manifest.yaml"), namespace=ns)
+
+    if backend_service == "backend":
+        _ = openshift.apply_yaml_file(os.path.join(os.getcwd(), "test/acceptance/resources/", backend_service + ".operator.manifest.yaml"), namespace=ns)
+        _ = openshift.apply_yaml_file(os.path.join(os.getcwd(), "test/acceptance/resources/backend_crd.yaml"), namespace=ns)
+    else:
+        _ = openshift.apply_yaml_file(os.path.join(os.getcwd(), "test/acceptance/resources/", backend_service + ".operator.manifest.yaml"), namespace=ns)
```
pedjak, line 356: let keep things simple - each fake operator is defined by single yaml file, i.e. all resources should be there, and not distributed across various files. In this way, the stop logic can be reverted to the previous state.

test/acceptance/features/steps/steps.py
```diff
@@ -557,3 +562,13 @@ def update_binding_secret(context):
...
+
+
+@given(u'CustomResourceDefinition backends.stable.example.com is available')
+def backends_crd(context):
+    openshift = Openshift()
+    if "namespace" in context:
+        ns = context.namespace.name
+    else:
+        ns = None
+    _ = openshift.apply_yaml_file(os.path.join(os.getcwd(), "test/acceptance/resources/backend_crd.yaml"), namespace=ns)
```
pedjak, line 574: please avoid DRY - this looks very similar to `operator_manifest_installed` method, let's generalize it if needed, and add additional step definition to it.

.github/workflows/examples_manually_trigger.yaml
```diff
@@ -0,0 +1,46 @@
+name: Run Examples
```
pedjak, line 1: is there any run of this job somewhere, so we can be sure it is green before merging this PR?

test/acceptance/features/steps/steps.py
pratikjagrut: Done.

test/acceptance/features/steps/steps.py
pratikjagrut: Done.

.github/workflows/examples_manually_trigger.yaml
pratikjagrut: I didn't find any way here, so I loaded this new job on my fork's master and I'll update the link once everything is fine.

.github/workflows/examples_manually_trigger.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+on: workflow_dispatch
+
+env:
+  GO111MODULE: on
+  SDK_VERSION: "1.3.0"
+  OPM_VERSION: "1.15.2"
```
Avni-Sharma, line 8: The target `release-operator` does not require opm . The target `prepare-operatorhub-pr` requires opm binary

pmacik: @Avni-Sharma `release-operator` actually does depend on `opm` transitively from [`index-image`](https://github.com/redhat-developer/service-binding-operator/blob/master/Makefile#L252) through `release-operator`->`push-index-image`->`index-image`

.github/workflows/examples_manually_trigger.yaml
pratikjagrut: @pedjak The test run of this workflow https://github.com/pratikjagrut/service-binding-operator/actions/workflows/examples_manually_trigger.yaml

pedjak (approved):
please squash

pratikjagrut:
> please squash

Squashed.

pratikjagrut:
Rebased.

PR #881, opened 2021-02-24
Consolidate roles landing into bundle

pedjak:
* Several roles were bound to a wrong service account, fixed to use `service-binding-operator`
* we are not protecting /metric endpoints for now, hence the auth proxy roles are removed from the bundle
* service binding viewer and editor clusterroles added to the bundle so that cluster roles can make use of them
  and enable appropriate users/service-accounts to manage service bindings

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ GIT_COMMIT_ID ?= $(shell git rev-parse --short=8 HEAD)
 OPERATOR_REGISTRY ?= quay.io
 OPERATOR_REPO_REF ?= $(OPERATOR_REGISTRY)/redhat-developer/servicebinding-operator
 OPERATOR_IMAGE_REF ?= $(OPERATOR_REPO_REF):$(GIT_COMMIT_ID)
+OPERATOR_IMAGE_SHA_REF ?= $(shell $(CONTAINER_RUNTIME) inspect --format='{{index .RepoDigests 0}}' $(OPERATOR_IMAGE_REF))
 OPERATOR_BUNDLE_IMAGE_REF ?= $(OPERATOR_REPO_REF):bundle-$(VERSION)-$(GIT_COMMIT_ID)
 OPERATOR_INDEX_IMAGE_REF ?= $(OPERATOR_REPO_REF):index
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -9,6 +9,7 @@ GIT_COMMIT_ID ?= $(shell git rev-parse --short=8 HEAD)
 OPERATOR_REGISTRY ?= quay.io
 OPERATOR_REPO_REF ?= $(OPERATOR_REGISTRY)/redhat-developer/servicebinding-operator
 OPERATOR_IMAGE_REF ?= $(OPERATOR_REPO_REF):$(GIT_COMMIT_ID)
+OPERATOR_IMAGE_SHA_REF ?= $(shell $(CONTAINER_RUNTIME) inspect --format='{{index .RepoDigests 0}}' $(OPERATOR_IMAGE_REF))
```
pmacik, line 12: Are we sure when this line gets evaluated? I mean, when the `docker inspect...` command is actually executed?
The thing is that the image SHA (a.k.a RepoDigest) is created/generated/assigned to the image only after pushing it into the repository. That is why the shell command to retrieve the digest was placed in-line.

In case the suggested line is actually evaluated before the image is actually pushed to the repository, it would be empty.

pedjak: it is a lazy evaluation, happens when the var is referenced, see: https://www.gnu.org/software/make/manual/make.html#Flavors

pedjak:
@pmacik `Bind an imported quarkus app which is deployed as knative service to PostgreSQL database` is failing both for OCP 4.6 and 4.7 - it looks something got changed on knative side again. Can we somehow stabilize it, or remove it?

Makefile
pmacik: Ok, cool!

pmacik:
> @pmacik `Bind an imported quarkus app which is deployed as knative service to PostgreSQL database` is failing both for OCP 4.6 and 4.7 - it looks something got changed on knative side again. Can we somehow stabilize it, or remove it?

@pedjak I'll take a look to fix it. However, the knative scenario will be tagged as `@example` and therefore excluded from the CI runs anyway, in https://github.com/redhat-developer/service-binding-operator/pull/880

PR #903, opened 2021-03-11
Drop rebinding on Service and workload/application changes

Avni-Sharma:
### Motivation:
Improving the stability of the SBO prompted reducing watches which would reduce the load on the API Server.

### Changes:
Watch only SBR resources and drop watches for services and application workloads.
Remove test cases, unit and acceptance, that watches on services and workloads.

### Testing:
Run `make test-acceptance`


controllers/mapper.go
```diff
@@ -206,7 +68,6 @@ type Referable interface {
 
 type K8STypeLookup interface {
 	ResourceForReferable(obj Referable) (*schema.GroupVersionResource, error)
-	KindForReferable(obj Referable) (*schema.GroupVersionKind, error)
```
pedjak, line 209: if this is not used anywhere, then please remove the implementation method from the same file.

controllers/mapper_test.go
```diff
@@ -59,29 +58,7 @@ func TestSBRRequestMapperMap(t *testing.T) {
...
-							Kind:       "Secret",
-						},
-					},
-				}
-			},
-			expectedRequestsLen: 0,
-		},
-		{
-			description: "resource declared as application of a service binding",
+			description: "Service Binding resource declared",
```
pedjak, line 61: can you also say something about test expectations?

pedjak (commented):
overall looks good, just a few small things to be addressed.

pmacik:
This PR has a pretty impressive impact on performance in a loaded cluster. The memory usage dropped from  up to ~11GiB to up to ~80 MiB (~140x less memory usage) and the resulted CPU usage dropped from up to ~4.5 vCPU to up to ~0.0034 vCPU (~1323,5x less CPU usage):
image
as opposed to what the performance was without this patch:
image

pedjak (approved):

PR #904, opened 2021-03-12
Fix non-OLM acceptance tests use SBO installed in Kubernetes, not local instance

pmacik:
### Motivation

Currently the `Acceptance tests running on Kubernetes without using OLM` GHA job runs the acceptance tests in a local mode (The default mode determined by the `TEST_ACCEPTANCE_START_SBO` variable (with the locally started instance of SBO) while it builds operator's images and than installs SBO into the Kubernetes, at the same time. The acceptnace tests should be run in a remote mode to use that instance, that is being installed into Kubernetes.

### Changes

This PR:
* Switches the non-OLM acceptance testing GitHub Action to remote mode by setting `TEST_ACCEPTANCE_START_SBO` variable to `remote`.
* Removes useless `OPERATOR_VERSION` variable

### Testing

This PR's GitHub Action check: `Acceptance tests running on Kubernetes without using OLM`

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -102,7 +102,6 @@ jobs:
     env:
       EXTRA_BEHAVE_ARGS: "--tags=~@knative --tags=~@openshift --tags=~@olm"
       UMOCI_VERSION: "0.4.5"
-      OPERATOR_VERSION: "0.4.0"
 
     steps:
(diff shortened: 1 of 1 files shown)


pedjak:
Thanks for this PR. Merging manually, because openshift CI has some issues, and the change in PR is only related to GitHub jobs.

cc @Avni-Sharma

PR #906, opened 2021-03-16
feat: allow constant values in CRD annotations (#843)

### LINKED ISSUE #843: Allow constant values in service.binding/name annotations in CRD
It would be cool to allow constant values in annotations so something like 

`service.binding/provider: postgresql`

For example, the `provider` and `type` entries are [recommended](https://github.com/k8s-service-bindings/spec#well-known-secret-entries) and can be the same across all underlying CR which warrants them to be constants. This also facilitates ad-hoc fixes. Thanks !

johnpoth:
### Motivation

fixes #843. There also is some interest for [kafka](https://github.com/quarkusio/quarkus/issues/15631#issuecomment-799305457) use case. Thanks !

### Changes

Add `constantDefinition` 

### Testing

make test

Thanks !

### DIFF
--- a/docs/User_Guide.md
+++ b/docs/User_Guide.md
@@ -869,3 +869,12 @@ spec:
   ```
     service.binding: path={.data},elementType=map
   ```
+11. ### Use a constant
+
+  Requirement: *Use a constant to be present in all binding Secrets*
+
+  Annotation
+
+  ```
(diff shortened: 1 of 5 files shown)


DhritiShikhar:
@johnpoth Can you please add an acceptance test for this scenario?

PR #908, opened 2021-03-17
update gogo/protobuf

cdlliuy:
There is a known vulnerability exposed for gogo/protobuf, and fixed in its version v1.3.2

But currently,  per the dependency of k8s claim in `go.mod`,  only v1.3.1 is pickedup. 
The kubernetes dependency will be only picked v1.3.2 version of gogo/protobuf  in v0.21.x,  which are a little bit far away from service binding operator current `requires`
i.e.
https://github.com/kubernetes/apiserver/commit/0d37c2610c65d617d6f9efbf7f39ce08c6c2cf39#diff-33ef32bf6c23acb95f5902d[…]8ca061167ec0716715b0b9eeaa5f6 

So,  I proposed to add the correct version as a `indirect` require directly


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.15
 require (
 	github.com/go-logr/logr v0.3.0
 	github.com/go-logr/zapr v0.3.0 // indirect
+	github.com/gogo/protobuf v1.3.2 // indirect
```
pedjak, line 8: where is the replace directive that fix the version?

cdlliuy: yes, I updated the commit.  

Previously, I am using the `replace` , but with the latest push,  I changed to use indirect require.

PR #912, opened 2021-03-24
Add check for the presence of conflict into linters

pmacik:
### Motivation

Currently it is possible to accidently commit conflict notes (such as `<<<<<<< HEAD`) in source files without actually noticing it.

### Changes

This PR:
* Adds a new linter for detecting git conflict notes in tracked and staged files.

### Testing

`make lint-conflicts`

### DIFF
--- a/hack/check-conflicts.sh
+++ b/hack/check-conflicts.sh
@@ -1,0 +1,34 @@
+#!/bin/bash -e
+
+echo -e "\nChecking for presence of conflict notes in source files..."
+
+check_results(){
+    found=0
+    for f in "$@"; do
+        grep -PiIrnH '<<<<''<<<|>>>''>>>>|===''====' "$f"
+        if [ $? -ne 1 ]; then
+            found=1
+        fi
+    done
+    if [ $found -eq 1 ]; then
+        echo FAIL
+    else
+        echo PASS
+    fi
+    return $found
+}
+
+overall_failed=0
+
+echo -e "\nChecking staged files... "
+check_results $(git diff --cached --name-only | grep -v "vendor/") || overall_failed=1
+
+echo -e "\nChecking tracked files... "
+check_results $(git ls-tree --full-tree -r HEAD --name-only | grep -v "vendor/") || overall_failed=1
+
+if [ $overall_failed -eq 0 ]; then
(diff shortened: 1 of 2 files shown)


hack/check-conflicts.sh
```diff
@@ -0,0 +1,41 @@
...
+    else
+        echo FAIL
+        echo ""
+        echo "$result"
+        echo ""
+        overall_failed=1
+    fi
+else
+    echo PASS
+fi
```
pedjak, line 21: why checking files on stagging, when is this useful? We do not care if there are conflicts until they are checked into the repo, rigth?

pmacik: It is useful for checking the new files you are about to commit (that are not tracked yet) to avoid accidental commits.

pedjak: ok, make sense - would you mind to wrap then these check in a function? you perform the identical check two times, once on staged files, another time on files in repo?

pmacik: @pedjak Fixed that, thx

hack/check-conflicts.sh
```diff
@@ -0,0 +1,38 @@
...
+        echo "$1"
+        echo ""
+        return 1
+    fi
+    return 0
+}
+
+echo -ne "\n\tChecking staged files... "
+if [ $(git diff --cached --name-only | wc -l) != "0" ]; then
+    results=$(git diff --cached --name-only | grep -v "vendor/" | xargs grep -PiIrnH '<<<<''<<<|>>>''>>>>|===''====' || true)
```
pedjak, line 22: we can move even check logic into `check_result` (or whatever we call it later on) function. Something along these lines:

```
check_merge_conflicts() {
  for f in "$@"; do 
    grep -PiIrnH '<<<<''<<<|>>>''>>>>|===''====' "$f" && echo "$f" && break
  done || found=1
if [ -z "$found" ]; then
  echo PASS
else
  echo FAIL
  exit 1
fi
}

check_merge_conflicts $(git diff --cached --name-only | grep -v "vendor/")
check_merge_conflicts $(git ls-tree --full-tree -r HEAD --name-only | grep -v "vendor/")
```

PR #913, opened 2021-03-25
Successful Service Binding Resource should be Immutable

### LINKED ISSUE #896: Make Successful Service Binding Resource Immutable
Reference EPIC #892 

Service Binding resources should be immutable.

Avni-Sharma:
When `Ready` condition gets equal to `True`, the service binding cannot be updated anymore.
However, it can be removed, and recreated. The described behavior has been achieved by
adding validation webhooks, for both APIs. Further changes:

* `kustomize` resources under `config` folder were simplified, so that generated OLM bundle is OCP 4.6+ compatible
* both OLM bundle and `release.yaml` are generated now by using `kustomize` only and from files under `config` folder, hence `/hack/release-manifest.py` has been removed
* non-OLM installation requires now `cert-manager` to be available on the cluster
* removed acceptance tests that were demonstrating service binding updates
* new acceptance tests added to verify that ready service binding cannot be updated

fixes #896


test/acceptance/features/immutabeSuccessfulBinding.feature
```diff
@@ -0,0 +1,56 @@
+Feature: Successful Service Binding Resource should be Immutable 
+
+    As a user of Service Binding operator
+    I should not be able to apply changes of an already appled and successful SBR
```
pedjak, line 4: typo: appled

what is SBR?

test/acceptance/features/immutabeSuccessfulBinding.feature
```diff
@@ -0,0 +1,56 @@
...
+            apiVersion: binding.operators.coreos.com/v1alpha1
+            kind: ServiceBinding
+            metadata:
+                name: sbr-immutable
+            spec:
+                services:
+                    group: stable.example.com
+                    version: v1
+                    kind: Backend
+                    name: service-immutable
```
pedjak, line 35: this should be list, right? like:

```
- group: stable.example.com
  version: v1
  kind: Backend
  name: service-immutable
```

test/acceptance/features/immutabeSuccessfulBinding.feature
```diff
@@ -0,0 +1,56 @@
...
+        kind: Backend
+        metadata:
+            name: service-immutable
+            annotations:
+                service.binding/host: path={.spec.host}
+        spec:
+            host: foo
+        """
+        And Generic test application "app-immutable" is running
+        When Service Binding is applied
```
pedjak, line 24: Use `And` here, because this is still test setup. We should have only one `When` and `Then` block

test/acceptance/features/immutabeSuccessfulBinding.feature
```diff
@@ -0,0 +1,56 @@
...
+            kind: ServiceBinding
+            metadata:
+                name: sbr-immutable
+            spec:
+                application:
+                    name: app-immutable
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
```
pedjak, line 55: This would fail even today on master, because `services` must be defined, i.e. they could not be removed. You need to come up with some other valid Service Binding.

test/acceptance/features/mutableFailingSBR.feature
```diff
@@ -0,0 +1,116 @@
+Feature: Unsuccessful Service Binding Resource should be mutable
+
+    As a user of Service Binding operator
+    I should be able to reapply my SBR since the previous SBR was unsuccessful
```
pedjak, line 4: if it was unsuccessful, we are not reapplying the binding, instead we update the binding. Can we reflect it in the above text?

test/acceptance/features/mutableFailingSBR.feature
```diff
@@ -0,0 +1,116 @@
...
+
+    Background:
+        Given Namespace [TEST_NAMESPACE] is used
+        * Service Binding Operator is running
+        * CustomResourceDefinition backends.stable.example.com is available
+
+    Scenario: Reapply SBR when SB CR status ready is false due to Application Not Found
```
pedjak, line 11: see above comment.

test/acceptance/features/mutableFailingSBR.feature
```diff
@@ -0,0 +1,116 @@
...
+                    kind: Backend
+                    name: service2
+                application:
+                    name: app3
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        Then Service Binding "sbr-1" is not ready
+        And The service "service2" does not exist
```
pedjak, line 86: not sure if we need to check this explicitly at all

pedjak (changes requested):
there are also some existing tests that will start to fail because of immutability. They should be removed as well.

test/acceptance/features/immutabeSuccessfulBinding.feature
Avni-Sharma: Done

test/acceptance/features/mutableFailingSBR.feature
Avni-Sharma: While we can mention update, yes, in the text, that would map to a kubectl apply action right? 
For example, I am envisioning a scenario wherein I applied a yaml and my Status ready is false. Then I am reapplying as a user.(not waiting on reconcile as reconcile is triggered on create and delete).

test/acceptance/features/immutabeSuccessfulBinding.feature
Avni-Sharma: Done

test/acceptance/features/immutabeSuccessfulBinding.feature
Avni-Sharma: yes. Done

test/acceptance/features/immutabeSuccessfulBinding.feature
Avni-Sharma: fixed typo. SBR is Service Binding Resource

Avni-Sharma:
> there are also some existing tests that will start to fail because of immutability. They should be removed as well.

Those tests were removed as part of drop watches(wherein we were reapplying resources and waiting for reconcile to capture those) imo acceptance tests arent missed. But can you please point out some that you think might fail.

test/acceptance/features/immutabeSuccessfulBinding.feature
pedjak: ok, then please avoid the abbreviation for improved clarity.

test/acceptance/features/mutableFailingSBR.feature
pedjak: ok

Avni-Sharma:
There is a scenario After updating mountpath bindings are reinjected to the new location
Here the same SBR resource is being updated with new mountPath . Should we let go of this scenario?

DhritiShikhar:
Hey @Avni-Sharma are you going to continue to work on this PR?

pmacik:
@isutton FYI, PR checks fail because of an issue with the workflows. This PR https://github.com/redhat-developer/service-binding-operator/pull/973 fixes that.

test/acceptance/features/mutableFailingSBR.feature
isutton: I've kept this in the "arrange" part of the test.

test/acceptance/features/mutableFailingSBR.feature
isutton: Changed the copy.

config/manager/manager.yaml
```diff
@@ -22,7 +22,7 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
-      serviceAccountName: service-binding-operator
+      serviceAccountName: default
```
isutton, line 25: Should be reverted, we still should use service-binding-operator account name.

config/rbac/role_binding.yaml
```diff
@@ -8,5 +8,5 @@ roleRef:
   name: manager-role
 subjects:
 - kind: ServiceAccount
-  name: service-binding-operator
+  name: default
```
isutton, line 11: Should be reverted, we still should use service-binding-operator account name.

test/acceptance/features/readyServiceBinding.feature
```diff
@@ -0,0 +1,60 @@
+Feature: Ready Service Binding Should Be Read-Only
```
isutton, line 1: Merge feature files in one, and rename feature to "Successful Service Binding Resources are Immutable".

test/acceptance/features/readyServiceBinding.feature
```diff
@@ -0,0 +1,60 @@
...
+                    kind: Backend
+                    name: service-immutable
+                application:
+                    name: app-immutable
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        When Service Binding "sbr-immutable" is ready
+        Then Immutable Service Binding is unable to be applied
```
isutton, line 43: Adjust the workflow to:

```
And Service Binding "sbr..." is ready
When Service Binding is applied ""
Then $ERROR_STEP
```

test/acceptance/features/unreadyServiceBinding.feature
```diff
@@ -0,0 +1,116 @@
...
+
+    As a user of Service Binding operator
+    I should be able to modify a Service Binding if it is not yet ready
+
+    Background:
+        Given Namespace [TEST_NAMESPACE] is used
+        * Service Binding Operator is running
+        * CustomResourceDefinition backends.stable.example.com is available
+
+    Scenario: Allow modifying a Service Binding when it is not yet ready due to "Application Not Found"
```
isutton, line 11: Scenario should be the opposite of scenario present in the the ready service binding feature file.

This scenario should be moved next to the one in ready service binding feature file.

baijum:
@isutton @pedjak Can we introduce an annotation to the ServiceBinding resource to control mutability. Something like this:

```
binding.operators.coreos.com/mutable: true
```

I am not sure what should be the default value. But an option like this would help us to meet the requirements of those who want mutability and those who don't want it.

If this sounds like a good idea, I can create a separate issue. Let me know your thoughts.

baijum:
We need to document this feature with expected behavior in the user guide.

Some notes:

1. After a successfull completion of Service Binding reconcilation, the Service Binding resource will be immutable. However the resource can be deleted and recreated to trigger the reconciliation.
2. After a successfull completion of Service Binding reconcilation, if there is any change in the backing service, it will not be immediately reflected in the application.
   So, if the application is misbehaving, users can delete the Service Binding resource and recreate it again.
3. After a successfull completion of Service Binding reconcilation, if there is a change in the application related to binding, say the environment variable has been removed, it will not be added back to it.
3. After a successfull completion of Service Binding reconcilation, if there is a new application comes up matching the label, the binding is not going to take place.

config/manager/manager.yaml
isutton: Done in a37d1320.

config/rbac/role_binding.yaml
isutton: Done in a37d1320.

isutton:
There are a couple of options to proceed with testing the admission
webhook in this PR:

1.  Extract OpenShift's `tls.key` and `tls.crt` files to
    `/tmp/k8s-webhook-server/serving-certs`, which is the default
    location used by the ServiceBinding Operator (required to receive
    payloads from the API server).
    -   Requires logic to extract `tls.key` and `tls.crt` files from vanilla
        Kubernetes.
    -   Requires configuration to access the mutation webhook available
        outside the cluster.
    -   Is it possible to use `kubectl port-forward` for this?
        -   It might be possible that `kubectl port-forward` could help with
            this, as it is possible to define a specific `service`, although
            to be seen is whether it requires at least one active Pod
            available, or `kube port-forward` does work without a running
            Pod, making the client effectively a Pod of some sorts.

2.  Build the image locally, push to `quay.io` and execute the tests
    using a `remote` instance of the ServiceBinding Operator initialized
    through `make deploy`.
    
    This workflow invokes the following operations, regardless if the
    build happens at the same physical host as the destination cluster:
    `build` -> `push to remote` -> `pull from remote`.
    
    This solution seems to be the least effort, but waste a lot of
    resources since even when testing on the same host, a roundtrip to
    `quay.io` is required. If this approach is chosen, must be considered
    technical debt and scheduled for further developments.

3.  Build the image locally, and copy it to the destination cluster
    using `skopeo`; from what I've seen it can not copy or sync images
    from local Docker registry cache, only from the remote registry in
    the case `--src docker` is informed.

4.  Do not rely on a Kubernetes or OpenShift cluster, and adapt the
    integration tests to use [envtest](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest) instead.

pedjak:
> There are a couple of options to proceed with testing the admission
webhook in this PR:

I would suggest that we keep the scope of this PR focused and avoid spending time on crafting a superb dev experience, covering both CRC and Minikube. For local testing and for most of the cases, we can have an option to disable webhooks, so that we can still start a local process to connect to a cluster. Probably that could be also the default behaviour as well.

Otherwise, for running all acceptance tests, I would pick today option number 2, and work on improving over time. There is no need to use quay at all, if using minikube, we can use local cluster registry, and hence even run the tests without an internet connection. This is how our CI jobs are running tests even today, see:

https://github.com/redhat-developer/service-binding-operator/blob/4cf9811fe24804cc884ea0e2c2a83bd4cd065fd9/.github/workflows/pr-checks.yaml#L97

in case of CRC you can either use its local registry or have another registry running on your machine and use the same approach. IMHO, there is really no need to use CRC locally anymore, minikube works great and covers all our needs meanwhile.

cc @pmacik - wdyt?

pmacik:
@isutton SBO's deployment is most likely called `controller-manager` (https://github.com/redhat-developer/service-binding-operator/pull/913/files#diff-acdef1a63fe932b7de3c48f1a4fce36f79787a3f5b9136539c3d214491a27cacR4), however your  detector in tests is looking for `service-binding-operator-controller-manager` (https://github.com/redhat-developer/service-binding-operator/pull/913/files#diff-6b43685101835ebb3b69db6de7d76eb64e396b3cdf560619630e0055d4c593b8R44) that is why it does not find it

The `Collect Logs` in GHA (the archived sbo.log) does not care about the particular name - it takes whatever is there (https://github.com/redhat-developer/service-binding-operator/blob/master/.github/workflows/pr-checks.yaml#L114) so that is why the log is fine.

isutton:
> @isutton SBO's deployment is `service-binding-operator`
> (https://github.com/redhat-developer/service-binding-operator/pull/913/checks?check_run_id=2804236225#step:8:72),
> however your detector in tests is looking for
> `service-binding-operator-controller-manager`

`make deploy` uses kustomize, [which I could find](https://github.com/redhat-developer/service-binding-operator/blob/immutable-sbr/config/default/kustomization.yaml#L4-L9) a `namePrefix` field set to  `service-binding-operator-`, resulting in a Deployment named `service-binding-operator-controller-manager`, as seen below:

	$ kubectl -n service-binding-operator-system get all
	NAME                                                               READY   STATUS              RESTARTS   AGE
	pod/service-binding-operator-controller-manager-68fbd6f645-9grwz   0/1     ContainerCreating   0          4m37s

	NAME                                               TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
	service/service-binding-operator-webhook-service   ClusterIP   10.97.49.240   <none>        443/TCP   4m37s

	NAME                                                          READY   UP-TO-DATE   AVAILABLE   AGE
	deployment.apps/service-binding-operator-controller-manager   0/1     1            0           4m37s

	NAME                                                                     DESIRED   CURRENT   READY   AGE
	replicaset.apps/service-binding-operator-controller-manager-68fbd6f645   1         1         0       4m37s

To build the image, I've used `make image OPERATOR_REGISTRY=$(minikube ip):5000` producing the image `192.168.49.2:5000/redhat-developer/servicebinding-operator:991bc81d`, which looks correct. The image is being built using Minikube's Docker, which has been exposed by running `eval $(minikube docker-env)`.

[As seen in here](https://github.com/redhat-developer/service-binding-operator/blob/immutable-sbr/config/manager/manager.yaml#L33-L37) `WATCH_NAMESPACE` is being set by default to the resource's namespace:

    env:
    - name: WATCH_NAMESPACE
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.namespace


@pedjak If you do not oppose, I'll remove this configuration from `manager.yaml`.

test/acceptance/features/environment.py
```diff
@@ -41,7 +41,7 @@ def before_all(_context):
     elif start_sbo == "remote":
         output, code = cmd.run(
             f"{ctx.cli} get deployment --all-namespaces -o json"
-            + " | jq -rc '.items[] | select(.metadata.name == \"service-binding-operator\").metadata.namespace'")
+            + " | jq -rc '.items[] | select(.metadata.name == \"service-binding-operator-controller-manager\").metadata.namespace'")
```
pedjak, line 44: this line should be reverted, our operator deployment should be still called `service-binding-operator`, at least in this PR.

test/acceptance/features/steps/servicebindingoperator.py
```diff
@@ -20,7 +20,7 @@ def check_crd(self):
         return True
 
     def check_deployment(self):
-        sbo_namespace = self.openshift.lookup_namespace_for_resource("deployments", "service-binding-operator")
+        sbo_namespace = self.openshift.lookup_namespace_for_resource("deployments", "service-binding-operator-controller-manager")
```
pedjak, line 23: this line should be reverted as well, see above.

pedjak:
> env:
> - name: WATCH_NAMESPACE
>  valueFrom:
 >   fieldRef:
 >     apiVersion: v1
 >     fieldPath: metadata.namespace
>
> @pedjak If you do not oppose, I'll remove this configuration from manager.yaml.

How is this connected to the work done in this PR? That env var is usually empty, but its purpose is to configure the operator at the installation time so that it can just watch for changes in the given namespace.

test/acceptance/features/environment.py
isutton: I changed this because the deployment name, when installing using `make deploy`, is called `service-binding-operator-controller-manager` (see the output of `./bin/kustomize build config/default`):

```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    control-plane: controller-manager
  name: service-binding-operator-controller-manager
  namespace: service-binding-operator-system
spec:
  replicas: 1
  selector:
    matchLabels:
      control-plane: controller-manager
  template:
    metadata:
      labels:
        control-plane: controller-manager
    spec:
      containers:
      - args:
        - --leader-elect
        command:
        - /manager
        image: 192.168.49.2:5000/sbo:991bc81d
```

I got into this command from the following log:

```
$ make deploy                                    \
	   OPERATOR_REPO_REF=$(minikube ip):5000/sbo \
	   SKIP_REGISTRY_LOGIN=true
> > /home/isuttonl/src/service-binding-operator/bin/controller-gen "crd:trivialVersions=true,preserveUnknownFields=false" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
docker build -f Dockerfile -t 192.168.49.2:5000/sbo:991bc81d .
Sending build context to Docker daemon  53.19MB
(snip)
cd config/manager && /home/isuttonl/src/service-binding-operator/bin/kustomize edit set image controller=192.168.49.2:5000/sbo:991bc81d
/home/isuttonl/src/service-binding-operator/bin/kustomize build config/default | kubectl apply -f -
(snip)
```

Then confirmed in the cluster:

```
$ kubectl -n service-binding-operator-system get all
NAME                                                               READY   STATUS              RESTARTS   AGE
pod/service-binding-operator-controller-manager-64dc64c7cb-x59bl   0/1     ContainerCreating   0          22m
pod/service-binding-operator-controller-manager-68fbd6f645-wfwjf   0/1     ContainerCreating   0          3h20m

NAME                                               TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
service/service-binding-operator-webhook-service   ClusterIP   10.97.49.240   <none>        443/TCP   3h26m

NAME                                                          READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/service-binding-operator-controller-manager   0/1     1            0           3h26m
```

.github/workflows/pr-checks.yaml
```diff
@@ -181,7 +181,12 @@ jobs:
 
           eval $(minikube docker-env)
           make release-manifests
+          make deploy-cert-manager
```
pedjak, line 184: we can combine this with previous line:

```
make deploy-cert-manager release-manifest
```

.github/workflows/pr-checks.yaml
```diff
@@ -181,7 +181,12 @@ jobs:
...
+          echo "Waiting 120s for cert-manager to start..."
+          sleep 120
```
pedjak, line 186: instead of this, could we wait on cert manager using something like

```
kubectl rollout status ....
```

Makefile
```diff
@@ -132,9 +132,13 @@ install: manifests kustomize
 uninstall: manifests kustomize
 	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
+.PHONY: deploy-cert-manager
+deploy-cert-manager:
+	kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.1.0/cert-manager.yaml
```
pedjak, line 137: perhaps here we can wait that cert manageris up and running? That would benefical for the local dev cycle as well?

api/v1alpha1/servicebinding_webhook.go
```diff
@@ -0,0 +1,71 @@
...
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
+// +kubebuilder:webhook:path=/validate-binding-operators-coreos-com-v1alpha1-servicebinding,mutating=false,failurePolicy=fail,sideEffects=None,groups=binding.operators.coreos.com,resources=servicebindings,verbs=create;update,versions=v1alpha1,name=vservicebinding.kb.io,admissionReviewVersions={v1beta1}
+
+var _ webhook.Validator = &ServiceBinding{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *ServiceBinding) ValidateCreate() error {
+	servicebindinglog.Info("validate create", "name", r.Name)
```
pedjak, line 44: do we really need this log statement? it is going to make a lot of noise, right?

api/v1alpha1/servicebinding_webhook.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// TODO(user): fill in your validation logic upon object creation.
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *ServiceBinding) ValidateUpdate(old runtime.Object) error {
+	servicebindinglog.Info("validate update", "name", r.Name)
```
pedjak, line 52: same as above?

api/v1alpha1/servicebinding_webhook.go
```diff
@@ -0,0 +1,71 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *ServiceBinding) ValidateDelete() error {
+	servicebindinglog.Info("validate delete", "name", r.Name)
```
pedjak, line 67: same as above, not seeing any value in this statement.

config/crd/kustomization.yaml
```diff
@@ -8,12 +8,12 @@ resources:
 patchesStrategicMerge:
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
 # patches here are for enabling the conversion webhook for each CRD
-#- patches/webhook_in_servicebindings.yaml
+# - patches/webhook_in_servicebindings.yaml
```
pedjak, line 11: look like no-op change.

config/crd/patches/cainjection_in_servicebindings.yaml
```diff
@@ -2,6 +2,4 @@
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
-  annotations:
-    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
```
pedjak, line 6: why do we need to remove these lines?

hack/release-manifest.py
```diff
@@ -54,51 +61,151 @@ def create_cluster_role_yaml(csv):
...
+        "apiVersion" : 'v1',
+        "kind" : 'ServiceAccount',
+        "metadata" : {
+            "name" : 'service-binding-operator',
+            "namespace" : 'service-binding-operator'
+        },
+    }
     return service_account
 
+kind2prefix = {
```
pedjak, line 147: can we come up with some better name?

test/acceptance/features/immutable-ready-service-binding-resources.feature
```diff
@@ -0,0 +1,167 @@
...
+                    kind: Backend
+                    name: service-immutable
+                application:
+                    name: app-immutable
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        When Service Binding "sbr-immutable" is ready
+        Then Immutable Service Binding is unable to be applied
```
pedjak, line 43: not sure what "Immutable" here means, but IMHO it would read better if we say:

```
Service Binding cannot be applied
```

test/acceptance/features/immutable-ready-service-binding-resources.feature
```diff
@@ -0,0 +1,167 @@
...
+                    kind: Backend
+                    name: service1
+                application:
+                    name: app1
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        And jq ".status.conditions[] | select(.type=="Ready").status" of Service Binding "sbr-1" should be changed to "False"
+        And The application "app1" does not exist
```
pedjak, line 93: not sure if this step adds any value, because when reading the scenario it is obvious that the application is not available, because it is installed in the next step. IMHO, I would remove this step.

test/acceptance/features/immutable-ready-service-binding-resources.feature
```diff
@@ -0,0 +1,167 @@
...
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        Then Service Binding "sbr-1" is ready
+
+
+    Scenario: Allow modifying a Service Binding when it is not yet ready due to "Service Not Found"
+        Given Generic test application "app3" is running
+        And The service "service2" does not exist
```
pedjak, line 118: this step does not provide any added value, because clearly the service is not declared earlier. IMHO I would remove it.

pedjak (changes requested):

hack/release-manifest.py
isutton: How would you call a dictionary or map that maps a Kind to a Prefix? I'm not sure `kindToPrefixMap` or `kindToPrefixDict` would be much more descriptive.

pedjak: IMHO, you do not need this map at all, the most of the content can be simply copied from CSV, here is what is to be found there:

```yaml
webhookdefinitions:
  - admissionReviewVersions:
    - v1beta1
    containerPort: 443
    deploymentName: service-binding-operator
    failurePolicy: Fail
    generateName: vservicebinding.kb.io
    rules:
    - apiGroups:
      - binding.operators.coreos.com
      apiVersions:
      - v1alpha1
      operations:
      - CREATE
      - UPDATE
      resources:
      - servicebindings
    sideEffects: None
    targetPort: 9443
    type: ValidatingAdmissionWebhook
    webhookPath: /validate-binding-operators-coreos-com-v1alpha1-servicebinding
```

We only need to:

* rename `generateName` to `name`
* create `clientConfig` and use `webhookPath` as `path`

`type` should be used for deciding what kind of configuration will be made.

hack/release-manifest.py
```diff
@@ -54,51 +61,151 @@ def create_cluster_role_yaml(csv):
...
+        }
+    }
+
+def create_webhook_certificate():
+    return {
+        "apiVersion": "cert-manager.io/v1",
+        "kind": "Certificate",
+        "metadata": {
+            "name": "service-binding-operator-serving-cert",
+            "namespace": "service-binding-operator",
```
pedjak, line 121: can we create a global var with `service-binding-operator` namespace and use it in the script - it appears in some many places already.

.github/workflows/pr-checks.yaml
isutton: Kept both separated since `deploy-cert-manager` is paired with a `kubectl rollout status`. This might change though if the rollout status is moved to the Makefile's target `deploy-cert-manager`.

.github/workflows/pr-checks.yaml
isutton: Done in 916e27d0e405455eaf20b87c8bbbf964591b0fc9.

.github/workflows/pr-checks.yaml
isutton: Done in 176d73f10d5047ed3b2383d4136d23e693c874a5.

Makefile
isutton: Done in 176d73f10d5047ed3b2383d4136d23e693c874a5.

api/v1alpha1/servicebinding_webhook.go
isutton: Done in 6f7cf943ef89efd6153fa55e3cc12842bea92415.

api/v1alpha1/servicebinding_webhook.go
isutton: Done in 6f7cf943ef89efd6153fa55e3cc12842bea92415.

api/v1alpha1/servicebinding_webhook.go
isutton: Done in 6f7cf943ef89efd6153fa55e3cc12842bea92415.

config/crd/kustomization.yaml
isutton: @pedjak It seems this change comes from ea79796cba3f5a82e2343d760785179eaa02c3fe, so I'd leave it be unless there's a stronger reason to revert this.

config/crd/patches/cainjection_in_servicebindings.yaml
isutton: From what I could understand, this interfered with the same annotation configured by `release-manifest.py`.

pedjak (approved):
approving since I contributed to PR meanwhile, but therefore asking @baijum and @pmacik to review it.

test/acceptance/features/immutableBindings.feature
```diff
@@ -0,0 +1,192 @@
...
+                    version: v1
+                    kind: Backend
+                    name: service2
+                application:
+                    name: app3
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        And jq ".status.conditions[] | select(.type=="Ready").status" of Service Binding "binding-immutable-3" should be changed to "False"
```
pmacik, line 124: A step for verifying the actual reason is "Service Not Found" defined in the scenario should be added here so that the test fails propely on any other reason then expected.

test/acceptance/features/immutableBindings.feature
```diff
@@ -0,0 +1,192 @@
...
+                    kind: Backend
+                    name: service-immutable
+                application:
+                    name: app-immutable
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        When Service Binding "binding-immutable" is ready
+        Then Immutable Service Binding is unable to be applied
```
pmacik, line 43: IMHO, the word "Immutable" is unnecessary here, in the step name. Every binding that is ready is supposed to be immutable, so I suggest to name it just `Service Binding is unable to be applied"

test/acceptance/features/immutableBindings.feature
```diff
@@ -0,0 +1,192 @@
...
+                  apiVersion: stable.example.com/v1
+                  kind: Backend
+                  name: service-immutable
+                application:
+                    name: spec-app-immutable
+                    apiVersion: apps/v1
+                    kind: Deployment
+            """
+        When Service Binding "spec-binding-immutable" is ready
+        Then Immutable Service Binding is unable to be applied
```
pmacik, line 177: Same as above

test/acceptance/features/steps/service_binding.py
```diff
@@ -68,6 +68,7 @@ def sbr_is_applied(context):
 
 
 @when(u'Invalid Service Binding is applied')
+@then(u'Immutable Service Binding is unable to be applied')
```
pmacik, line 71: Same as above

test/acceptance/features/immutableBindings.feature
```diff
@@ -0,0 +1,192 @@
...
+                    version: v1
+                    kind: Backend
+                    name: service-immutable
+                application:
+                    name: app1
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        And jq ".status.conditions[] | select(.type=="Ready").status" of Service Binding "binding-immutable-2" should be changed to "False"
```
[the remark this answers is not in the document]
pedjak: The focus of this test is not on testing if `reason` is populated accordingly - it assert that a service binding cannot be updated if `Ready == true`, the reasons for that are unimportant. This steps is a part of test setup, it is not part of `Then` block. Hence, I would propose that we add such dedicated test in a separate PR if needed

test/acceptance/features/immutableBindings.feature
pedjak: see above.

test/acceptance/features/immutableBindings.feature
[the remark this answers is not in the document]
pmacik: What you said makes sense. However, my point is to ensure the initial conditions (the test setup) are correct - that the SB resource is not ready due to a reason that is specified in the scenario - before checking that it behaves as expected in that particular situation. I did not mean to test that `reason` is populated properly here - that should indeed be part of another scenario.

That detail is part of the `Scenario` definition, so IMHO it is important to make sure that it is met.

test/acceptance/features/immutableBindings.feature
```diff
@@ -0,0 +1,140 @@
...
+        metadata:
+            name: service-immutable
+            annotations:
+                service.binding/host: path={.spec.host}
+        spec:
+            host: foo
+        """
+
+    Scenario: Cannot update a ready Service Binding
+        And Generic test application "app-immutable" is running
```
pmacik, line 23: This should start with `Given` for better readability

test/acceptance/features/immutableBindings.feature
```diff
@@ -0,0 +1,140 @@
...
+                    resource: deployments
+                services:
+                  - group: stable.example.com
+                    version: v1
+                    kind: Backend
+                    name: service-immutable
+            """
+
+    Scenario: Allow modifying a not-ready Service Binding
+        And Service Binding is applied
```
pmacik, line 63: This should start with `Given` for better readability

test/acceptance/features/immutableBindings.feature
```diff
@@ -0,0 +1,140 @@
...
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        Then Service Binding "binding-immutable-2" is ready
+
+
+    @spec
+    Scenario: SPEC Cannot update a ready Service Binding
+        And Generic test application "spec-app-immutable" is running
```
pmacik, line 106: This should start with `Given` for better readability

test/acceptance/features/immutableBindings.feature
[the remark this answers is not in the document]
pmacik: Aggreed that the actual reason is not a focus here since it is no longer mentioned in the `Scenario` definition.

baijum:
Overall looks good to me. The suggestion from Pavel about tests should be addressed.

/lgtm

PR #914, opened 2021-03-25
Remove output redirect in install.sh to file named 1 

pmacik:
### Motivation

Currently, `install.sh` script produces unused file named `1` as a result of an output redirect.

### Changes

This PR:
* Removes the redirect in  `install.sh` script to file named `1` to keep the output visible.

### Testing

After running `./install.sh` check the root directory for the presence of a file name `1` it should not be there.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -6,9 +6,8 @@ OPERATOR_PACKAGE=${OPERATOR_PACKAGE:-service-binding-operator}
 DOCKER_CFG=${DOCKER_CFG:-$HOME/.docker/config.json}
 CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-docker}
 
-if [ "$(kubectl get crd clusterserviceversions.operators.coreos.com &>1 || echo false)" == "false" ]; then
+kubectl get crd clusterserviceversions.operators.coreos.com || \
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -6,7 +6,7 @@ OPERATOR_PACKAGE=${OPERATOR_PACKAGE:-service-binding-operator}
 DOCKER_CFG=${DOCKER_CFG:-$HOME/.docker/config.json}
 CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-docker}
 
-if [ "$(kubectl get crd clusterserviceversions.operators.coreos.com &>1 || echo false)" == "false" ]; then
+if [ "$(kubectl get crd clusterserviceversions.operators.coreos.com &>/dev/null || echo false)" == "false" ]; then
```
pedjak, line 9: IMHO, it would be good to see errors, it can help with debugging. Hence, the fix could add the missing `&` in front of `1`

While we are here, this `if` block can be simplified further:

```
kubectl get crd clusterserviceversions.operators.coreos.com || \
  curl -sL https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v${OLM_VERSION}/install.sh | bash -s v${OLM_VERSION}
```

pmacik: @pedjak  `&>&1` would not work (https://stackoverflow.com/questions/24793069/what-does-do-in-bash/24793436#24793436)

But I take your further suggestion. Thx

pedjak (approved):

PR #916, opened 2021-03-25
Collect the operator logs after running PR test jobs

pedjak:
The logs are copied into `sbo.log` file and packed together with other build artifacts

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -15,7 +15,7 @@ env:
   MINIKUBE_VERSION: "1.15.1"
   TEST_ACCEPTANCE_CLI: "kubectl"
   CONTAINER_RUNTIME: "docker"
-  TEST_RESULTS: "out/acceptance-tests/TEST*.xml"
+  TEST_RESULTS: "out/acceptance-tests"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-checks.yaml
```diff
@@ -15,7 +15,7 @@ env:
   MINIKUBE_VERSION: "1.15.1"
   TEST_ACCEPTANCE_CLI: "kubectl"
   CONTAINER_RUNTIME: "docker"
-  TEST_RESULTS: "out/acceptance-tests/TEST*.xml"
```
pmacik, line 18: Testspace (later on) requires set of result files (rather then the whole directory), otherwise the publishing of the results fails with [`Error: non-existing result file "out/acceptance-tests" specified!`](https://github.com/redhat-developer/service-binding-operator/runs/2197292599?check_suite_focus=true#step:9:23).

So, to remove `TEST*.xml` from this variable to be able to use it also for the archiving the artifacts you also need to append this to the `Publish tests results to Testspace` steps like: `testspace [${{ env.TEST_RUN }}]${{ env.TEST_RESULTS }}/TEST*.xml`

PR #919, opened 2021-03-26
refactor reconcile loop

### LINKED ISSUE #884: Refactoring Service Binding Operator Reconcile Loop
#### Problem Statement

Current call graph for the reconcile function (look for entry point: **Reconcile** function)

sbo-reconcile-loop

Created using [go-callvis](https://github.com/ofabry/go-callvis) project.

### Terminology

*   **Binding Definition**- a (key, source) pair describing the location of a data item that is going to be harvested from the service data plane and contribute to binding data set. The definition can encoded as:
    *   OLM descriptor in ClusterServiceVersion (CSV)
    *   Resource annotation on service CRD and CR
*   **Binding Data** - a set of (name, value) pairs that are projected into application and accessible 
*   **Binding Data Projection** - binding data injection and transformation method. It defines how the binding data are accessible to application: 
    *   As environment variables (**binding name = env var name**, **binding value= env var value**) 
    *   As files (**binding name=filename**, **binding value=file content**)
*   **Intermediate Binding Secret** - prior performing data projection, binding data are stored in a k8s secret that can be injected into application either as
    *   A set of environment variables, or
    *   A set of files through a volume mount

#### Proposed Solution

Restructure the reconcile process as a pipeline, moving the incoming Service Binding instance through the following stages:

*   Gather **binding definitions**
*   Extract **binding data** from services and/or referred resources using **binding data**
*   Generate **synthetic binding data**, whose definitions are part of Service Binding
*   Adjust **binding item names** if required through submitted Service Binding
*   Create **intermediate binding secret** from the gathered binding data
*   If requested, project **binding data** as a set of environment variables in application container
*   If requested, project **binding data** as a set of files mounted inside application container
*   Persist changes in cluster

Refactoring Service Binding Operator Reconcile Loop

The pipeline implements a form of  [Intercepting Filter](https://www.oracle.com/java/technologies/intercepting-filter.html) pattern to give a developer full control over how a change on Service Binding is handled and how the stages in the pipeline interact with each other:

*   A stage can decide if processing stops or the context is passed down to the next stage
*   Stages communicates with each other only via reading/writing data to the context
*   Communication with cluster is abstracted via context

Required components and interfaces:

*   Context provider
*   Pipeline for driving the flow between stages
*   Stage accepting and processing context
*   Error handler that translates received errors from pipeline into proper reconcile actions

#### Benefits

*   Separation of concern - each stage does only one thing and does it good
*   Improved code readability/maintainability
*   Easy unit testing/integration testing
*   Single communication point with cluster allow easy mocking of cluster API
*   Possible reconciliation of service bindings outside of operator scope - use it as library and allow tighter integration with other projects, e.g. odo
*   Opens up an opportunity for better logging
*   Opens up an opportunity for better error handling

### Risks

Low. Service Binding Operator is accompanied with a good set of acceptance tests, covering all advertised use cases. Hence, we can perform refactoring with a high confidence.

pedjak:
Reconcilition process is restructured, moving the incoming Service Binding instance through the following stages:

* Gather binding definitions
* Collect binding data
* If detect binding resources is true, collect binding data from owned resources like secrets, configmaps, routes
* Generate additional binding items, if `mappings` block is provided
* Tweak binding names if required
* If requested inject binding secret ref to the specified location inside application resource
* If requested, project bindings as files inside application containers
* If requested, project bindings as env vars inside application container

Service bindings are pushed from the reconcilation loop through the pipeline, but a good side-effect of
this refactoring is that the pipeline can be used in a standalone mode, i.e. as a library. Hence, the operator reconcile
logic can be integrated in other tools, so that service binding can be performed even without having the operator available in the cluster.

A pipeline instance can be obtained as follows:

```
var client dynamic.Interface

pipeline = builder.DefaultBuilder.WithContextProvider(context.Provider(client, context.ResourceLookup(mgr.GetRESTMapper()))).Build()

```

Additional improvements:
* removed dead code
* better unit testability through separation of concerns
* unit test rewritten using Ginko and gomock library

Fixes #884


pedjak:
Call graph in `controllers` package:

sbo-controllers

and in `pipeline` package:

sbo-pipeline

wtrocki:
@pedjak Amazing work! I have picked up your changes and going to write minimal cli/tool to test the SDK usability/path. Leaving Operator test to the core team.

api/v1alpha1/conditions.go
```diff
@@ -0,0 +1,89 @@
+package v1alpha1
+
+import v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+type conditionsBuilder struct {
+	t       string
+	Status  v1.ConditionStatus
+	reason  string
```
Avni-Sharma, line 8: Any particular reason why reason is non exported whereas Status and Message are. Also, can we rename t to type/Type itself?

pkg/reconcile/pipeline/context/application_test.go
```diff
@@ -0,0 +1 @@
+package context
```
Avni-Sharma, line 1: This almost seems to be empty. Any tests that need t be added?

pkg/reconcile/pipeline/api.go
```diff
@@ -0,0 +1,175 @@
...
+
+// Application to be bound to service(s)
+type Application interface {
+
+	// Application resource
+	Resource() *unstructured.Unstructured
+
+	// dot-separated path inside the application resource locating container resources
+	// the returns value follows foo.bar.bla convention
+	// it cannot be empty
```
Avni-Sharma, line 69: It defaults to a value but it is optional (Reference https://github.com/redhat-developer/service-binding-operator/blob/master/api/v1alpha1/servicebinding_types.go#L185)

pkg/reconcile/pipeline/api.go
```diff
@@ -0,0 +1,175 @@
+package pipeline
+
+import (
+	"fmt"
+	olmalpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
```
Avni-Sharma, line 5: Let's keep this consistent as olmv1alpha1 https://github.com/redhat-developer/service-binding-operator/pull/919/files#diff-031de41097d908cf1f0f4728e2564fa6f8bc797433c9718a2bef50507012fdcfR6

At some places we have it as olmalpha1

sbose78:
Approach looks good, Predrag!

vendor/github.com/golang/mock/CONTRIBUTORS
```diff
@@ -0,0 +1,37 @@
+# This is the official list of people who can contribute (and typically
```
wtrocki, line 1: Out of curiosity: any reason to go with censoring Vs go modules?

controllers/servicebinding_controller.go
```diff
@@ -60,7 +62,45 @@ type ServiceBindingReconciler struct {
...
+	}
+	if serviceBinding.MaybeAddFinalizer() {
+		if err = r.Update(ctx, serviceBinding); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+	log.Info("Reconciling", "sb", serviceBinding)
+	retry, err := r.pipeline.Process(serviceBinding)
+	if !retry && err == nil {
+		if !serviceBinding.DeletionTimestamp.IsZero() {
```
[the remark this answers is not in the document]
pedjak: this is how kubernets API notifies us when a resource is about to be deleted, so that some housekeeping can be performed.

vendor/github.com/golang/mock/CONTRIBUTORS
pedjak: ah, that is a separate debate. Initially vendoring made our productization effort easier, but we could think about dropping it, but I would not block this PR with that decision.

pkg/reconcile/pipeline/api.go
pedjak: The contract on `pipeline.Application` is that it cannot be empty. This contact is different from the one you referred on ServiceBinding.

pmacik:
This PR has a positive impact on performance in a loaded cluster (~3000 users each with 3 namespaces and each binding a backing service (`Route`) and an application (`Deployment`) via `ServiceBinding`).

The memory usage dropped from up to ~110MiB to up to ~38 MiB (almost 3x less memory usage) and the overall CPU usage dropped from up to ~0.030 vCPU to up to ~0.024 vCPU (~20% less CPU usage).

While in terms of time needed to perform a binding (from when the binding was picked up by the operator to the time when it finished the binding - a.k.a `Reconciled to Ready` time) the average dropped from 498 ms to 391 ms (which is ~27% faster).

This PR:
image
as opposed to `v0.6.0`:
image

kadel:
I've tried to use this in a standalone mode. I've created a pipeline similar to an example in the description and then called `Process`  with the following ServiceBinding.

```json
{
  "kind": "ServiceBinding",
  "apiVersion": "binding.operators.coreos.com/v1alpha1",
  "metadata": {
   "name": "mynode-database-mydb",
   "namespace": "test",
   "creationTimestamp": null,
   "ownerReferences": [
    {
     "apiVersion": "apps/v1",
     "kind": "Deployment",
     "name": "mynode",
     "uid": "46d9ba4c-adb4-48c2-a567-6444bb114df5"
    }
   ]
  },
  "spec": {
   "services": [
    {
     "group": "postgresql.dev4devs.com",
     "version": "v1alpha1",
     "kind": "Database",
     "name": "mydb",
     "namespace": "test"
    }
   ],
   "application": {
    "group": "apps",
    "version": "v1",
    "resource": "deployments",
    "name": "mynode"
   },
   "detectBindingResources": true,
   "bindAsFiles": false
  },
  "status": {
   "secret": ""
  }
 }
```
But I always get an error: `Secret "mynode-database-mydb-8626f7de" is invalid: metadata.ownerReferences.uid: Invalid value: "": uid must not be empty`

Haven't got a chance to dig deeper into the code, but it looks like it is trying to create a `Secret` with an owner reference set to `ServiceBinding` that I passed as an argument. But the `ServiceBinding` doesn't exist one the cluster.

I thought that this is going to work without creating `ServiceBinding` CRD on the cluster. 
Or it is also quite possible that I'm missing something.

pedjak:
> I thought that this is going to work without creating ServiceBinding CR/D on the cluster. This won't help our use-case if it still requires creating CRDs :-(
> Or it is also quite possible that I'm missing something.

@kadel good catch. I have pushed a change that should fix the problem. Please let me know.

PR #949, opened 2021-04-13
regression fix: when binding as files use 'none' naming strategy by default

### LINKED ISSUE #945: Regression/Breaking changes in 0.7.0
## Different 

It looks like sbo binds files by default in openshift console.

0.6.0 release
``` 
sh-4.2$ ls /bindings/nodejs-sample-d-ajay-test-akc/
clientsecret user password type clientid
```
After 0.7.0
```
 sh-4.2$ ls /bindings/nodejs-sample-d-ajay-test-akc/
KAFKACONNECTION_BOOTSTRAPSERVERS  KAFKACONNECTION_CLIENTSECRET  KAFKACONNECTION_PROVIDER       KAFKACONNECTION_SECURITYPROTOCOL  KAFKACONNECTION_USER
KAFKACONNECTION_CLIENTID          KAFKACONNECTION_PASSWORD      KAFKACONNECTION_SASLMECHANISM  KAFKACONNECTION_TYPE
```
0.7.0 is not aligned with how spec does it.

pedjak:
It got broken in 0.7.0 - default situation was not properly detected.

Fixes #945


wtrocki (approved):
Defaults work great.
Tests cover all possible cases! Amazing work

config/manager/kustomization.yaml
```diff
@@ -11,6 +11,6 @@ configMapGenerator:
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
-- name: controller
-  newName: sbo
-  newTag: latest
+- digest: sha256:c46fc750d0234f1624607e321053f7af94034e649e820a5d4e784a1117787d70
+  name: controller
+  newName: 192.168.49.2:5000/sbo
```
pmacik, line 16: @pedjak I believe this is a accidentally part of the PR. The `newName` is likely for you local minikube and the digest will also change with every change...

api/v1alpha1/servicebinding_types.go
```diff
@@ -96,7 +96,8 @@ type ServiceBindingSpec struct {
 	// BindAsFiles makes available the binding values as files in the application's container
 	// See MountPath attribute description for more details.
 	// +optional
-	BindAsFiles *bool `json:"bindAsFiles,omitempty"`
+	// +kubebuilder:default:=true
```
pmacik, line 99: @pedjak This is making the default value for `bindAsFiles` to `true`, this should be also reflected in the [UserGuide](https://github.com/redhat-developer/service-binding-operator/blob/master/docs/User_Guide.md#accessing-the-binding-data-from-the-application) where it is written that the default is `false`

PR #959, opened 2021-04-30
Check if unready binding can be removed

pedjak:
* added missing acceptance tests
* in the reconcile loop, add finalizer only if deletion timestamp is unset
* unbind handler: stop processing and do not retry in case of any error when reading the application resource


test/acceptance/features/unbindAppToService.feature
```diff
@@ -154,4 +154,42 @@ Feature: Unbind an application from a service
...
+                    resource: deployments
+                services:
+                -   group: stable.example.com
+                    version: v1
+                    kind: Backend
+                    name: example-backend
+            """
+        * jq ".status.conditions[] | select(.type=="Ready").status" of Service Binding "unready-binding" should be changed to "False"
+        When Service binding "unready-binding" is deleted
+
```
DhritiShikhar, line 194: line:194 -> unnecessary whitespace

DhritiShikhar:
LGTM

Issue #808, opened 2020-12-17
Service binding is unable to process annotation for secretRef from containers path of deployment


pratikjagrut:
I'm trying to collect secret stated in the annotation of deployment.

*Following is the deployment I'm using*

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: mysql-db
  annotations:
    service.binding: path={.spec.template.spec.containers[0].envFrom[0].secretRef.name},objectType=Secret
  labels:
    app: wordpress
spec:
  selector:
    matchLabels:
      app: wordpress
      tier: mysql
  template:
    metadata:
      labels:
        app: wordpress
        tier: mysql
    spec:
      containers:
      - image: mysql:5.7
        name: mysql
        envFrom:
        - secretRef:
            name: db-secret
        ports:
        - containerPort: 3306
          name: mysql
```
*Secret*
```yaml
apiVersion: v1
kind: Secret
metadata:
    name: db-secret
stringData:
    MYSQL_DATABASE: "wordpress"
    MYSQL_ROOT_PASSWORD: "somewordpress"
    MYSQL_USER: "wordpress"
    MYSQL_PASSWORD: "wordpress"
```

Service Binding
```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: ServiceBinding
metadata:
  name: service-binding-multiple-services
spec:
  services:
  - group: apps
    version: v1
    kind: Deployment
    name: mysql-db
  detectBindingResources: true
```

**Expected Output:**
Secret `service-binding-multiple-services` with data from `db-secret`.

**Current output:**
Empty secret `service-binding-multiple-services`
```shell
Name:         service-binding-multiple-services
Namespace:    default
Labels:       <none>
Annotations:  <none>

Type:  Opaque

Data
====
```

pratikjagrut:
Logs:
```shell
2020-12-17T15:03:51.345+0530    INFO    reconciler      Reconciling ServiceBinding...   {"Request.Namespace": "default", "Request.Name": "service-binding-multiple-services"}
2020-12-17T15:03:51.347+0530    DEBUG   reconciler      Found service binding request to inspect        {"Request.Namespace": "default", "Request.Name": "service-binding-multiple-services", "ServiceBinding.Name": "service-binding-multiple-services"}
[...]

PR #964, opened 2021-05-14
Support for selection of array element in paths

### LINKED ISSUE #963: Support for array element selection in jsonpath for paths in annotations

Minikube

master

- Create a statefulset:

```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: db1
  annotations:
    service.binding/address: path={.spec.serviceName}
    service.binding/postgre_user: path={.spec.template.spec.containers[0].env[?(@.name=="POSTGRESQL_USER")].value}
spec:
  selector:
    matchLabels:
      app: db1
  serviceName: db1
  replicas: 1
  template:
    metadata:
      labels:
        app: db1
    spec:
      containers:
      - name: db1
        image: centos/postgresql-96-centos7
        env:
        - name: POSTGRESQL_USER
          value: user1
        - name: POSTGRESQL_PASSWORD
          value: k33p5ecret
        - name: POSTGRESQL_DATABASE
          value: mydb
        ports:
        - containerPort: 5432
          name: pg
        volumeMounts:
        - name: data
          mountPath: /var/lib/pgsql/data
      volumes:
      - name: data
        emptyDir: {}
```
- Create a servicebinding:

```
kind: ServiceBinding

metadata:
  name: binding-request
  namespace: prj1

spec:
  application:
    group: apps
    version: v1
    resource: deployments
    name: api

  services:
    - group: apps
      version: v1
      kind: StatefulSet
      name: db1
      namespace: prj1
```

- The operator should be able to find the value `user1` for the path `.spec.template.spec.containers[0].env[?(@.name=="POSTGRESQL_USER")].value` as the path is valid jsonpath

The annotations are ignored as the paths are not recognized as valid jsonpath

### Additional Information (Screenshots, etc)

Fix proposal at #964

feloy:
### Changes

Calls to `unstructured.NestedFieldCopy` replaced by usage of [JSONPath) parser from client-go library](https://github.com/kubernetes/client-go/tree/master/util/jsonpath).

Refactored `model` and builders to keep jsonpath in original form, without spliting into steps
    
 * This simplified the annotation parsing logic
 * Since we now support full JSON path, we cannot now detect selected field name in all possible cases.
    Hence using "service.binding" annotation key alone is allowed only when `objectType` is set to `Secret` or `ConfigMap`
    In all other cases, "service.binding/xxx" convention must be used. A number of unit tests adjusted accordingly
 * reenabled the acceptance tests that requires `path` to be a regular JSONPath expression
      (the provided expression uses subscript operator to locate the field containing the secret name)

Fixes #963. #808

### DIFF
--- a/pkg/binding/definition.go
+++ b/pkg/binding/definition.go
@@ -38,42 +38,44 @@ const (
 //go:generate mockgen -destination=mocks/mocks.go -package=mocks . Definition,Value
 
 type Definition interface {
-	GetPath() []string
 	Apply(u *unstructured.Unstructured) (Value, error)
+	GetPath() string
 }
 
 type DefinitionBuilder interface {
 	Build() (Definition, error)
 }
 
+type definition struct {
+	path string
+}
+
+func (d *definition) GetPath() string {
+	return d.path
+}
+
 type stringDefinition struct {
 	outputName string
-	path       []string
+	definition
 }
 
 var _ Definition = (*stringDefinition)(nil)
 
-func (d *stringDefinition) getOutputName() string {
-	outputName := d.outputName
-	if len(outputName) == 0 {
-		outputName = d.path[len(d.path)-1]
-	}
-	return outputName
-}
-
-func (d *stringDefinition) GetPath() []string { return d.path[0 : len(d.path)-1] }
-
 func (d *stringDefinition) Apply(u *unstructured.Unstructured) (Value, error) {
-	val, ok, err := unstructured.NestedFieldCopy(u.Object, d.path...)
+	if d.outputName == "" {
+		return nil, fmt.Errorf("cannot use generic service.binding annotation for string elements, need to specify binding key like service.binding/foo")
+	}
+	val, err := getValuesByJSONPath(u.Object, d.path)
 	if err != nil {
 		return nil, err
 	}
-	if !ok {
-		return nil, errors.New("not found")
+
+	if len(val) != 1 {
+		return nil, fmt.Errorf("only one value should be returned for %v but we got %v", d.path, val)
 	}
 
 	m := map[string]interface{}{
-		d.getOutputName(): fmt.Sprintf("%v", val),
+		d.outputName: val[0].Interface(),
 	}
 
 	return &value{v: m}, nil
@@ -83,26 +85,26 @@ type stringFromDataFieldDefinition struct {
 	secretConfigMapReader *secretConfigMapReader
 	objectType            objectType
 	outputName            string
-	path                  []string
-	sourceKey             string
+	definition
+	sourceKey string
 }
 
 var _ Definition = (*stringFromDataFieldDefinition)(nil)
-
-func (d *stringFromDataFieldDefinition) GetPath() []string { return d.path }
 
 func (d *stringFromDataFieldDefinition) Apply(u *unstructured.Unstructured) (Value, error) {
 	if d.secretConfigMapReader == nil {
 		return nil, errors.New("kubeClient required for this functionality")
 	}
 
-	resourceName, ok, err := unstructured.NestedString(u.Object, d.path...)
+	res, err := getValuesByJSONPath(u.Object, d.path)
 	if err != nil {
 		return nil, err
 	}
-	if !ok {
-		return nil, errors.New("not found")
+
+	if len(res) != 1 {
+		return nil, fmt.Errorf("only one value should be returned for %v but we got %v", d.path, res)
 	}
+	resourceName := res[0].String()
 
 	var otherObj *unstructured.Unstructured
 	if d.objectType == secretObjectType {
@@ -140,25 +142,24 @@ type mapFromDataFieldDefinition struct {
 	objectType            objectType
 	outputName            string
 	sourceValue           string
-	path                  []string
+	definition
 }
 
 var _ Definition = (*mapFromDataFieldDefinition)(nil)
-
-func (d *mapFromDataFieldDefinition) GetPath() []string { return d.path }
 
 func (d *mapFromDataFieldDefinition) Apply(u *unstructured.Unstructured) (Value, error) {
 	if d.secretConfigMapReader == nil {
 		return nil, errors.New("kubeClient required for this functionality")
 	}
 
-	resourceName, ok, err := unstructured.NestedString(u.Object, d.path...)
+	res, err := getValuesByJSONPath(u.Object, d.path)
 	if err != nil {
 		return nil, err
 	}
-	if !ok {
-		return nil, errors.New("not found")
+	if len(res) != 1 {
+		return nil, fmt.Errorf("only one value should be returned for %v but we got %v", d.path, res)
 	}
+	resourceName := res[0].Elem().String()
 
 	var otherObj *unstructured.Unstructured
 	if d.objectType == secretObjectType {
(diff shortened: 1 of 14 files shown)


pedjak:
@feloy many thanks for the PR - for some time we noticed that `path` should be able to handle any jsonpath complaint expression. Your PR is a step in the right direction.  Would you mind to extend this PR to help us to make it fully compliant and replace our custom-made logic with full JSONpath parser? There are a few good parser available, one is even a part of `client-go` library we use in the project:

* https://github.com/kubernetes/client-go/blob/master/util/jsonpath/jsonpath_test.go
* or https://github.com/PaesslerAG/jsonpath

In order to replace  the current custom logic, we need to perform changes in the following places:

* https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/binding/definition.go#L67
* https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/binding/definition.go#L99
* https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/binding/definition.go#L155
* https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/binding/definition.go#L218
* https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/binding/definition.go#L249
* https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/binding/definition.go#L281

We just need to invoke the parser to get to the right element instead of calling `unstructured.NestedFieldCopy`

Looking forward to your contribution.

feloy:
> In order to replace the current custom logic, we need to perform changes in the following places:
> 
* [ ] https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/binding/definition.go#L67
* [ ] https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/binding/definition.go#L99
* [ ] https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/binding/definition.go#L155
* [ ] https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/binding/definition.go#L218
* [ ] https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/binding/definition.go#L249
* [ ] https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/binding/definition.go#L281

pkg/binding/nested/field_copy.go
```diff
@@ -0,0 +1,23 @@
+package nested
```
pedjak, line 1: Let's keep this single function under `binding` package - this opens up possibility to make it private.

pkg/binding/nested/field_copy.go
```diff
@@ -0,0 +1,23 @@
...
+
+func GetNested(obj map[string]interface{}, path_segments ...string) (string, bool, error) {
+	path := "{." + strings.Join(path_segments, ".") + "}"
+	j := jsonpath.New("")
+	err := j.Parse(path)
+	if err != nil {
+		return "", false, err
+	}
+	buf := new(bytes.Buffer)
+	err = j.Execute(buf, obj)
```
pedjak, line 18: using j.FindResults(obj) is more appropriate, because jsonpath can return multiple results that could be then converted into individual bindings by the caller. Using `Execute` always return just single result.

Having said that, this function should return a pair of `[][]reflect.Value, error` to reflect that.

pkg/binding/nested/field_copy.go
```diff
@@ -0,0 +1,23 @@
+package nested
+
+import (
+	"bytes"
+	"strings"
+
+	"k8s.io/client-go/util/jsonpath"
+)
+
+func GetNested(obj map[string]interface{}, path_segments ...string) (string, bool, error) {
```
pedjak, line 10: Let's rename this function to reflect better what it does: how about `getValuesByJSONPath`?

pkg/reconcile/pipeline/context/impl.go
```diff
@@ -92,7 +92,7 @@ func (i *impl) Mappings() map[string]string {
 func (i *impl) Services() ([]pipeline.Service, error) {
 	if i.services == nil {
 		serviceRefs := i.serviceBinding.Spec.Services
-		for idx := 0; idx<len(serviceRefs); idx++ {
+		for idx := 0; idx < len(serviceRefs); idx++ {
```
pedjak, line 95: whitespace changes unrelated to this PR, please revert.

pkg/binding/model.go
```diff
@@ -41,19 +41,32 @@ func (m *model) hasDataField() bool {
 }
 
 func newModel(annotationValue string) (*model, error) {
+
+	reObf := regexp.MustCompile(`{.*?}`)
+	obfuscatedValue := reObf.ReplaceAllStringFunc(annotationValue, func(s string) string {
+		return "{" + strings.Repeat("_", len(s)-2) + "}"
+	})
+
```
pedjak, line 49: why the existing `newModel` logic does not work anymore? the jsonpath was specified in `path` in format pretty much suitable for the parser. The value found under `path` key can be moved to`model` as is, there is no need to split it anymore.

pedjak (changes requested):
and we need at least one acceptance test demonstrating the use case.

pkg/binding/model.go
feloy: `re := regexp.MustCompile("[=,]")` splits the annotation value `key1={path},key2=value2,...`

Now that `path` can contain `=` signs, we have to update the way the annotation is splitted.

pkg/binding/definition.go
```diff
@@ -215,14 +215,20 @@ var _ Definition = (*stringOfMapDefinition)(nil)
...
 	if !ok {
 		return nil, errors.New("not found")
 	}
 
+	var val interface{}
+	err = json.Unmarshal([]byte(strval), &val)
+	if err != nil {
+		return nil, err
+	}
+
```
pedjak, line 231: Instead of repeating this block 3 times in the code, how about that `getValuesByJSONPath` returns `[]interface` directly? Then we could make some assertions based on the expected outcome (e..g if just one string should be selected, then we should expect just single entry, etc.)?

pkg/binding/field_copy.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"bytes"
+	"strings"
+
+	"k8s.io/client-go/util/jsonpath"
+)
+
+func getValuesByJSONPath(obj map[string]interface{}, path_segments ...string) (string, bool, error) {
+	path := "{." + strings.Join(path_segments, ".") + "}"
```
pedjak, line 11: if we need path onward and do not use path_segements anywhere else, then let's pass to the function as argument path directly. This would also remove the need to split the path before calling this method.

pkg/binding/field_copy.go
```diff
@@ -0,0 +1,23 @@
+package binding
+
+import (
+	"bytes"
+	"strings"
+
+	"k8s.io/client-go/util/jsonpath"
+)
+
+func getValuesByJSONPath(obj map[string]interface{}, path_segments ...string) (string, bool, error) {
```
pedjak, line 10: this function now returns bool = `false` only when error != nil. Hence, we can drop returning `bool` completely. As mentioned above, the function could return `[]interfaces{}` or something more appropriate, so that we we do not need to deserialize again at the caller.

The library offers `FindResults` function that can be used for that purpose.

pkg/binding/model.go
pedjak: > Now that path can contain = signs, we have to update the way the annotation is splitted.

True. Does the regex should be then `{.+}` or even to extract the path directly: `path={.+},?` If later, we can use `FindStringIndex` function to get `path={xxx}` and extract `{xxx}` directly, and then split the rest of keys as before?

pkg/binding/model.go
```diff
@@ -127,7 +140,19 @@ func newModel(annotationValue string) (*model, error) {
...
+	obfPathParts := strings.Split(obfuscatedPath, ".")
+
+	pathParts := make([]string, len(obfPathParts))
+	start = 0
+	for i, part := range obfPathParts {
+		pathParts[i] = path[start : start+len(part)]
+		start += len(part) + 1
+	}
```
pedjak, line 155: given that json parser takes the full path as the argument, we do not need to split paths anymore, we should modify `model.path` field to be of type `string`.

pkg/binding/field_copy_test.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+type splitPathResult struct {
+	StartPath        []string
+	EndPath          []string
+	IsSlice          bool
+	SliceIndexString string
+}
+
+func TestNestedFieldCopy(t *testing.T) {
```
pedjak, line 16: the function under test is getValueByJSONPath, so let's reflect that in the test name.

pkg/binding/field_copy_test.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+		{
+			Path:    []string{"spec", "template", "spec", "containers[?(@.name==\"db1\")]", "env[?(@.name==\"POSTGRESQL_USER\")]", "value"},
+			Want:    "user1",
+			WantOK:  true,
+			WantErr: false,
+		},
+	}
+
+	for _, test := range tests {
```
pedjak, line 94: let's make true table tests by using `t.Run` functions

pedjak (changes requested):

pkg/binding/model.go
feloy: Yes, I explored this change, and that needs a lot of changes on the package. I will need some more time to work on this

pedjak:
@feloy I took freedom to refactor the model so that the patch is kept as the whole, this makes the annotation value parsing a way easier. See https://github.com/redhat-developer/service-binding-operator/pull/964/commits/94466fa1de993e79db27ad2529bbded076aaae64 for the details. At the same I needed to rebase PR to the latest master due to the merge conflicts.

pedjak (approved):

pedjak:
@baijum @isutton I have coauthored this PR, hence need others to take a look.

pkg/binding/jsonpath.go
```diff
@@ -0,0 +1,28 @@
...
+	err := j.Parse(path)
+	if err != nil {
+		return nil, err
+	}
+	result, err := j.FindResults(obj)
+	if err != nil {
+		return nil, err
+	}
+	if len(result) > 1 {
+		return nil, fmt.Errorf("Expecting at most a list of values, but not list of lists: %v", result)
```
baijum, line 25: This error message is a bit abmigous.
How about: `more than one item found in the result: %v`

pkg/binding/definition.go
```diff
@@ -38,42 +37,44 @@ const (
...
+	}
+	val, err := getValuesByJSONPath(u.Object, d.path)
 	if err != nil {
 		return nil, err
 	}
-	if !ok {
-		return nil, errors.New("not found")
+
+	if len(val) != 1 {
+		return nil, fmt.Errorf("Only one value should be returned for %v but we got %v", d.path, val)
```
baijum, line 73: Error strings should not be capitalized.
Ref. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/binding/definition.go
```diff
@@ -83,26 +84,26 @@ type stringFromDataFieldDefinition struct {
...
-	resourceName, ok, err := unstructured.NestedString(u.Object, d.path...)
+	res, err := getValuesByJSONPath(u.Object, d.path)
 	if err != nil {
 		return nil, err
 	}
-	if !ok {
-		return nil, errors.New("not found")
+
+	if len(res) != 1 {
+		return nil, fmt.Errorf("Only one value should be returned for %v but we got %v", d.path, res)
```
baijum, line 104: Error strings should not be capitalized.
Ref. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/binding/definition.go
```diff
@@ -140,25 +141,24 @@ type mapFromDataFieldDefinition struct {
...
 
-	resourceName, ok, err := unstructured.NestedString(u.Object, d.path...)
+	res, err := getValuesByJSONPath(u.Object, d.path)
 	if err != nil {
 		return nil, err
 	}
-	if !ok {
-		return nil, errors.New("not found")
+	if len(res) != 1 {
+		return nil, fmt.Errorf("Only one value should be returned for %v but we got %v", d.path, res)
```
baijum, line 159: Error strings should not be capitalized.
Ref. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/binding/definition.go
```diff
@@ -207,99 +207,115 @@ func (d *mapFromDataFieldDefinition) Apply(u *unstructured.Unstructured) (Value,
...
-func (d *stringOfMapDefinition) GetPath() []string { return d.path }
-
 func (d *stringOfMapDefinition) Apply(u *unstructured.Unstructured) (Value, error) {
-	val, ok, err := unstructured.NestedFieldNoCopy(u.Object, d.path...)
+	val, err := getValuesByJSONPath(u.Object, d.path)
 	if err != nil {
 		return nil, err
 	}
+	if len(val) != 1 {
+		return nil, fmt.Errorf("Only one value should be returned for %v but we got %v", d.path, val)
```
baijum, line 221: Error strings should not be capitalized.
Ref. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

baijum (approved):
There are few minor comments

feloy:
> There are few minor comments

Thanks for the review. I made the changes

PR #971, opened 2021-06-04
Fix support for label selectors

### LINKED ISSUE #965: application.name field became mandatory/redundant in release 0.7.x
OCP 4.6.16 on premises

0.7.0
0.7.1

  1. Install the SBO operator (see attached `subscription.yaml` in attached archive)
  2. Deploy a sample app, using [these instructions](https://github.com/redhat-developer/service-binding-operator/tree/master/examples/nodejs_postgresql#import-an-application)
  3. Apply the binding request (see `binding-request-noname.yaml` in attached archive) , which uses label selectors to match the application:
      ```
      oc project service-binding-demo
      oc apply -f binding-request-noname.yaml
      ```
  4. Inspect logs for SBO pod (see `logs` in attached archive)
      ```
     {"level":"error","ts":1621375914.120951,"logger":"controller","msg":"Reconciler error","reconcilerGroup":"binding.operators.coreos.com","reconcilerKind":"ServiceBinding","controller":"servicebinding","name":"binding-request-noname","namespace":"service-binding-demo","error":"name is required"
      ```
  5. Now "fix" the `ServiceBinding` by adding the redundant `name` field:
      ```
      oc apply -f binding-request-with-name.yaml
      ```
  6. Inspect logs for SBO pod (see `logs` in attached archive) and see a successful message:
      ```
    {"level":"info","ts":1621376188.5513623,
     "logger":"controllers.ServiceBinding",
     "msg":"Done","serviceBinding":"service-binding-demo/binding-request-name"
      ```

Service binding should be processed without issue and mapped to the matching application.

Operator logs surface an error in the binding request, indicating that the `.spec.application.name` must be filled out, even though `.spec.application.labelSelector` is specified.

### Service Binding Operator Logs
See `logs` in attached archive.

### Additional Information (Screenshots, etc)
This same `binding-request-noname.yaml` file worked correctly ysing the operator version 0.6.0.
I can see the secret created by the operator:

```
 oc get Secret binding-request-noname-6f93c924eb98fbd528c74b610b632229352734c1 -n service-binding-demo -o jsonpath={.data}
map[env.json:ew...==]
```

And I can see it mounted to the application
```
oc get pod nodejs-rest-http-crud-7dfd6855db-r5s5s -n service-binding-demo -o jsonpath="{.spec.containers[].volumeMounts}" 
[map[name:binding-request-noname mountPath:/config] ...
```

[Archive.zip](https://github.com/redhat-developer/service-binding-operator/files/6504896/Archive.zip)

baijum:
- Fix `.spec.application.name` field becoming mandatory
- If more than one application is selected through label selectors, all will be considered for binding

Fixes #965

- [x] acceptance test demonstrating how to bind 2 applications having the same label.
- [x] Remove `InvalidApplicationReference` reason and corresponding code and tests
  - [x] Added issue #974
- [x] Wrong assertion for the test: `should return an error if application list returns error`

### DIFF
--- a/pkg/reconcile/pipeline/context/impl.go
+++ b/pkg/reconcile/pipeline/context/impl.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"crypto/sha1"
 	"encoding/hex"
+	"sort"
+
 	"github.com/redhat-developer/service-binding-operator/api/v1alpha1"
 	"github.com/redhat-developer/service-binding-operator/pkg/client/kubernetes"
 	"github.com/redhat-developer/service-binding-operator/pkg/converter"
@@ -13,9 +15,9 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
-	"sort"
 )
 
 var _ pipeline.Context = &impl{}
@@ -92,7 +94,7 @@ func (i *impl) Mappings() map[string]string {
 func (i *impl) Services() ([]pipeline.Service, error) {
 	if i.services == nil {
 		serviceRefs := i.serviceBinding.Spec.Services
-		for idx := 0; idx<len(serviceRefs); idx++ {
+		for idx := 0; idx < len(serviceRefs); idx++ {
 			serviceRef := serviceRefs[idx]
 			gvr, err := i.typeLookup.ResourceForReferable(&serviceRef)
 			if err != nil {
@@ -109,7 +111,7 @@ func (i *impl) Services() ([]pipeline.Service, error) {
 		}
 	}
 	services := make([]pipeline.Service, len(i.services))
-	for idx := 0; idx<len(i.services); idx++ {
+	for idx := 0; idx < len(i.services); idx++ {
 		services[idx] = i.services[idx]
 	}
 	return services, nil
@@ -122,11 +124,33 @@ func (i *impl) Applications() ([]pipeline.Application, error) {
 			if err != nil {
 				return nil, err
 			}
-			u, err := i.client.Resource(*gvr).Namespace(i.serviceBinding.Namespace).Get(context.Background(), ref.Name, metav1.GetOptions{})
-			if err != nil {
-				return nil, err
+			if i.serviceBinding.Spec.Application.Name != "" {
+				u, err := i.client.Resource(*gvr).Namespace(i.serviceBinding.Namespace).Get(context.Background(), ref.Name, metav1.GetOptions{})
+				if err != nil {
+					return nil, err
+				}
+				i.applications = append(i.applications, &application{gvr: gvr, persistedResource: u, bindingPath: i.serviceBinding.Spec.Application.BindingPath})
 			}
-			i.applications = append(i.applications, &application{gvr: gvr, persistedResource: u, bindingPath: i.serviceBinding.Spec.Application.BindingPath})
+			if i.serviceBinding.Spec.Application.LabelSelector != nil && i.serviceBinding.Spec.Application.LabelSelector.MatchLabels != nil {
+				matchLabels := i.serviceBinding.Spec.Application.LabelSelector.MatchLabels
+				opts := metav1.ListOptions{
+					LabelSelector: labels.Set(matchLabels).String(),
+				}
+
+				emptyResult := make([]pipeline.Application, 0)
+				objList, err := i.client.Resource(*gvr).Namespace(i.serviceBinding.Namespace).List(context.Background(), opts)
+				if err != nil {
+					return emptyResult, err
+				}
+
+				if len(objList.Items) == 0 {
+					return emptyResult, nil
+				}
+
+				for index := range objList.Items {
+					i.applications = append(i.applications, &application{gvr: gvr, persistedResource: &(objList.Items[index]), bindingPath: i.serviceBinding.Spec.Application.BindingPath})
+				}
+			}
 		} else {
 			i.applications = make([]*application, 0)
 		}
--- a/pkg/reconcile/pipeline/context/impl_test.go
+++ b/pkg/reconcile/pipeline/context/impl_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	e "errors"
 	"fmt"
+
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
@@ -21,6 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/fake"
+	"k8s.io/client-go/testing"
 )
 
 var _ = Describe("Context", func() {
@@ -91,7 +93,146 @@ var _ = Describe("Context", func() {
 			Entry("no binding path specified", nil, defaultContainerPath),
 			Entry("binding path specified", &v1alpha1.BindingPath{ContainersPath: "foo.bar"}, "foo.bar"),
 		)
+		DescribeTable("should return slice of size 2 if 2 applications are specified through label seclector", func(bindingPath *v1alpha1.BindingPath, expectedContainerPath string) {
+			ls := &metav1.LabelSelector{
+				MatchLabels: map[string]string{"env": "prod"},
+			}
 
+			ref := v1alpha1.Application{
+				Ref: v1alpha1.Ref{
+					Group:   "app",
+					Version: "v1",
+					Kind:    "Foo",
+				},
+				LabelSelector: ls,
+				BindingPath:   bindingPath,
+			}
+
+			sb := v1alpha1.ServiceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "sb1",
+					Namespace: "ns1",
+				},
+				Spec: v1alpha1.ServiceBindingSpec{
+					Application: &ref,
+				},
+			}
+			gvr := &schema.GroupVersionResource{Group: "app", Version: "v1", Resource: "foos"}
+			typeLookup.EXPECT().ResourceForReferable(&ref).Return(gvr, nil)
+
+			u1 := &unstructured.Unstructured{}
+			u1.SetName("app1")
+			u1.SetNamespace(sb.Namespace)
+			u1.SetGroupVersionKind(schema.GroupVersionKind{Group: "app", Version: "v1", Kind: "Foo"})
+			u1.SetLabels(map[string]string{"env": "prod"})
+
+			u2 := &unstructured.Unstructured{}
+			u2.SetName("app2")
+			u2.SetNamespace(sb.Namespace)
+			u2.SetGroupVersionKind(schema.GroupVersionKind{Group: "app", Version: "v1", Kind: "Foo"})
+			u2.SetLabels(map[string]string{"env": "prod"})
+
+			client := fake.NewSimpleDynamicClient(runtime.NewScheme(), u1, u2)
+
+			ctx := &impl{client: client, serviceBinding: &sb, typeLookup: typeLookup}
+
+			applications, err := ctx.Applications()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(applications).To(HaveLen(2))
+
+			Expect(applications[0].Resource().GetName()).NotTo(Equal(applications[1].Resource().GetName()))
+			Expect(applications[0].Resource()).Should(BeElementOf(u1, u2))
+			Expect(applications[1].Resource()).Should(BeElementOf(u1, u2))
+			Expect(applications[0].ContainersPath()).To(Equal(expectedContainerPath))
+			Expect(applications[1].ContainersPath()).To(Equal(expectedContainerPath))
+		},
+			Entry("no binding path specified", nil, defaultContainerPath),
+			Entry("binding path specified", &v1alpha1.BindingPath{ContainersPath: "foo.bar"}, "foo.bar"),
+		)
+		DescribeTable("should return slice of size 0 if no application is matching through label seclector", func(bindingPath *v1alpha1.BindingPath, expectedContainerPath string) {
+			ls := &metav1.LabelSelector{
+				MatchLabels: map[string]string{"env": "prod"},
+			}
+
+			ref := v1alpha1.Application{
+				Ref: v1alpha1.Ref{
+					Group:   "app",
+					Version: "v1",
+					Kind:    "Foo",
+				},
+				LabelSelector: ls,
+				BindingPath:   bindingPath,
+			}
+
+			sb := v1alpha1.ServiceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "sb1",
+					Namespace: "ns1",
+				},
+				Spec: v1alpha1.ServiceBindingSpec{
+					Application: &ref,
+				},
+			}
+			gvr := &schema.GroupVersionResource{Group: "app", Version: "v1", Resource: "foos"}
+			typeLookup.EXPECT().ResourceForReferable(&ref).Return(gvr, nil)
+
+			u := &unstructured.Unstructured{}
+			u.SetName("app")
+			u.SetNamespace(sb.Namespace)
+			u.SetGroupVersionKind(schema.GroupVersionKind{Group: "app", Version: "v1", Kind: "Foo"})
+			u.SetLabels(map[string]string{"env": "stage"})
+			client := fake.NewSimpleDynamicClient(runtime.NewScheme(), u)
+
+			ctx := &impl{client: client, serviceBinding: &sb, typeLookup: typeLookup}
+
+			applications, err := ctx.Applications()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(applications).To(HaveLen(0))
+		},
+			Entry("no binding path specified", nil, defaultContainerPath),
+			Entry("binding path specified", &v1alpha1.BindingPath{ContainersPath: "foo.bar"}, "foo.bar"),
+		)
+
+		It("should return error if application list returns error", func() {
+			ls := &metav1.LabelSelector{
+				MatchLabels: map[string]string{"env": "prod"},
+			}
+
+			ref := v1alpha1.Application{
+				Ref: v1alpha1.Ref{
+					Group:   "app",
+					Version: "v1",
+					Kind:    "Foo",
+				},
+				LabelSelector: ls,
+			}
+
+			sb := v1alpha1.ServiceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "sb1",
+					Namespace: "ns1",
+				},
+				Spec: v1alpha1.ServiceBindingSpec{
+					Application: &ref,
+				},
+			}
+
+			gvr := &schema.GroupVersionResource{Group: "app", Version: "v1", Resource: "foos"}
+			typeLookup.EXPECT().ResourceForReferable(&ref).Return(gvr, nil)
+
+			client := fake.NewSimpleDynamicClient(runtime.NewScheme())
+			expectedError := "Error listing foo"
+			client.PrependReactor("list", "foos",
+				func(action testing.Action) (handled bool, ret runtime.Object, err error) {
+					return true, nil, e.New(expectedError)
+				})
+
+			ctx := &impl{client: client, serviceBinding: &sb, typeLookup: typeLookup}
+
+			_, err := ctx.Applications()
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(Equal(expectedError))
+		})
 		It("should return error if application is not found", func() {
 			ref := v1alpha1.Application{
 				Ref: v1alpha1.Ref{
--- a/test/acceptance/features/steps/generic_testapp.py
+++ b/test/acceptance/features/steps/generic_testapp.py
@@ -38,6 +38,9 @@ class GenericTestApp(App):
         polling2.poll(lambda: requests.get(url=f"http://{self.route_url}{file_path}"),
                       check_success=lambda r: r.status_code == 404, step=5, timeout=400, ignore_exceptions=(requests.exceptions.ConnectionError,))
 
+    def set_label(self, label):
+        self.openshift.set_label(self.name, label, self.namespace)
+
 
 @step(u'Generic test application "{application_name}" is running')
 @step(u'Generic test application "{application_name}" is running with binding root as "{bindingRoot}"')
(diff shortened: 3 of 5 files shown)


pkg/reconcile/pipeline/context/impl.go
```diff
@@ -118,15 +121,44 @@ func (i *impl) Services() ([]pipeline.Service, error) {
 func (i *impl) Applications() ([]pipeline.Application, error) {
 	if i.applications == nil {
 		if ref := i.serviceBinding.Spec.Application; ref != nil {
-			gvr, err := i.typeLookup.ResourceForReferable(ref)
-			if err != nil {
-				return nil, err
+			if i.serviceBinding.Spec.Application.Name != "" &&
+				i.serviceBinding.Spec.Application.LabelSelector != nil &&
+				i.serviceBinding.Spec.Application.LabelSelector.MatchLabels != nil {
+				return nil, errs.New("Both label selector and name for application should be used")
```
pedjak, line 127: you mean - "should not be used"?

pkg/reconcile/pipeline/context/impl.go
```diff
@@ -118,15 +121,44 @@ func (i *impl) Services() ([]pipeline.Service, error) {
...
+				}
+				u, err := i.client.Resource(*gvr).Namespace(i.serviceBinding.Namespace).Get(context.Background(), ref.Name, metav1.GetOptions{})
+				if err != nil {
+					return nil, err
+				}
+				i.applications = append(i.applications, &application{gvr: gvr, persistedResource: u, bindingPath: i.serviceBinding.Spec.Application.BindingPath})
+
+			}
+			if i.serviceBinding.Spec.Application.LabelSelector != nil && i.serviceBinding.Spec.Application.LabelSelector.MatchLabels != nil {
+				gvr, err := i.typeLookup.ResourceForReferable(ref)
```
pedjak, line 142: this lines repeats also in line 130 - so it could be moved before back to the original place.

pkg/reconcile/pipeline/context/impl.go
```diff
@@ -118,15 +121,44 @@ func (i *impl) Services() ([]pipeline.Service, error) {
...
+					LabelSelector: labels.Set(matchLabels).String(),
+				}
+
+				objList, err := i.client.Resource(*gvr).Namespace(i.serviceBinding.Namespace).List(context.Background(), opts)
+				if err != nil {
+					return nil, err
+				}
+
+				if len(objList.Items) == 0 {
+					return nil, errs.New("application not found")
```
pedjak, line 154: the contract of this method is to return empty array, not array if not applications are not found, see the doc: https://github.com/redhat-developer/service-binding-operator/blob/d680987ea0f2869557ada3f3dcc621981cb1d427/pkg/reconcile/pipeline/api.go#L102

pkg/reconcile/pipeline/context/impl_test.go
```diff
@@ -91,6 +92,51 @@ var _ = Describe("Context", func() {
 			Entry("no binding path specified", nil, defaultContainerPath),
 			Entry("binding path specified", &v1alpha1.BindingPath{ContainersPath: "foo.bar"}, "foo.bar"),
 		)
+		DescribeTable("should return slice of size 1 if application is specifiedt through label seclector", func(bindingPath *v1alpha1.BindingPath, expectedContainerPath string) {
```
pedjak, line 95: typo "specifiedt"

pedjak: it would be also more interesting to cover the case when there at least two applications with the same label - with that we check if the loop in https://github.com/redhat-developer/service-binding-operator/pull/971/files#diff-0168dc7ac8aa540873af78cf38f886f6a5ace89fb75dce9880490cedbed1d46eR157 really works.

pedjak (changes requested):
a few more unit tests are missing:

* when both application name and selector are set
* when label selector is used but no application is annotated so the result is empty
* there is an error at performing `List` call

and acceptance tests demonstrating the usage of the label selector and the situation when both name and label selector are set.

pkg/reconcile/pipeline/context/impl.go
baijum: Returning `nil` returns an empty slice of `Application` objects.
See this example: https://play.golang.org/p/QvQLT9CH3Fg

pkg/reconcile/pipeline/context/impl.go
baijum: Done. 5326c6f674cb869fdaa8242bd4cc454fb72941cc

pkg/reconcile/pipeline/context/impl.go
baijum: Done. 5326c6f674cb869fdaa8242bd4cc454fb72941cc

pkg/reconcile/pipeline/context/impl_test.go
baijum: Done. 5326c6f674cb869fdaa8242bd4cc454fb72941cc

pkg/reconcile/pipeline/context/impl.go
pedjak: Thanks, nice. However, according to the contract, we should not return  error in this place, just empty slice.

baijum: Fixed and added a test case for the same: 588535ccd6305f26c5dfe8e2ac33909a1d313ad9

pedjak:
@baijum can you think about some better commit message/PR summary describing what was done and how?

isutton:
> @baijum commented on this pull request.
> Returning `nil` returns an empty slice of `Application` objects.
> See this example: https://play.golang.org/p/QvQLT9CH3Fg

In this particular case, 'nil' is an un-initialized slice of a
Application items, not an empty initialized slice of Application
items.

***@***.***/golang-nil-vs-empty-slice-87fd51c0a4d
I've found a good summary on why this matters.

baijum:
> @baijum can you think about some better commit message/PR summary describing what was done and how?

baijum:
> In this particular case, 'nil' is an un-initialized slice of a Application items, not an empty initialized slice of Application items. 

Done. 538772c41782fe9cff89843112f7227d5433e51b

baijum:
> and acceptance tests demonstrating the usage of the label selector and the situation when both name and label selector are set.

I missed this. I will work on this.

baijum:
> > and acceptance tests demonstrating the usage of the label selector and the situation when both name and label selector are set.
> 
> I missed this. I will work on this.

Done. 5e472beddfcb48376f4bcd3620282d8ec3437082

baijum:
> * when both application name and selector are set

Done. 7debb128ae498a821dd404ba57825f8deeddffe5

baijum:
> * there is an error at performing `List` call

I couldn't figure out how to produce an error in the `List` call. Let me if we can skip this for now.

test/acceptance/features/bindAppToServiceWithLabelSelector.feature
```diff
@@ -0,0 +1,111 @@
+Feature: Bind an application with label selector to a service
```
pedjak, line 1: should we really consider this as a separate feature? this is all about referring application resources using attached labels

test/acceptance/features/bindAppToServiceWithLabelSelector.feature
```diff
@@ -0,0 +1,111 @@
...
+
+    As a user of Service Binding Operator
+    I want to bind applications with label selector to services it depends on
+
+    Background:
+        Given Namespace [TEST_NAMESPACE] is used
+        * Service Binding Operator is running
+
+    Scenario: Bind an application with label selector to backend service
```
pedjak, line 10: application resource has labels attached, but label selector is used inside service binding to define what application should be bound. Could we be a bit more verbose about the setup and the expected outcome? For example:

```
Bind successfully an application owning a given label to a service
```

test/acceptance/features/bindAppToServiceWithLabelSelector.feature
```diff
@@ -0,0 +1,111 @@
...
+                        app: gen-app-a-s-c
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        Then Service Binding "service-binding-a-s-c" is ready
+        And The application env var "BACKEND_USERNAME" has value "AzureDiamond"
+        And The application env var "BACKEND_PASSWORD" has value "hunter2"
+
+    Scenario: Attempt to bind an application with label selector and name to a service
```
pedjak, line 60: please rephrase to include the outcome as well.. in this case service binding will be unsuccessful, right?

test/acceptance/features/bindAppToServiceWithLabelSelector.feature
```diff
@@ -0,0 +1,111 @@
...
+                      matchLabels:
+                        app: gen-app-a-s-e
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+
+        Then jq ".status.conditions[] | select(.type=="CollectionReady").status" of Service Binding "service-binding-a-s-e" should be changed to "True"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").status" of Service Binding "service-binding-a-s-e" should be changed to "False"
+        And jq ".status.conditions[] | select(.type=="InjectionReady").reason" of Service Binding "service-binding-a-s-e" should be changed to "ApplicationNotFound"
```
pedjak, line 110: IMHO - `ApplicationNotFound` is not correct reason - simply we do not allow name and label to cooexist - Hence, something like `InvalidApplicationReference` should be more appropriate.

pedjak (changes requested):
it would be good to have an additonal acceptance test demonstrating how to bind 2 applications having the same label.

pedjak:
> I couldn't figure out how to produce an error in the List call. Let me know if we can skip this for now.

@baijum use `PrependReactor` on the fake client and define for what verb and resource you would like to have an error. Something like: https://github.com/kubernetes/client-go/issues/742#issuecomment-633676780

baijum:
> > I couldn't figure out how to produce an error in the List call. Let me know if we can skip this for now.
> 
> @baijum use `PrependReactor` on the fake client and define for what verb and resource you would like to have an error. Something like: [kubernetes/client-go#742 (comment)](https://github.com/kubernetes/client-go/issues/742#issuecomment-633676780)

Done. ee27a52e503e89daa95b50a3f4549960af35dcee

pmacik:
@baijum FYI, PR checks fails because of an issue with the workflows. This PR https://github.com/redhat-developer/service-binding-operator/pull/973 fixes that.

pkg/reconcile/pipeline/context/impl.go
```diff
@@ -118,15 +121,44 @@ func (i *impl) Services() ([]pipeline.Service, error) {
...
+				i.serviceBinding.Spec.Application.LabelSelector != nil &&
+				i.serviceBinding.Spec.Application.LabelSelector.MatchLabels != nil {
+				return nil, errs.New("Both label selector and name for application should not be used")
+			}
 			gvr, err := i.typeLookup.ResourceForReferable(ref)
-			if err != nil {
-				return nil, err
+			if i.serviceBinding.Spec.Application.Name != "" {
+				if err != nil {
+					return nil, err
```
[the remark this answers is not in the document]
baijum: Done. af6e990baa021c05e9fba2e7cb713ac745500ba9

pkg/reconcile/pipeline/context/impl.go
```diff
@@ -118,15 +121,44 @@ func (i *impl) Services() ([]pipeline.Service, error) {
...
+				}
+
+				objList, err := i.client.Resource(*gvr).Namespace(i.serviceBinding.Namespace).List(context.Background(), opts)
+				if err != nil {
+					emptyResult := make([]pipeline.Application, 0)
+					return emptyResult, err
+				}
+
+				if len(objList.Items) == 0 {
+					return nil, nil
```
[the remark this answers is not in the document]
baijum: Done. af6e990baa021c05e9fba2e7cb713ac745500ba9

pkg/reconcile/pipeline/context/impl_test.go
```diff
@@ -91,7 +92,175 @@ var _ = Describe("Context", func() {
 			Entry("no binding path specified", nil, defaultContainerPath),
 			Entry("binding path specified", &v1alpha1.BindingPath{ContainersPath: "foo.bar"}, "foo.bar"),
 		)
+		DescribeTable("should return slice of size 1 if application is specified through label seclector", func(bindingPath *v1alpha1.BindingPath, expectedContainerPath string) {
```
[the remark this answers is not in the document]
baijum: Done. af6e990baa021c05e9fba2e7cb713ac745500ba9

pkg/reconcile/pipeline/context/impl_test.go
```diff
@@ -91,7 +92,175 @@ var _ = Describe("Context", func() {
...
+			u2.SetGroupVersionKind(schema.GroupVersionKind{Group: "app", Version: "v1", Kind: "Foo"})
+			u2.SetLabels(map[string]string{"env": "prod"})
+
+			client := fake.NewSimpleDynamicClient(runtime.NewScheme(), u1, u2)
+
+			ctx := &impl{client: client, serviceBinding: &sb, typeLookup: typeLookup}
+
+			applications, err := ctx.Applications()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(applications).To(HaveLen(2))
```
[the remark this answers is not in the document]
baijum: Done. af6e990baa021c05e9fba2e7cb713ac745500ba9

pkg/reconcile/pipeline/context/impl_test.go
```diff
@@ -91,7 +92,175 @@ var _ = Describe("Context", func() {
...
+
+			ref := v1alpha1.Application{
+				Ref: v1alpha1.Ref{
+					Group:   "app",
+					Version: "v1",
+					Kind:    "Foo",
+					Name:    "app1",
+				},
+				LabelSelector: ls,
+			}
```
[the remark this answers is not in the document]
baijum: The test is to check the effect of two kinds of application references: label & name.
Even though it is not directly used, this setup is required to validate the functionality.

pkg/reconcile/pipeline/context/impl_test.go
```diff
@@ -91,7 +92,175 @@ var _ = Describe("Context", func() {
...
+
+			sb := v1alpha1.ServiceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "sb1",
+					Namespace: "ns1",
+				},
+				Spec: v1alpha1.ServiceBindingSpec{
+					Application: &ref,
+				},
+			}
```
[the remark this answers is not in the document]
baijum: The test is to check the effect of two kinds of application references: label & name.
Even though it is not directly used, this setup is required to validate the functionality.

pkg/reconcile/pipeline/context/impl_test.go
```diff
@@ -91,7 +92,175 @@ var _ = Describe("Context", func() {
...
+			client := fake.NewSimpleDynamicClient(runtime.NewScheme())
+
+			ctx := &impl{client: client, serviceBinding: &sb, typeLookup: typeLookup}
+
+			_, err := ctx.Applications()
+			Expect(err).To(HaveOccurred())
```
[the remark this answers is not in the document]
baijum: Done. af6e990baa021c05e9fba2e7cb713ac745500ba9

test/acceptance/features/bindAppToServiceWithLabelSelector.feature
baijum: Done. af6e990baa021c05e9fba2e7cb713ac745500ba9

test/acceptance/features/bindAppToServiceWithLabelSelector.feature
baijum: Done. af6e990baa021c05e9fba2e7cb713ac745500ba9

baijum:
> it would be good to have an additonal acceptance test demonstrating how to bind 2 applications having the same label.

I will work on this.

test/acceptance/features/bindAppToServiceWithLabelSelector.feature
baijum: I need some more time to figure out how to introduce a new failure reason.

test/acceptance/features/bindAppToServiceWithLabelSelector.feature
baijum: I think a separate feature file for the label selector is fine.

pkg/reconcile/pipeline/context/impl.go
```diff
@@ -118,15 +121,43 @@ func (i *impl) Services() ([]pipeline.Service, error) {
...
+				if err != nil {
+					return emptyResult, err
+				}
+
+				if len(objList.Items) == 0 {
+					return emptyResult, nil
+				}
+
+				for index, _ := range objList.Items {
+					i.applications = append(i.applications, &application{gvr: gvr, persistedResource: &(objList.Items[index]), bindingPath: i.serviceBinding.Spec.Application.BindingPath})
```
baijum, line 157: The `persistedResource` value (address of the item) is taken using the index.
Ref. https://play.golang.org/p/ePs7st6f91g

test/acceptance/features/bindAppToServiceWithLabelSelector.feature
pedjak: can you backup your statement with some reason?

baijum: I am fine with moving to another file. Please suggest the file.

pkg/reconcile/pipeline/context/impl.go
```diff
@@ -118,15 +121,43 @@ func (i *impl) Services() ([]pipeline.Service, error) {
 func (i *impl) Applications() ([]pipeline.Application, error) {
 	if i.applications == nil {
 		if ref := i.serviceBinding.Spec.Application; ref != nil {
+			if i.serviceBinding.Spec.Application.Name != "" &&
+				i.serviceBinding.Spec.Application.LabelSelector != nil &&
+				i.serviceBinding.Spec.Application.LabelSelector.MatchLabels != nil {
+				return nil, errs.New("Both label selector and name for application should not be used")
```
pedjak, line 127: we can extract error into a var that can be used later in test assertion - instead of string copying.

pkg/reconcile/pipeline/context/impl_test.go
```diff
@@ -91,7 +94,182 @@ var _ = Describe("Context", func() {
...
+				Spec: v1alpha1.ServiceBindingSpec{
+					Application: &ref,
+				},
+			}
+
+			u := &unstructured.Unstructured{}
+			u.SetName("app")
+			u.SetNamespace(sb.Namespace)
+			u.SetGroupVersionKind(schema.GroupVersionKind{Group: "app", Version: "v1", Kind: "Foo"})
+			u.SetLabels(map[string]string{"env": "stage"})
```
pedjak, line 223: we do not need to create any app object at all, since we are going to fail earlier. Omitting this will also implicitly check that we do not try to access the cluster for this scenario.

pkg/reconcile/pipeline/context/impl_test.go
```diff
@@ -91,7 +94,182 @@ var _ = Describe("Context", func() {
...
+				func(action testing.Action) (handled bool, ret runtime.Object, err error) {
+					return true, &v1.DeploymentList{}, e.New("Error listing foo")
+				})
+
+			ctx := &impl{client: client, serviceBinding: &sb, typeLookup: typeLookup}
+
+			_, err := ctx.Applications()
+			Expect(err).To(HaveOccurred())
+			// FIXME: This is not the expected error message
+			Expect(err.Error()).To(ContainSubstring("no kind is registered for the type v1.DeploymentList in scheme"))
```
pedjak, line 271: the error should be the same as the one you put in line 263: "Error listing foo"

pedjak (changes requested):

pkg/reconcile/pipeline/context/impl_test.go
baijum: For some reason, that error message is not coming. I have used `client.PrependReactor` (line 261-264) as per earlier suggestion, so the listing failed as expected.
I tried to use `unstructured.Unstructured{}` instead of `v1.DeploymentList{}`, but that also seems to be not working. I will come back to this issue once everything else finished.

pkg/reconcile/pipeline/context/impl_test.go
```diff
@@ -91,7 +94,182 @@ var _ = Describe("Context", func() {
...
+
+			ctx := &impl{client: client, serviceBinding: &sb, typeLookup: typeLookup}
+
+			applications, err := ctx.Applications()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(applications).To(HaveLen(2))
+			Expect(applications[0].Resource()).To(Equal(u1))
+			Expect(applications[0].ContainersPath()).To(Equal(expectedContainerPath))
+			Expect(applications[1].Resource()).To(Equal(u2))
+			Expect(applications[1].ContainersPath()).To(Equal(expectedContainerPath))
```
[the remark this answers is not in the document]
baijum: Order is not reliable and not part of the API.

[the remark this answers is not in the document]
baijum: Check Omega assertion library.

baijum:
> > it would be good to have an additonal acceptance test demonstrating how to bind 2 applications having the same label.
> 
> I will work on this.

Done. 53bb65a118d53a3a3e299860f4b3473f88d2b42d

pkg/reconcile/pipeline/context/impl.go
baijum: Created separate issue: #974

pkg/reconcile/pipeline/context/impl_test.go
baijum: This code has been removed. Ref.  #974

pkg/reconcile/pipeline/context/impl_test.go
baijum: Done. 82294850ad30c8fea6de5e767b5852a5d0158c51

pkg/reconcile/pipeline/context/impl_test.go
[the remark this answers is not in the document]
baijum: Done. b46fe696d76e4318cfa1d8871a25b2d0c32a902a

pkg/reconcile/pipeline/context/impl_test.go
```diff
@@ -91,7 +93,145 @@ var _ = Describe("Context", func() {
...
+				},
+			}
+
+			gvr := &schema.GroupVersionResource{Group: "app", Version: "v1", Resource: "foos"}
+			typeLookup.EXPECT().ResourceForReferable(&ref).Return(gvr, nil)
+
+			client := fake.NewSimpleDynamicClient(runtime.NewScheme())
+			client.PrependReactor("list", "foos",
+				func(action testing.Action) (handled bool, ret runtime.Object, err error) {
+					return true, nil, e.New("Error listing foo")
```
pedjak, line 226: DRY: please extract error in a variable and use it later in the assertion

pkg/reconcile/pipeline/context/impl_test.go
```diff
@@ -91,7 +93,145 @@ var _ = Describe("Context", func() {
...
+				})
+
+			ctx := &impl{client: client, serviceBinding: &sb, typeLookup: typeLookup}
+
+			_, err := ctx.Applications()
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("Error listing foo"))
```
pedjak, line 233: The error we get must be the one we got from the client, and the message should be the identical. Such assertion assures that the error is propagated to the caller:

```
Expect(err).To(Equal(expectedError))
```

pkg/reconcile/pipeline/context/impl_test.go
```diff
@@ -122,6 +262,7 @@ var _ = Describe("Context", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(errors.IsNotFound(err)).To(BeTrue())
 		})
+
```
pedjak, line 265: please revert unrelated whitespace change.

test/acceptance/features/steps/generic_testapp.py
```diff
@@ -74,3 +76,33 @@ def check_file_value(context, file_path):
...
+@given(u'The common label "{label}" is set for both apps')
+def set_common_label(context, label):
+    context.application1.set_label(f"{label}")
+    context.application2.set_label(f"{label}")
+
+
+@then(u'The application env var "{name}" has value "{value}" in both apps')
+def check_env_var_value_in_both_apps(context, name, value):
+    found = polling2.poll(lambda: context.application1.get_env_var_value(name) == value, step=5, timeout=400)
+    assert found, f'Env var "{name}" should contain value "{value}" in {context.application1}'
```
pedjak, line 105: if we reach this line, then `found` is `true` always - otherwise Timeout exception occurs. Hence, this and the similar line at line 108 can be removed.

pedjak (changes requested):

test/acceptance/features/steps/generic_testapp.py
baijum: Done. 5e91f6d96db0011740c9422b96e2de2d0a654551

pkg/reconcile/pipeline/context/impl_test.go
baijum: Done. 5e91f6d96db0011740c9422b96e2de2d0a654551

pkg/reconcile/pipeline/context/impl_test.go
baijum: Done. 5e91f6d96db0011740c9422b96e2de2d0a654551

pkg/reconcile/pipeline/context/impl_test.go
baijum: Done. 5e91f6d96db0011740c9422b96e2de2d0a654551

pedjak (approved):
+1 after having required CI jobs passing. Please also squash the commits.

PR #972, opened 2021-06-07
[WIP] Add GHA workflow to push operator images for PRs

pmacik:
### Motivation

<< Why this change is relevant, what's the "story" behind it? >>

### Changes

### Testing

<< How to test those changes, reference implementation of unit/acceptance tests. >>

### DIFF
--- a/.github/actions/setup-cli/action.yaml
+++ b/.github/actions/setup-cli/action.yaml
@@ -1,16 +1,45 @@
 # action.yml
 name: 'Setup Tools'
 description: 'Setup CLI configuration with kubectl operator-sdk opm minikube, setup path, kubeconfig, minikube and docker info '
+inputs:
+  operator-sdk:
+    description: "Install Operator SDK"
+    required: false
+    default: false
+  opm:
+    description: "Install opm"
+    required: false
+    default: false
+  kubectl:
+    description: "Install kubectl"
+    required: false
+    default: false
+  minikube:
+    description: "Install minikube"
+    required: false
+    default: false
+  start-minikube:
+    description: "Start minikube cluster"
+    required: false
+    default: false
 runs:
   using: "composite"
   steps:
     - id: setup-path
       run: |
         mkdir -p $GITHUB_WORKSPACE/bin/
-        echo "PATH=$PATH:$GITHUB_WORKSPACE/bin/" >> $GITHUB_ENV
+        echo "PATH=$GITHUB_WORKSPACE/bin:$PATH" >> $GITHUB_ENV
       shell: bash
 
-    - run: $GITHUB_ACTION_PATH/setup_tools.sh
+    - id: setup-tools
+      env:
+        OPERATOR_SDK: ${{ inputs.operator-sdk }}
+        OPM: ${{ inputs.opm }}
+        KUBECTL: ${{ inputs.kubectl }}
+        MINIKUBE: ${{ inputs.minikube }}
+        START_MINIKUBE: ${{ inputs.start-minikube }}
+      run: |
+        $GITHUB_ACTION_PATH/setup_tools.sh
       shell: bash
 
     - id: set-kubeconfig
@@ -22,14 +51,7 @@ runs:
 
     - id: setup-minikube
       run: |
-        ./hack/start-minikube.sh start --kubernetes-version=v${K8S_VERSION} --driver=$CONTAINER_RUNTIME --cpus $(nproc) --memory 5g
-      shell: bash
-
-    - id: docker-info
-      run: |
-        eval $(minikube docker-env)
-        docker ps
-        kubectl get nodes -o yaml
-        kubectl cluster-info
-        docker info
+        if [ "${{ inputs.start-minikube }}" == "true" ]; then
+          ./hack/start-minikube.sh start --kubernetes-version=v${K8S_VERSION} --driver=$CONTAINER_RUNTIME --cpus $(nproc) --memory 5g
+        fi
       shell: bash
--- a/.github/actions/setup-podman/registries.conf
+++ b/.github/actions/setup-podman/registries.conf
@@ -1,0 +1,6 @@
+unqualified-search-registries = ['docker.io']
+
+[[registry]]
+prefix = "REGISTRY_PREFIX"
+insecure = true
+location = "localhost:5000"
--- a/.github/workflows/pr-checks-build-images.yaml
+++ b/.github/workflows/pr-checks-build-images.yaml
@@ -1,0 +1,70 @@
+name: PR checks - Build operator images
+
+on:
+  pull_request:
+    branches:
+      - master
+
+env:
+  SDK_VERSION: "1.3.0"
+  OPM_VERSION: "1.15.2"
+  CONTAINER_RUNTIME: "podman"
+  ARTIFACTS: "artifacts"
+  REGISTRY_PREFIX: quay.io/redhat-developer
+  REPO: servicebinding-operator
+
+jobs:
+  build-operator-images:
+    name: "Build operator images (PR)"
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout Git Repository
+        uses: actions/checkout@v2
+
+      - name: Setup CLI
+        uses: ./.github/actions/setup-cli
+        with:
+          operator-sdk: true
+          opm: true
+
+      - name: Setup local registry
+        run: |
+          .github/actions/setup-podman/setup-local-registry.sh
+          .github/actions/setup-podman/setup-podman-wrapper.sh
+
+      - name: Build images
+        env:
+          PR_NUMBER: ${{github.event.pull_request.number}}
+          PR_SHA: ${{github.event.pull_request.head.sha}}
+          OPERATOR_REPO_REF: ${{env.REGISTRY_PREFIX}}/${{env.REPO}}
+        run: |
+          export TAG=pr-${PR_NUMBER}-${PR_SHA:0:8}
+          export OPERATOR_IMAGE_REF=${OPERATOR_REPO_REF}:${TAG}
+          export OPERATOR_BUNDLE_IMAGE_REF=${OPERATOR_IMAGE_REF}-bundle
+          export OPERATOR_INDEX_IMAGE_REF=${OPERATOR_IMAGE_REF}-index
+
+          which podman
+
+          BUILDAH_FORMAT=docker make SKIP_REGISTRY_LOGIN=true release-operator -o registry-login
+
+          mkdir -p ${ARTIFACTS}
+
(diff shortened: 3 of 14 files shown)


.github/actions/setup-cli/action.yaml
```diff
@@ -22,14 +51,7 @@ runs:
 
     - id: setup-minikube
       run: |
-        ./hack/start-minikube.sh start --kubernetes-version=v${K8S_VERSION} --driver=$CONTAINER_RUNTIME --cpus $(nproc) --memory 5g
-      shell: bash
-
-    - id: docker-info
-      run: |
-        eval $(minikube docker-env)
```
pedjak, line 30: this line configured usage of internal minikube docker engine, not sure if we still want to use it.

.github/actions/setup-cli/action.yaml
```diff
@@ -22,14 +51,7 @@ runs:
...
-        docker ps
-        kubectl get nodes -o yaml
-        kubectl cluster-info
-        docker info
```
pedjak, line 34: this was displaying a few stat infos about docker engine and the cluster.

.github/actions/setup-podman/registries.conf
```diff
@@ -0,0 +1,6 @@
+unqualified-search-registries = ['docker.io']
```
pedjak, line 1: this is a template file, not the final registrues.conf.. perhaps we should name it like this to reflect that, e.g. `registries.conf.tpl` ?

.github/workflows/pr-checks-build-images.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+          which podman
+
+          BUILDAH_FORMAT=docker make SKIP_REGISTRY_LOGIN=true release-operator -o registry-login
+
+          mkdir -p ${ARTIFACTS}
+
+          echo -n "${OPERATOR_INDEX_IMAGE_REF}" > ${ARTIFACTS}/operator-index.ref
+          echo -n "${OPERATOR_BUNDLE_IMAGE_REF}" > ${ARTIFACTS}/operator-bundle.ref
+          echo -n "${OPERATOR_IMAGE_REF}" > ${ARTIFACTS}/operator-image.ref
```
pedjak, line 55: instead of 3 individual files, how about creating a file that can be sourced directly in bash, with the following cmd:

```
source operator.refs
```

and that file looks like:

```
OPERATOR_IMAGE_REF=quay.io/redhat-developer/servicebinding-operator:pr-xxx
OPERATOR_BUNDLE_IMAGE_REF=quay.io/redhat-developer/servicebinding-operator:pr-bundle-xxx
OPERATOR_INDEX_IMAGE_REF=quay.io/redhat-developer/servicebinding-operator:pr-index-xxx
```

.github/workflows/pr-checks-clean-images.yaml
```diff
@@ -0,0 +1,27 @@
...
+  OPERATOR_REPO_REF: quay.io/redhat-developer/servicebinding-operator
+
+jobs:
+  clean-operator-images:
+    name: "Clean operator images (PR)"
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout Git Repository
+        uses: actions/checkout@v2
```
pedjak, line 19: Do we really need to checkout the project since we do not refer any files later?

.github/workflows/pr-checks-push-images.yaml
```diff
@@ -0,0 +1,50 @@
...
+  push-operator-images:
+    name: "Push operator images (PR)"
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout Git Repository
+        uses: actions/checkout@v2
+
+      - name: Setup CLI
+        uses: ./.github/actions/setup-cli
```
pedjak, line 22: Do we really need these two steps?

.github/workflows/pr-checks-push-images.yaml
```diff
@@ -0,0 +1,50 @@
...
+          tar -xvf registry.tar.gz
+          .github/actions/setup-podman/setup-local-registry.sh
+
+      - name: Push operator, bundle and index images
+        run: |
+          podman images
+          set -x
+          for tag in $(skopeo list-tags --tls-verify=false docker://localhost:5000/${REPO} | jq -r '.Tags[] | select(startswith("pr-${{github.event.pull_request.number}}-"))'); do
+            sha=$(skopeo inspect --tls-verify=false docker://localhost:5000/${REPO}:${tag} | jq -r '.Digest')
+            skopeo copy --dest-creds ${{secrets.QUAY_USERNAME}}:${{secrets.QUAY_TOKEN}} --all --src-tls-verify=false docker://localhost:5000/${REPO}@${sha} docker://${REGISTRY_PREFIX}/${REPO}@${sha}
```
pedjak, line 48: from my testing it should be enough to copy images using tags, i.e. this line can be removed.

.github/workflows/pr-checks.yaml
```diff
@@ -89,15 +84,34 @@ jobs:
...
+        uses: marcofaggian/action-download-multiple-artifacts@v3.0.8
+        with:
+          names: operator-refs-${{github.event.pull_request.number}}-${{github.event.pull_request.head.sha}}
 
       - name: Acceptance tests
         timeout-minutes: 60
         run: |
-          eval $(minikube docker-env)
-          make OPERATOR_REPO_REF=$(minikube ip):5000/sbo SKIP_REGISTRY_LOGIN=true release-operator -o registry-login test-acceptance-with-bundle
+          export OPERATOR_INDEX_IMAGE_REF=$(cat operator-index.ref)
```
pedjak, line 106: if you go with the proposal from above, then all vars could be initialized via simple:

```
source ./operator.refs
```

.github/workflows/pr-checks.yaml
```diff
@@ -14,7 +14,7 @@ env:
   K8S_VERSION: "1.19.2"
   MINIKUBE_VERSION: "1.15.1"
   TEST_ACCEPTANCE_CLI: "kubectl"
-  CONTAINER_RUNTIME: "docker"
+  CONTAINER_RUNTIME: "podman"
```
pedjak, line 17: not sure if we need to flip here the values because we do not use neither of them during the installation. Perhaps dropping the line completely?

hack/prepare-operatorhub-pr.sh
```diff
@@ -9,7 +9,7 @@
 
 TMP_OCI_PATH=$(mktemp -d out/sbo-bundle-oci.XXX)
 
-skopeo copy --src-tls-verify=false docker://$2 oci:$TMP_OCI_PATH:bundle
+skopeo copy --src-tls-verify=false --src-no-creds docker://$2 oci:$TMP_OCI_PATH:bundle
```
pedjak, line 12: how is this change related to PR?

pedjak (approved):
look really good, a few remarks, feel free to decide on your own what would you like to take or not

.github/workflows/pr-checks-build-images.yaml
```diff
@@ -0,0 +1,70 @@
...
+          .github/actions/setup-podman/setup-local-registry.sh
+          .github/actions/setup-podman/setup-podman-wrapper.sh
+
+      - name: Build images
+        env:
+          PR_NUMBER: ${{github.event.pull_request.number}}
+          PR_SHA: ${{github.event.pull_request.head.sha}}
+          OPERATOR_REPO_REF: ${{env.REGISTRY_PREFIX}}/${{env.REPO}}
+        run: |
+          export TAG=pr-${PR_NUMBER}-${PR_SHA:0:8}
```
pedjak, line 42: IMHO, we should not have PR_SHA as the part of the tag - if we are adding commits to PR before merge and while testing, we are going to produce a number of images, but only the last one after merge will be removed. `TAG=pr-${PR_NUMBER}` will override images created by past commits on the same branch/PR.

pmacik: Actually, the `Clean PR images` step deletes all the images that starts with `pr-<PR_NUMBER>-` - not only the last one, so all the images related to the PR are deleted.

Additionally, I correlate and ensure the correct image built from the matching commit using the PR and SHA at OpenShift CI side...

pedjak: ah, thanks, I missed that.

.github/actions/setup-podman/registries.conf
pmacik: Good idea, thx @pedjak

.github/workflows/pr-checks-build-images.yaml
pmacik: Good idea, thx @pedjak

.github/workflows/pr-checks-clean-images.yaml
pmacik: We are refering the `./.github/actions/clean-images` action

.github/workflows/pr-checks-push-images.yaml
pmacik: We need the first for the `Setup local registry` step were we refering `.github/actions/setup-podman/setup-local-registry.sh` script and I've used the other to set up PATH, but that can be done in a standalone simple step.

.github/workflows/pr-checks-push-images.yaml
pmacik: Good point, thx @pedjak

.github/workflows/pr-checks.yaml
pmacik: Good point, thx @pedjak

.github/actions/setup-cli/action.yaml
pmacik: Good catch, I've put it back but changed to `podman` thx @pedjak

hack/prepare-operatorhub-pr.sh
pmacik: Without it, the hack script fails with the following:
```
./hack/prepare-operatorhub-pr.sh 0.7.1 quay.io/pmacik/sbo-pr-checks:pr-115-d724d340-bundle
time="2021-06-07T23:26:11Z" level=fatal msg="Error initializing source docker://quay.io/pmacik/sbo-pr-checks:pr-115-d724d340-bundle: error getting username and password: 1 error occurred:\n\t* error reading JSON file \"/run/containers/1001/auth.json\": open /run/containers/1001/auth.json: permission denied\n\n"
make: *** [Makefile:272: prepare-operatorhub-pr] Error 1
```

.github/actions/setup-cli/action.yaml
pmacik: Right, we do not use that any more... thx, @pedjak

PR #976, opened 2021-06-10
Optimize acceptance testing jobs on OpenShift CI

pmacik:
### Motivation

Currenly, the acceptance testing jobs running with OpenShift CI build a custom test image in which they are executed. The acceptance tests are running in the local mode which means that couple of tools such as Go, operator-sdk and others are needed to build the operator binary locally.

Since #973 has been merged it is now possible and required to run acceptance tests on OpenShift with SBO installed in the cluster the test image only needs to be able to run the acceptance tests in the remote mode which requires only small number of tools like Python and make.

Additionally, having a `root` image Dockerfile in a repository is making it hard to change since the way [how OpenShift CI jobs work](https://docs.ci.openshift.org/docs/architecture/ci-operator/#building-artifacts).

ISSUE: https://issues.redhat.com/browse/APPSVC-884

### Changes

This PR:
 * Moves the OpenShift CI's `root` image definition directly into the CI's job configuration (via https://github.com/openshift/release/pull/14755)
 * Updates main Makefile to be able to run the acceptance tests without Go (it does not need it)
 * is a dependency for https://github.com/openshift/release/pull/14755

### Testing


Makefile
```diff
@@ -57,11 +57,13 @@ TEST_ACCEPTANCE_TAGS_ARG ?= --tags="~@disabled" --tags="~@examples"
 endif
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
+ifneq (,$(shell go version || echo ""))
 ifeq (,$(shell go env GOBIN))
 GOBIN=$(shell go env GOPATH)/bin
 else
 GOBIN=$(shell go env GOBIN)
 endif
+endif
```
pedjak, line 66: I am not seeing that we use `GOBIN` anywhere at all in the project - can we simply remove this block?

pmacik: Good point - I don't see it used anywhere, too... let me remove it. Thx @pedjak

Issue #863, opened 2021-02-05
Consolidate User Guide


pedjak:
The current user guide living under doc folder needs to be consolidated:
*    list all features
*    explain each field of service binding resource
*    link to/mention spec where appropriate

wtrocki:
@akashshinde Do you have any update for the guide?

pedjak:
Closing in favor of #1019

PR #978, opened 2021-06-14
Add provisioned service support

### LINKED ISSUE #548: Support duck type from spec
The Service Binding Specification introduces a [duck type](https://github.com/k8s-service-bindings/spec#provisioned-service) for provisioned services to make themselves bindable.  

To support the desire of being spec compliant, this issue should enhance SBO to support this pattern for target services.

pedjak:
Application can be bound to a provisioned service:

https://github.com/k8s-service-bindings/spec#provisioned-service

* collect.ProvisionedService handler detects if referred service contains a reference to secret containing the bindings
* unit and acceptance tests demonstrate the contact and exposed functionality

Fixes #548

### DIFF
--- a/pkg/reconcile/pipeline/context/impl.go
+++ b/pkg/reconcile/pipeline/context/impl.go
@@ -10,6 +10,7 @@ import (
 	"github.com/redhat-developer/service-binding-operator/pkg/client/kubernetes"
 	"github.com/redhat-developer/service-binding-operator/pkg/converter"
 	"github.com/redhat-developer/service-binding-operator/pkg/reconcile/pipeline"
+	"github.com/redhat-developer/service-binding-operator/pkg/util"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -34,6 +35,8 @@ type impl struct {
 	applications []*application
 
 	bindingItems pipeline.BindingItems
+
+	bindings []pipeline.Bindings
 
 	retry bool
 	stop  bool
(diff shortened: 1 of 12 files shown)


pkg/reconcile/pipeline/context/impl.go
```diff
@@ -167,14 +170,36 @@ func (i *impl) AddBindingItem(item *pipeline.BindingItem) {
 }
 
 func (i *impl) BindingItems() pipeline.BindingItems {
-	return i.bindingItems
+	var allItems pipeline.BindingItems
+	for _, b := range i.bindings {
+		items, err := b.Items()
```
isutton, line 175: Is it possible to drop the error return value here?

pkg/reconcile/pipeline/context/impl.go
```diff
@@ -227,16 +267,20 @@ func (i *impl) persistBinding() error {
 }
 
 func (i *impl) persistSecret() (string, error) {
-	if i.bindingItems == nil || len(i.bindingItems) == 0 {
+	name, secretExist := i.bindingSecretName()
```
isutton, line 270: Perhaps `checkBindingSecretName` would fit better? Another option would be split between the `secretName` getter and the check whether the secret exists in the internal `bindingItems`.

isutton (approved):
LGTM with some minors; will leave to you @pedjak to modify the minors or squash-and-merge.

baijum:
What if the Provisioned Service also got CR/CRD annotations? Does the binding Secret resource contain values from Provisioned Service Secret and the values extracted through CR/CRD annotations?

We need to document the behavior. Probably we can add some acceptance tests also.

pkg/reconcile/pipeline/context/impl.go
pedjak: not sure what do you mean - would you like that return error to the caller, instead of trying to collect as much as possible items?

pkg/reconcile/pipeline/context/impl.go
pedjak: the function does not perform a check - it return the binding secret name that can be used in stages for injecting it into proper resources. That name can be computed, or it could be reused from a already existing secret, if not additional bindings need to be added. However, the caller does not need to know that. Hence, the function return the secret name and information if the secret under that name already exist. This is pure internal method, and that information is needed in `Close()` method to decide if the secret need to be created or not.

pedjak:
> What if the Provisioned Service also got CR/CRD annotations? Does the binding Secret resource contain values from Provisioned Service Secret and the values extracted through CR/CRD annotations?

@baijum  the code submitted in this PR allows that - I have added the acceptance test demonstrating this. The documentation will be added as the part of https://github.com/redhat-developer/service-binding-operator/issues/863

pkg/reconcile/pipeline/context/impl.go
isutton: The error is captured but ignored, and also the fact that `Items()` doesn't read as a method that might return an error.

pedjak: you are right - according to the current contract `Items()` does not return error, it is just a getter.

Issue #900, opened 2021-03-04
Add RBAC Support


jasperchui:
Reference EPIC #892 

Spec:

https://github.com/k8s-service-bindings/spec#role-based-access-control-rbac

Issue #454, opened 2020-05-08
Narrow down what the controller SA can 'update'


sbose78:
I would like to propose that we must reduce the following [permission](https://github.com/redhat-developer/service-binding-operator-manifests/blob/master/0.1.1-293/service-binding-operator.v0.1.1-293.clusterserviceversion.yaml#L145) to allow `update` of only those resources which we know for sure are going to have the binding secret injected. 

```
- apiGroups:
          - '*'
          resources:
          - '*'
          verbs:
          - get
          - list
          - watch
          - update
```

That way we'll be able to reduce the privileges of the service account to have permissions to `update` popular podSpec-based workloads and any other well-known resource types.

- deployments
- deploymentconfigs
- daemonsets
- replicasets
- statefulsets
- services ( knative )

We only need to be able to get, read and watch all resources because we are dealing with CRDs and we don't know in advance what is to be used for binding.

sbose78:
cc @redhat-developer/service-binding

sbose78:
Note: Changes should be made in this repo's /deploy directory.

pedjak:
@sbose78 how about that we drop the above block from `role.yaml` and then define an additional rolebinding for the service account:

```yaml
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: service-binding-operator
subjects:
- kind: ServiceAccount
  name: service-binding-operator
roleRef:
  kind: ClusterRole
  name: edit
  apiGroup: rbac.authorization.k8s.io
```

`edit` is aggregating cluster role, so we automatically get access to other resources as well, if the other operators behave properly.

sbose78:
@pedjak +1

pedjak:
Fixed by #979 and #1002

PR #979, opened 2021-06-16
Add rbac support for spec alignment

bamachrn:
### Motivation
https://github.com/redhat-developer/service-binding-operator/issues/900

### Changes
* Added Aggregated cluster role for the operator to get access to the resources
* Added rolebindings for the openshift-operators and operators namespace

### Testing

For testing this we need to 
* re-create the bundle manifests
* Deploy the operator and bundle to the cluster

### DIFF
--- a/config/rbac/servicebinding_controller_role.yaml
+++ b/config/rbac/servicebinding_controller_role.yaml
@@ -1,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: controller-role
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
+      service.binding/controller: "true"
+rules: []
--- a/config/rbac/servicebinding_controller_rolebinding.yaml
+++ b/config/rbac/servicebinding_controller_rolebinding.yaml
@@ -1,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: controller-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: service-binding-controller-role
+subjects:
+- kind: ServiceAccount
+  name: service-binding-operator
+  namespace: operators
+- kind: ServiceAccount
+  name: service-binding-operator
+  namespace: openshift-operators
--- a/test/acceptance/features/steps/openshift.py
+++ b/test/acceptance/features/steps/openshift.py
@@ -144,8 +144,12 @@ spec:
             print('Resource list is empty under namespace - {}'.format(namespace))
             return None
 
-    def is_resource_in(self, resource_type):
-        output, exit_code = self.cmd.run(f'{ctx.cli} get {resource_type}')
+    def is_resource_in(self, resource_type, resource_name=None):
+        if resource_name is None:
+            _, exit_code = self.cmd.run(f'{ctx.cli} get {resource_type}')
+        else:
+            _, exit_code = self.cmd.run(f'{ctx.cli} get {resource_type} {resource_name}')
+
         return exit_code == 0
 
     def wait_for_pod(self, pod_name_pattern, namespace, interval=5, timeout=600):
@@ -299,6 +303,14 @@ spec:
         else:
             print(f'Error getting value for {resource_type}/{name} in {namespace} path={json_path}: {output}')
             return None
+
+    def get_json_resource(self, resource_type, name, namespace):
+        error_msg = f"Error in getting resource: '{resource_type}' '{name}' namespace: '{namespace}'"
+        output, exit_code = self.cmd.run(f'{ctx.cli} get {resource_type} {name} -n {namespace} -o json')
+        assert exit_code == 0, error_msg
+        json_output = json.loads(output)
+        assert json_output is not None, "Error in parsing JSON"
+        return json_output
 
     def get_resource_info_by_jq(self, resource_type, name, namespace, jq_expression, wait=False, interval=5, timeout=120):
         output, exit_code = self.cmd.run(f'{ctx.cli} get {resource_type} {name} -n {namespace} -o json | jq  \'{jq_expression}\'')
--- a/test/acceptance/features/steps/servicebindingoperator.py
+++ b/test/acceptance/features/steps/servicebindingoperator.py
@@ -15,8 +15,9 @@ class Servicebindingoperator():
         self.name = name
 
     def check_crd(self):
-        crd_name = "servicebinding"
-        assert self.openshift.is_resource_in(crd_name), f"CRD '{crd_name}' does not exist"
+        crd_type = "crd"
+        crd_name = "servicebindings.service.binding"
+        assert self.openshift.is_resource_in(crd_type, crd_name), f"CRD '{crd_name}' does not exist"
         return True
 
     def check_deployment(self):
--- a/test/acceptance/features/steps/steps.py
+++ b/test/acceptance/features/steps/steps.py
@@ -436,3 +436,30 @@ def update_binding_secret(context):
     output = openshift.apply(yaml.dump(secret_yaml), context.namespace.name)
     result = re.search(rf'.*{context.sb_secret}.*(created|unchanged|configured)', output)
     assert result is not None, f"Unable to apply YAML for binding secret '{context.sb_secret}': {output}"
+
+
+@then(u'cluster role "{cluster_role_name}" is available in the cluster')
+def check_cluster_role_exists(context, cluster_role_name):
+    openshift = Openshift()
+    assert openshift.is_resource_in("clusterroles", cluster_role_name), f"Could not find the cluster role : '{cluster_role_name}' in the cluster"
+
+
+@then(u'operator service account is bound to "{cluster_role_name}" in "{cluster_rolebinding_name}"')
+def check_service_account_bound(context, cluster_role_name, cluster_rolebinding_name):
+    openshift = Openshift()
+    sbr = Servicebindingoperator()
+    operator_namespace = openshift.lookup_namespace_for_resource("deployments", sbr.name)
+    sa_name = openshift.get_resource_info_by_jsonpath("deployments", sbr.name, operator_namespace, "{.spec.template.spec.serviceAccount}")
+
+    rolebinding_json = openshift.get_json_resource("clusterrolebinding", cluster_rolebinding_name, operator_namespace)
+    assert rolebinding_json["roleRef"]["name"] == cluster_role_name, f"Could not find rolebinding for the role '{cluster_role_name}'"
+
+    subjects_list = rolebinding_json["subjects"]
+    subject_found = False
+    for subject in subjects_list:
+        if subject["kind"] == "ServiceAccount" and subject["name"] == sa_name:
+            subject_found = True
+        else:
+            continue
+
+    assert subject_found, f"Could not find rolebinding for the role '{cluster_role_name}'"
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -6,6 +6,8 @@ resources:
 - leader_election_role_binding.yaml
 - servicebinding_editor_role.yaml
 - servicebinding_viewer_role.yaml
+- servicebinding_controller_role.yaml
+- servicebinding_controller_rolebinding.yaml
 # Comment the following 4 lines if you want to disable
 # the auth proxy (https://github.com/brancz/kube-rbac-proxy)
 # which protects your /metrics endpoint.
--- a/test/acceptance/features/pluggableRBAC.feature
+++ b/test/acceptance/features/pluggableRBAC.feature
@@ -1,0 +1,6 @@
+Feature: Making Service Biniding Operator RBAC pluggable so that other controllers/admins can add additional rules for the service binding operator.
+    Scenario: Service Binding Operator installation should contain aggregating cluster role and are bound with Namespace.
+        Given Namespace [TEST_NAMESPACE] is used
+        When Service Binding Operator is running
+        Then cluster role "service-binding-controller-role" is available in the cluster
+        And operator service account is bound to "service-binding-controller-role" in "service-binding-controller-rolebinding"


controllers/servicebinding_controller.go
```diff
@@ -49,7 +49,6 @@ type ServiceBindingReconciler struct {
 // +kubebuilder:rbac:groups="",resources=pods;services;endpoints;persistentvolumeclaims;events;configmaps;secrets,verbs=*
 // +kubebuilder:rbac:groups=apps,resources=deployments;daemonsets;replicasets;statefulsets,verbs=*
 // +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
-// +kubebuilder:rbac:groups=*,resources=*,verbs=get;list;watch;update
```
pedjak, line 52: do we really need to touch this here? IMHO, that should addressed in https://github.com/redhat-developer/service-binding-operator/issues/454

pedjak (changes requested):
we need acceptance test(s) checking that the added role is available in the cluster and that the rolebinding is established to the service account of the operator deployment.

controllers/servicebinding_controller.go
bamachrn: yes. make sense. I was thinking both the roles and access should go together.

test/acceptance/features/steps/steps.py
```diff
@@ -576,3 +576,37 @@ def update_binding_secret(context):
...
     result = re.search(rf'.*{context.sb_secret}.*(created|unchanged|configured)', output)
     assert result is not None, f"Unable to apply YAML for binding secret '{context.sb_secret}': {output}"
+
+
+@then(u'cluster role "{cluster_role_name}" is available in the cluster')
+def check_cluster_role_exists(context, cluster_role_name):
+    openshift = Openshift()
+    roles_present, exit_code = openshift.is_resource_in("clusterroles")
+    if exit_code==0:
+        role_list = json.load(roles_present)
```
pmacik, line 586: `roles_present` is actually a string containig output of `kubectl get clusterroles` - see the `openshift.is_role_in` function.
Furthermore, `json.load` expects a file not a string and so this line causes the acceptance tests to [fail](https://github.com/redhat-developer/service-binding-operator/pull/979/checks?check_run_id=3042889774#step:8:123) with the following:
```
Scenario: Service Binding Operator installation should contain aggregating cluster role and are bound with Namespace.  # test/acceptance/features/addRBACForOperator.feature:2
    Given Namespace [TEST_NAMESPACE] is used                                                                             # test/acceptance/features/steps/steps.py:43
    When Service Binding Operator is running                                                                             # test/acceptance/features/steps/steps.py:83
    Then cluster role "servicebinding-controller-role" is available in the cluster                                       # test/acceptance/features/steps/steps.py:590
      Traceback (most recent call last):
        File "/home/runner/work/service-binding-operator/service-binding-operator/out/venv3/lib/python3.7/site-packages/behave/model.py", line 1329, in run
          match.run(runner.context)
        File "/home/runner/work/service-binding-operator/service-binding-operator/out/venv3/lib/python3.7/site-packages/behave/matchers.py", line 98, in run
          self.func(context, *args, **kwargs)
        File "test/acceptance/features/steps/steps.py", line 595, in check_cluster_role_exists
          role_list = json.load(roles_present)
        File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/json/__init__.py", line 293, in load
          return loads(fp.read(),
      AttributeError: 'str' object has no attribute 'read'
      
      Captured stdout:
      TEST_NAMESPACE = test-namespace-772a9d798240
      ,---------,-
      | COMMAND : kubectl get ns test-namespace-772a9d798240
      '---------'-
      Namespace test-namespace-772a9d798240 is created!!!
      ,---------,-
      | COMMAND : kubectl get servicebinding
      '---------'-
      ,---------,-
      | COMMAND : kubectl get deployments --all-namespaces -o json | jq -rc '.items[] | select(.metadata.name == "service-binding-operator").metadata.namespace'
      '---------'-
      Service binding operator is running!!!
      ,---------,-
      | COMMAND : kubectl get clusterroles
      '---------'-
```

test/acceptance/features/steps/steps.py
```diff
@@ -576,3 +576,37 @@ def update_binding_secret(context):
...
+def check_cluster_role_exists(context, cluster_role_name):
+    openshift = Openshift()
+    roles_present, exit_code = openshift.is_resource_in("clusterroles")
+    if exit_code==0:
+        if cluster_role_name in roles_present:
+            assert True, "Found the cluster role in the cluster"
+        else:
+            assert False, "Could not find the cluster roler '{cluster_role_name}'"
+    else:
+        assert False, "Cluster role does not exist in the cluster"
```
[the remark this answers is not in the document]
pmacik: Please use asserts properly (https://www.w3schools.com/python/ref_keyword_assert.asp)

test/acceptance/features/steps/steps.py
```diff
@@ -576,3 +576,37 @@ def update_binding_secret(context):
...
+           subjects_list = rolebinding_json.subjects
+           for i in len(subjects_list):
+                if subjects_list[i].kind == "ServiceAccount" and subjects_list[i].name == "servicebinding-operator":
+                   assert True
+                else:
+                    continue
+        else:
+            assert False, "Could not find rolebing for the role '{cluster_role_name}'"
+    else:
+        assert False, "Could not find the rolebinding in the cluster"
```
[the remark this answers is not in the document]
pmacik: Please use asserts properly (https://www.w3schools.com/python/ref_keyword_assert.asp)

test/acceptance/features/addRBACForOperator.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Making Service Biniding Operator RBAC pluggable so that other controllers/admins can add additional rules for the service binding operator.
+    @rbactest
```
pedjak, line 2: do we need an extra annotation here?

test/acceptance/features/steps/openshift.py
```diff
@@ -146,7 +146,7 @@ def search_resource_in_namespace(self, resource_plural, name_pattern, namespace)
 
     def is_resource_in(self, resource_type):
         output, exit_code = self.cmd.run(f'{ctx.cli} get {resource_type}')
-        return exit_code == 0
+        return (output, exit_code)
```
pedjak, line 149: we changed here fully return values, so `is_resource_in` is not a proper name anymore.. btw, why do we need to change the return value?

test/acceptance/features/steps/openshift.py
```diff
@@ -300,6 +300,10 @@ def get_resource_info_by_jsonpath(self, resource_type, name, namespace, json_pat
             print(f'Error getting value for {resource_type}/{name} in {namespace} path={json_path}: {output}')
             return None
 
+    def get_json_resource(self, resource_type, name, namespace):
+        output, exit_code = self.cmd.run(f'{ctx.cli} get {resource_type} {name} -n {namespace} -o json')
+        return output, exit_code
```
pedjak, line 305: why we are returning exit code? also, what we return here is just raw representation, perhaps we could already parse it and return json data structure?

I would definitely assert exist for 0, not return it to the caller.

test/acceptance/features/steps/servicebindingoperator.py
```diff
@@ -16,7 +16,9 @@ def __init__(self,  name="service-binding-operator", namespace="openshift-operat
 
     def check_crd(self):
         crd_name = "servicebinding"
-        assert self.openshift.is_resource_in(crd_name), f"CRD '{crd_name}' does not exist"
```
pedjak, line 19: the removed line what doing exactly what the added lines are doing now. What is the motivation for the change?

test/acceptance/features/steps/steps.py
```diff
@@ -436,3 +437,34 @@ def update_binding_secret(context):
     output = openshift.apply(yaml.dump(secret_yaml), context.namespace.name)
     result = re.search(rf'.*{context.sb_secret}.*(created|unchanged|configured)', output)
     assert result is not None, f"Unable to apply YAML for binding secret '{context.sb_secret}': {output}"
+
+
+@then(u'cluster role "{cluster_role_name}" is available in the cluster')
+def check_cluster_role_exists(context, cluster_role_name):
+    openshift = Openshift()
+    roles_present, exit_code = openshift.is_resource_in("clusterroles")
```
pedjak, line 445: instead of getting all cluster roles, and then checking if `cluster_role_name` is present, we can try to get read directly `cluster_role_name`. If not found, we are going to get exit code different from 0.

you can use here `openshift.get_json_resource` instead.

test/acceptance/features/steps/steps.py
```diff
@@ -436,3 +437,34 @@ def update_binding_secret(context):
...
+    else:
+        assert False, "Cluster role does not exist in the cluster"
+
+
+@then(u'operator service account is bound to "{cluster_role_name}" cluster role')
+def check_service_account_bound(context, cluster_role_name):
+    openshift = Openshift()
+    sbr = Servicebindingoperator()
+    operator_namespace = sbr.check_deployment()
+    (rolebinding, exitcode) = openshift.get_json_resource("clusterrolebinding", "service-binding-operator-controller-rolebinding", operator_namespace)
```
pedjak, line 457: see my earlier comment - the code can be simplified if assert exit code in `get_json_resource`

test/acceptance/features/steps/steps.py
```diff
@@ -436,3 +437,34 @@ def update_binding_secret(context):
...
+    if exitcode == 0:
+        rolebinding_json = json.loads(rolebinding)
+        if rolebinding_json["roleRef"]["name"] == cluster_role_name:
+            subjects_list = rolebinding_json["subjects"]
+            for subject in subjects_list:
+                if subject["kind"] == "ServiceAccount" and subject["name"] == "service-binding-operator":
```
pedjak, line 463: we should not hard code service account to `service-binding-operator`, instead we should ready from SBO deployment what is the service account in use and check against it.

pedjak (changes requested):

test/acceptance/features/addRBACForOperator.feature
bamachrn: no this was for running the tests locally. reverting

test/acceptance/features/steps/openshift.py
bamachrn: this function checks if the resource type exists in the cluster. Here we want to make sure if a specific resource exists in the cluster irrespective of the namespace. I suppose we can use resource name directly here without changing the return value as well.

test/acceptance/features/steps/openshift.py
bamachrn: Here we are trying to handle use cases where there are multiple entries for a specific resource type.
Like when we are checking for the `rolebindings` we can have multiple rules/ subjects for a single rolebinding. With the exit code we can make sure if the resource is present in the cluster. and with the json output we can check for different values in the step implementation.

test/acceptance/features/steps/servicebindingoperator.py
bamachrn: we changed the return type for the `is_resource_in` function. that's why updated this function to assert on the proper value.

test/acceptance/features/steps/steps.py
bamachrn: ack

config/rbac/servicebinding_controller_rolebinding.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: controller-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: service-binding-operator-controller-role
```
baijum, line 8: Where is this ClusterRole defined?

bamachrn: we are adding this ClusterRole as an aggregator ClusterRole, to make it pluggable for other controllers to add additional rules. We are adding this to manifests here: https://github.com/redhat-developer/service-binding-operator/pull/979/files#diff-3fd064f2f1027a954b035a6d5885b96cc5b483f04889d14fa840ace5e9c29c55R8-R9

config/rbac/servicebinding_controller_role.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: controller-role
```
baijum, line 4: @bamachrn So, this should be changed to `service-binding-operator-controller-role`, right?

bamachrn: yes

bamachrn: my bad no, this gets prefixed with `service-binding-operator` by kustomaization while generating the bundle

config/rbac/servicebinding_controller_rolebinding.yaml
pedjak: but should not be the task of `kustomize` to prepend `service-binding-operator` prefix to this name?

test/acceptance/features/addRBACForOperator.feature
```diff
@@ -0,0 +1,6 @@
+Feature: Making Service Biniding Operator RBAC pluggable so that other controllers/admins can add additional rules for the service binding operator.
```
pedjak, line 1: can we name `.feature` file to reflect this - e.g. `pluggableRBAC.feature`?

test/acceptance/features/steps/openshift.py
```diff
@@ -300,6 +305,18 @@ def get_resource_info_by_jsonpath(self, resource_type, name, namespace, json_pat
             print(f'Error getting value for {resource_type}/{name} in {namespace} path={json_path}: {output}')
             return None
 
+    def get_json_resource(self, resource_type, name, namespace):
+        error_msg = f"Error in getting resource: '{resource_type}' '{name}' namespace: '{namespace}'"
+        output, exit_code = self.cmd.run(f'{ctx.cli} get {resource_type} {name} -n {namespace} -o json')
+        assert exit_code == 0, error_msg
+        json_output = None
```
pedjak, line 312: you can remove this line.

test/acceptance/features/steps/openshift.py
```diff
@@ -300,6 +305,18 @@ def get_resource_info_by_jsonpath(self, resource_type, name, namespace, json_pat
...
+        try:
+            json_output = json.loads(output)
+        except JSONDecodeError:
```
pedjak, line 315: we do not need to catch `JSONDecodeError` to throw another exception here. It is enough to do:

```python
json_output = json.loads(output)
assert json_output is not None, "Error parsing as JSON"
return json_output
```

config/rbac/servicebinding_controller_rolebinding.yaml
bamachrn: kustomization only changes `.metadata.name` not in the roleRef

test/acceptance/features/addRBACForOperator.feature
bamachrn: yeah makes sense

pedjak (approved):

PR #980, opened 2021-06-16
[acceptance-tests] Collect kube resources as well as SBO logs in GH workflows

pmacik:
### Motivation

Currently, only SBO logs are collected for debugging.

### Changes

This PR:
* Adds steps to acceptance testing PR checks to collect additional resources such as Subscription, CatalogSource, Pods and Deployments related to SBO for debugging purposes

### Testing

PR checks will do.

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -107,11 +107,14 @@ jobs:
 
           make SKIP_REGISTRY_LOGIN=true -o registry-login test-acceptance-with-bundle
 
-      - name: Collect logs
+      - name: Collect Kube resources
         continue-on-error: true
-        run: |
-          mkdir -p ${TEST_RESULTS}
-          kubectl logs $(kubectl get pods --no-headers -o custom-columns=":metadata.name" -n operators | tail -1) -n operators > ${TEST_RESULTS}/sbo.log
+        uses: ./.github/actions/collect-kube-resources
+        with:
+          operator-namespace: operators
+          olm-namespace: olm
+          test-namespace-file: out/test-namespace
+          output-path: ${{env.TEST_RESULTS}}
         if: always()
 
       - name: Setup Testspace
@@ -184,11 +187,13 @@ jobs:
           kubectl apply -f out/release.yaml
           make TEST_ACCEPTANCE_START_SBO=remote test-acceptance
 
-      - name: Collect logs
+      - name: Collect Kube resources
         continue-on-error: true
-        run: |
-          mkdir -p ${TEST_RESULTS}
-          kubectl logs $(kubectl get pods --no-headers -o custom-columns=":metadata.name" -n operators | tail -1) -n operators > ${TEST_RESULTS}/sbo.log
+        uses: ./.github/actions/collect-kube-resources
+        with:
+          operator-namespace: service-binding-operator
+          test-namespace-file: out/test-namespace
+          output-path: ${{env.TEST_RESULTS}}
         if: always()
 
       - name: Setup Testspace
--- a/.github/actions/collect-kube-resources/action.yaml
+++ b/.github/actions/collect-kube-resources/action.yaml
@@ -1,0 +1,53 @@
+# action.yml
+name: 'Collect Kube resources from Kubernetes cluster'
+description: 'Collects resources for later debugging from OLM operator testing.'
+inputs:
+  operator-namespace:
+    description: "Operator namespace"
+    required: false
+    default: ''
+  olm-namespace:
+    description: "OLM namespace"
+    required: false
+    default: ''
+  test-namespace-file:
+    description: "File with test namespace"
+    required: false
+    default: 'out/test-namespace'
+  output-path:
+    description: "A path to directory under which to store output files"
+    required: false
+    default: '.'
+runs:
+  using: "composite"
+  steps:
+    - id: collect-kube-resources
+      run: |
+        kubectl api-resources --verbs=list --namespaced -o name > resources.list
+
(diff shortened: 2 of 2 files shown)


.github/workflows/pr-checks.yaml
```diff
@@ -107,6 +107,21 @@ jobs:
 
           make SKIP_REGISTRY_LOGIN=true -o registry-login test-acceptance-with-bundle
 
+      - name: Collect Kube resources
+        continue-on-error: true
+        run: |
+          mkdir -p ${TEST_RESULTS}
+          ns=operators
+          for res in pods deployments subscriptions; do
```
pedjak, line 115: how about that we do not list just a few things here, instead we can collect really all resources with:

```
kubectl get all -n $ns -o yaml > ${TEST_RESULTS}/resources.${ns}.yaml
```

pedjak (commented):
apart from this PR, we need the counterpart PR for openshift CI jobs.

.github/workflows/pr-checks.yaml
pmacik: Good ides, I'll update it. Thx @pedjak

pmacik: @pedjak `kubectl get all` does not return all resources in the namespace, only selected ones such as `Pod`, `Service`, `Deployment`, `ReplicaSet`, `Job`. No custom resources including `Subscription` or `CatalogSource`.

To get really all resources in namespace, we should to something like [this](https://stackoverflow.com/questions/47691479/listing-all-resources-in-a-namespace):

```bash
ns=operators
for res in $(kubectl api-resources --verbs=list --namespaced -o name); do
   kubectl get ${res} --ignore-not-found -n ${ns} -o yaml > ${TEST_RESULTS}/${res}.${ns}.yaml;
done                                    
```

pmacik: I'm just a bit concerned whether we would expose any sensitive data (stored in `Secret` resources) that way.

pedjak: do we really have any sensitive data in our tests/test clusters?

pmacik: Well, I meant in general.
For example the `install.sh` script creates a secret `sbo-operator-secrets` with a docker config which is internal for the GH workflow. And to be honest, I'm not sure what is inside :)

pmacik: Other then that, I don't see any possibly sensitive data in secrets from tests. Any other idea?

pmacik: I just want to make sure that we do not expose anything we shouln't.

pedjak: well, you can filter out secrets - what I am asking myself is what kind of insights we can get by collecting these resources anyway?

pmacik: Well, those are meant for the case of a installation/deploment failure - to investigate what went wrong and those resources contain additional information that can help to underatand the failure - especially SBO's Pod and Deployment, as well as Subscription and CatalogSource, that is why those were in the orinal PR.

pmacik: to investigate tests we would need to collect resources from the test namespace, anyway

pedjak: +1 on collecting resources from the test namespace

.github/workflows/pr-checks.yaml
```diff
@@ -107,6 +107,27 @@ jobs:
...
           make SKIP_REGISTRY_LOGIN=true -o registry-login test-acceptance-with-bundle
 
+      - name: Collect Kube resources
+        continue-on-error: true
+        run: |
+          kubectl api-resources --verbs=list --namespaced -o name > resources.list
+
+          for ns in olm operators; do
+            for res in $(cat resources.list | grep -v secrets); do
+              mkdir -p ${TEST_RESULTS}/${ns}
```
pedjak, line 117: move out this command into outerloop

.github/workflows/pr-checks.yaml
```diff
@@ -107,6 +107,27 @@ jobs:
...
+              kubectl get ${res} --ignore-not-found -n ${ns} -o yaml > ${TEST_RESULTS}/${ns}/${res}.yaml;
+            done
+          done
+
+          if [ -f out/test-namespace ]; then
+            ns=$(cat out/test-namespace)
+            for res in $(cat resources.list); do
+              mkdir -p ${TEST_RESULTS}/${ns}
```
pedjak, line 125: same as above

.github/workflows/pr-checks.yaml
```diff
@@ -184,6 +205,26 @@ jobs:
...
+              kubectl get ${res} --ignore-not-found -n ${ns} -o yaml > ${TEST_RESULTS}/${ns}/${res}.yaml;
+            done
+          fi
+        if: always()
+
```
pedjak, line 227: how about that we make a custom action, so to avoid having this code twice in the workflows?

pedjak (changes requested):

.github/workflows/pr-checks.yaml
pmacik: +1

pmacik: +1

.github/workflows/pr-checks.yaml
pmacik: +1

.github/workflows/pr-checks.yaml
pmacik: +1

pmacik:
The OpenShift CI jobs failing for infrastructure issues - re-running them

PR #981, opened 2021-06-17
Run goimports across all files

baijum:
### Motivation

Using [goimports](https://github.com/golang/go/wiki/CodeReviewComments#gofmt) is a best practice in Go community.

### DIFF
--- a/controllers/servicebinding_controller.go
+++ b/controllers/servicebinding_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	ctx "context"
+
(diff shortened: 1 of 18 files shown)


pedjak (changes requested):
It is good to agree about some common formatting rule, but we need to introduce some lint check in this PR that is going to enforce it. Otherwise, after a while, we are going drift.

Please hook the relevant check to `lint` make target. Also, please state in PR summary what it contains, not what tool was used for that.

cc @pmacik

PR #984, opened 2021-06-22
Add support for Direct Secret reference

### LINKED ISSUE #872: Direct Secret reference
I'm trying to create a binding between Deployment and Secret as described in https://github.com/k8s-service-bindings/spec#direct-secret-reference

I have a secret with passwords for my database
```
▶ kubectl describe secrets my-mysql
Name:         my-mysql
Namespace:    tkral-test
Labels:       app.kubernetes.io/instance=my-mysql
              app.kubernetes.io/managed-by=Helm
              app.kubernetes.io/name=mysql
              helm.sh/chart=mysql-8.3.1
Annotations:  meta.helm.sh/release-name: my-mysql
              meta.helm.sh/release-namespace: tkral-test

Type:  Opaque

Data
====
mysql-password:       10 bytes
mysql-root-password:  10 bytes
```

I'm trying to reference this  `Secret` directly in `ServiceBinding` as described in https://github.com/k8s-service-bindings/spec#direct-secret-reference
```
▶ kubectl get servicebindings.operators.coreos.com mysql-binding  -o yaml
```
```yaml
kind: ServiceBinding
metadata:
  name: mysql-binding
  namespace: tkral-test
spec:
  application:
    group: apps
    name: python-mysql
    resource: deployments
    version: v1
  detectBindingResources: true
  services:
  - group: ""
    kind: Secret
    name: my-mysql
    version: v1
status:
  conditions:
  - lastHeartbeatTime: "2021-02-12T16:54:40Z"
    lastTransitionTime: "2021-02-12T16:52:34Z"
    status: "True"
    type: CollectionReady
  - lastHeartbeatTime: "2021-02-12T16:54:40Z"
    lastTransitionTime: "2021-02-12T16:52:34Z"
    status: "True"
    type: InjectionReady
  - lastHeartbeatTime: "2021-02-12T16:54:40Z"
    lastTransitionTime: "2021-02-12T16:52:34Z"
    status: "True"
    type: Ready
  secret: mysql-binding
```

But Secret that was created by `ServiceBinding` is empty
```
▶ kubectl describe secret mysql-binding
Name:         mysql-binding
Namespace:    tkral-test
Labels:       <none>
Annotations:  <none>

Type:  Opaque

Data
====
```

I'm using Service Binding Operator v0.4.0

baijum:
Fixes #872

### DIFF
--- a/pkg/reconcile/pipeline/handler/collect/impl.go
+++ b/pkg/reconcile/pipeline/handler/collect/impl.go
@@ -4,6 +4,9 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"reflect"
+	"strings"
+
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/redhat-developer/service-binding-operator/api/v1alpha1"
 	"github.com/redhat-developer/service-binding-operator/pkg/binding"
@@ -12,8 +15,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"reflect"
-	"strings"
 )
 
 var DataNotMap = errors.New("Returned data are not a map, skip collecting")
@@ -127,6 +128,31 @@ func ProvisionedService(ctx pipeline.Context) {
 				requestRetry(ctx, ErrorReadingBindingReason, fmt.Errorf("CRD of service %v/%v indicates provisioned service, but no secret name provided under .status.binding.name", res.GetNamespace(), res.GetName()))
 				return
 			}
+		}
+	}
+}
+
+func DirectSecretReference(ctx pipeline.Context) {
+	// Error is ignored as this check is there in the PreFlight stage.
+	// That stage was created to perform common checks for all followup stages.
+	services, _ := ctx.Services()
+
+	for _, service := range services {
+		res := service.Resource()
+		if res.GetKind() == "Secret" && res.GetAPIVersion() == "v1" && res.GroupVersionKind().Group == "" {
+			annotations := res.GetAnnotations()
+			for k := range annotations {
+				if strings.HasPrefix(k, binding.AnnotationPrefix) {
+					return
+				}
+			}
+			name := res.GetName()
+			secret, err := ctx.ReadSecret(res.GetNamespace(), name)
+			if err != nil {
+				requestRetry(ctx, ErrorReadingSecret, err)
+				return
+			}
+			ctx.AddBindings(&pipeline.SecretBackedBindings{Service: service, Secret: secret})
 		}
 	}
 }
--- a/pkg/reconcile/pipeline/handler/collect/impl_test.go
+++ b/pkg/reconcile/pipeline/handler/collect/impl_test.go
@@ -4,6 +4,8 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"reflect"
+
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
@@ -15,7 +17,6 @@ import (
 	bindingmocks "github.com/redhat-developer/service-binding-operator/pkg/binding/mocks"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"reflect"
 
 	"github.com/redhat-developer/service-binding-operator/pkg/reconcile/pipeline"
 	"github.com/redhat-developer/service-binding-operator/pkg/reconcile/pipeline/handler/collect"
@@ -745,6 +746,105 @@ var _ = Describe("Collect From Provisioned Service", func() {
 		})
 	})
 
+})
+
+var _ = Describe("Collect From Direct Secret", func() {
+	BeforeEach(func() {
+		mockCtrl = gomock.NewController(GinkgoT())
+		ctx = mocks.NewMockContext(mockCtrl)
+	})
+
+	AfterEach(func() {
+		mockCtrl.Finish()
+	})
+
+	Describe("successful processing", func() {
+		var (
+			services []pipeline.Service
+
+			defService = func() (*mocks.MockService, *unstructured.Unstructured) {
+				service := mocks.NewMockService(mockCtrl)
+				serviceContent := &unstructured.Unstructured{}
+				service.EXPECT().Resource().Return(serviceContent)
+				services = append(services, service)
+				return service, serviceContent
+			}
+			err = errors.New("e")
+		)
+
+		BeforeEach(func() {
+			services = []pipeline.Service{}
+			ctx.EXPECT().Services().DoAndReturn(func() ([]pipeline.Service, error) { return services, nil })
+		})
+
+		It("should collect from secret referred in services", func() {
+			secretName1 := "foo"
+			secretName2 := "bar"
+			ns1 := "ns1"
+			ns2 := "ns2"
+			svc1, content1 := defService()
+			content1.Object = map[string]interface{}{
+				"status": map[string]interface{}{
+					"binding": map[string]interface{}{
+						"name": secretName1,
+					},
+				},
+			}
+			content1.SetNamespace(ns1)
+			content1.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Secret"})
+			content1.SetName(secretName1)
+
+			svc2, content2 := defService()
+			content2.Object = map[string]interface{}{
+				"status": map[string]interface{}{
+					"binding": map[string]interface{}{
+						"name": secretName2,
+					},
+				},
+			}
+			content2.SetNamespace(ns2)
+			content2.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Secret"})
+			content2.SetName(secretName2)
+
+			ctx.EXPECT().ReadSecret(ns1, secretName1).Return(content1, nil)
+			ctx.EXPECT().ReadSecret(ns2, secretName2).Return(content2, nil)
+
+			ctx.EXPECT().AddBindings(&pipeline.SecretBackedBindings{Service: svc1, Secret: content1})
+			ctx.EXPECT().AddBindings(&pipeline.SecretBackedBindings{Service: svc2, Secret: content2})
+
+			collect.DirectSecretReference(ctx)
+		})
+
+		It("should retry processing if reading secret fails", func() {
+			_, content := defService()
+			secretName1 := "foo"
+			ns1 := "ns1"
+			content.SetName(secretName1)
+			content.SetNamespace(ns1)
+			content.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Secret"})
+			content.SetName(secretName1)
+
+			ctx.EXPECT().ReadSecret(ns1, secretName1).Return(nil, err)
+
+			ctx.EXPECT().RetryProcessing(err)
+			ctx.EXPECT().SetCondition(v1alpha1.Conditions().NotCollectionReady().Reason(collect.ErrorReadingSecret).Msg(err.Error()).Build())
+			collect.DirectSecretReference(ctx)
+		})
+
+		It("ignore secret having a secret binding annotation", func() {
+			_, content := defService()
+			secretName1 := "foo"
+			ns1 := "ns1"
+			ann := map[string]string{"service.binding": "path={.data},elementType=map"}
+			content.SetAnnotations(ann)
+			content.SetName(secretName1)
+			content.SetNamespace(ns1)
+			content.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Secret"})
+			content.SetName(secretName1)
+			collect.DirectSecretReference(ctx)
+		})
+
+	})
 })
 
 var _ = Describe("Integration Collect definitions + items", func() {
--- a/test/acceptance/features/bindAppToServiceUsingSecret.feature
+++ b/test/acceptance/features/bindAppToServiceUsingSecret.feature
@@ -759,3 +759,202 @@ Feature: Bind values from a secret referred in backing service resource
             """
             bla
             """
+
+    Scenario: Inject binding to an application from a Secret resource referred as service
+        Given The Secret is present
+            """
+            apiVersion: v1
+            kind: Secret
+            metadata:
+                name: provisioned-secret-1
+            stringData:
+                username: foo
+                password: bar
+            """
+        * Generic test application "myaop-provision-srv" is running
+        When Service Binding is applied
+          """
+          apiVersion: binding.operators.coreos.com/v1alpha1
+          kind: ServiceBinding
+          metadata:
+              name: bind-provisioned-service-1
+          spec:
+              services:
+              - group: ""
+                version: v1
+                kind: Secret
+                name: provisioned-secret-1
+              application:
+                name: myaop-provision-srv
+                group: apps
+                version: v1
+                resource: deployments
+          """
+        Then Service Binding "bind-provisioned-service-1" is ready
+        And jq ".status.secret" of Service Binding "bind-provisioned-service-1" should be changed to "provisioned-secret-1"
+        And Content of file "/bindings/bind-provisioned-service-1/username" in application pod is
+            """
+            foo
+            """
+        And Content of file "/bindings/bind-provisioned-service-1/password" in application pod is
+            """
+            bar
+            """
+
+    Scenario: Inject binding to an application from a Secret resource referred as service with mappings
+        Given The Secret is present
+            """
+            apiVersion: v1
+            kind: Secret
+            metadata:
+                name: provisioned-secret-1
+            stringData:
+                username: foo
+                password: bar
+            """
+        * Generic test application "myaop-provision-srv" is running
+        When Service Binding is applied
+          """
+          apiVersion: binding.operators.coreos.com/v1alpha1
+          kind: ServiceBinding
+          metadata:
+              name: bind-provisioned-service-1
+          spec:
+              bindAsFiles: true
+              services:
+              - group: ""
+                version: v1
+                kind: Secret
+                name: provisioned-secret-1
+                id: sec
+              application:
+                name: myaop-provision-srv
+                group: apps
+                version: v1
+                resource: deployments
+              mappings:
+                - name: username_with_password
+                  value: '{{ .username }}:{{ .password }}'
+          """
+        Then Service Binding "bind-provisioned-service-1" is ready
+        And Service Binding "bind-provisioned-service-1" has the binding secret name set in the status
+        And Content of file "/bindings/bind-provisioned-service-1/username" in application pod is
+            """
+            foo
+            """
+        And Content of file "/bindings/bind-provisioned-service-1/password" in application pod is
+            """
+            bar
+            """
+        And Content of file "/bindings/bind-provisioned-service-1/username_with_password" in application pod is
+            """
+            foo:bar
+            """
+
+    Scenario: Inject binding to an application from a Secret resource created later referred as service 
+        Given Generic test application "myaop-provision-srv" is running
+        When Service Binding is applied
+          """
+          apiVersion: binding.operators.coreos.com/v1alpha1
+          kind: ServiceBinding
+          metadata:
+              name: bind-provisioned-service-4
+          spec:
+              services:
+              - group: ""
+                version: v1
+                kind: Secret
+                name: provisioned-secret-4
+              application:
+                name: myaop-provision-srv
+                group: apps
+                version: v1
+                resource: deployments
+          """
+        * jq ".status.conditions[] | select(.type=="Ready").status" of Service Binding "bind-provisioned-service-4" should be changed to "False"
+        * The Secret is present
+            """
+            apiVersion: v1
+            kind: Secret
+            metadata:
+                name: provisioned-secret-4
+            stringData:
+                username: foo
+                password: bar
+            """
+
+        Then Service Binding "bind-provisioned-service-4" is ready
+        And Content of file "/bindings/bind-provisioned-service-4/username" in application pod is
+            """
+            foo
+            """
+        And Content of file "/bindings/bind-provisioned-service-4/password" in application pod is
+            """
+            bar
+            """
+
+    Scenario: Inject binding to an application from two Secret resources referred as services
+        Given Generic test application "myaop-provision-srv" is running
+        When Service Binding is applied
+          """
+          apiVersion: binding.operators.coreos.com/v1alpha1
+          kind: ServiceBinding
+          metadata:
+              name: bind-provisioned-service-1
+          spec:
+              services:
+              - group: ""
+                version: v1
+                kind: Secret
+                name: provisioned-secret-1
+                id: one
+              - group: ""
+                version: v1
+                kind: Secret
+                name: provisioned-secret-2
+                id: two
+              application:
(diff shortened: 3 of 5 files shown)


pkg/reconcile/pipeline/context/service.go
```diff
@@ -49,6 +50,17 @@ type service struct {
 	bindingDefinitions    []binding.Definition
 }
 
+func (s *service) IsDirectSecretReference() bool {
+	if s.serviceRef.Kind == "Secret" && s.serviceRef.Version == "v1" {
```
baijum, line 54: group should be empty

pkg/reconcile/pipeline/handler/collect/impl.go
```diff
@@ -131,6 +132,22 @@ func ProvisionedService(ctx pipeline.Context) {
 	}
 }
 
+func DirectSecretReference(ctx pipeline.Context) {
+	services, _ := ctx.Services()
+
+	for _, service := range services {
+		if service.IsDirectSecretReference() {
+			res := service.Resource()
```
baijum, line 140: Extract secret directly from the object

pkg/reconcile/pipeline/handler/collect/impl.go
```diff
@@ -131,6 +132,22 @@ func ProvisionedService(ctx pipeline.Context) {
 	}
 }
 
+func DirectSecretReference(ctx pipeline.Context) {
```
baijum, line 135: Unit test for this function

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -761,3 +761,44 @@ Feature: Bind values from a secret referred in backing service resource
             """
             bla
             """
+
+    Scenario: Bind application through direct secret reference
```
baijum, line 765: Inject binding to an application from a Secret resource referred as service

baijum: done. d11782fdf7ece95cffe1f7ac009f10850ef4964b

pkg/reconcile/pipeline/handler/collect/impl.go
baijum: done. d11782fdf7ece95cffe1f7ac009f10850ef4964b

pkg/reconcile/pipeline/context/service.go
baijum: Done. 1bf562899688f37d3b28faaa2b8ca8c680441719

baijum: I had to change commit because if DCO.  This is the new commit: 273c29d7b6ae20431f7c54d534e4ecddcdb16b8e

pkg/reconcile/pipeline/handler/collect/impl.go
```diff
@@ -131,6 +132,23 @@ func ProvisionedService(ctx pipeline.Context) {
...
+	services, _ := ctx.Services()
```
isutton, line 136: I see `ctx.Services()` implementation might return errors, any special reason this is being ignored?

pkg/reconcile/pipeline/handler/collect/impl.go
```diff
@@ -131,6 +132,23 @@ func ProvisionedService(ctx pipeline.Context) {
...
+
+	for _, service := range services {
+		res := service.Resource()
+		if res.GetKind() == "Secret" && res.GetAPIVersion() == "v1" && res.GroupVersionKind().Group == "" {
+			name := res.GetName()
+			secret, err := ctx.ReadSecret(res.GetNamespace(), name)
+			if err != nil {
+				requestRetry(ctx, ErrorReadingSecret, err)
```
isutton, line 144: The request will be retried infinitely in the case the secret does not exist; `err` must be checked and retries shouldn't be requested in unrecoverable cases.

baijum: What if someone creates a secret after the ServiceBinding resource creation?
I can see we have similar logic in ProvisionedService also. line 111-112.

pkg/reconcile/pipeline/handler/collect/impl.go
baijum: Fixed. 58aea3667d70a507d6f3f32d6ea0dda8d26cb465

pkg/reconcile/pipeline/handler/collect/impl.go
isutton: Kubernetes offers mechanisms to react on creation, update and deletion of resources to avoid situations such as the indicated. The appropriated approach would be a) fail without retry if a Secret is not found at the time of the reconciliation and b) observe the life-cycle (either fully or partially) of Secrets to perform the reconciliation of the related ServiceBinding.

pkg/reconcile/pipeline/handler/collect/impl.go
```diff
@@ -131,6 +132,27 @@ func ProvisionedService(ctx pipeline.Context) {
 	}
 }
 
+func DirectSecretReference(ctx pipeline.Context) {
+	services, err := ctx.Services()
+	if err != nil {
+		requestRetry(ctx, ErrorReadingSecret, err)
```
isutton, line 138: The error might not be related to reading a secret in this particular case.

pkg/reconcile/pipeline/handler/collect/impl.go
baijum: As I understand we removed all watches as part of optimization. So, I am not sure how to reconcile it.
As I pointed out, this design approach is there in other parts of our codebase.

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -761,3 +761,44 @@ Feature: Bind values from a secret referred in backing service resource
             """
             bla
             """
+
+    Scenario: Inject binding to an application from a Secret resource referred as service
```
isutton, line 765: Given that it is expected the operator to recover from situations where the referred Secret comes to existence after the ServiceBinding resource is created, it'd be nice to see a reproduction of this scenario in the acceptance test suite.

isutton (changes requested):

test/acceptance/features/bindAppToServiceUsingSecret.feature
baijum: Done. 494077eab6dd88dd5b5db0a7060bf41b78d9fddf

pkg/reconcile/pipeline/handler/collect/impl.go
baijum: Fixed. cf2da0c816f48b61546bb862397dace9fb581ec6

pkg/reconcile/pipeline/handler/collect/impl.go
baijum: Done. fee11f2015b720a1d5984d357d438f8853a40e43

isutton (approved):
LGTM

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -761,3 +761,86 @@ Feature: Bind values from a secret referred in backing service resource
...
+            kind: Secret
+            metadata:
+                name: provisioned-secret-1
+            stringData:
+                username: foo
+                password: bar
+            """
+
+        Then Service Binding "bind-provisioned-service-1" is ready
+        And jq ".status.secret" of Service Binding "bind-provisioned-service-1" should be changed to "provisioned-secret-1"
```
isutton, line 838: Is this the expected behavior?

Now I wonder how this works with multiple secrets, given the ServiceBinding contains only one `.status.secret` field.

I believe we need an acceptance test for that scenario as well.

isutton (changes requested):

test/acceptance/features/bindAppToServiceUsingSecret.feature
baijum: Updated the current test. a5aedb0121563ceed50f7e2210be631620a08ba3

pkg/reconcile/pipeline/handler/collect/impl.go
```diff
@@ -131,6 +132,27 @@ func ProvisionedService(ctx pipeline.Context) {
 	}
 }
 
+func DirectSecretReference(ctx pipeline.Context) {
+	services, err := ctx.Services()
+	if err != nil {
+		requestRetry(ctx, ErrorReadingServicesReason, err)
+		return
+	}
```
pedjak, line 140: You can omit this here, because we have this check already in PreFlight stage:

https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/reconcile/pipeline/handler/collect/impl.go#L29

This stage was created to perform common checks for all followup stages.

pkg/reconcile/pipeline/handler/collect/impl_test.go
```diff
@@ -745,6 +746,111 @@ var _ = Describe("Collect From Provisioned Service", func() {
...
+		})
+		It("should retry processing if service listing return error", func() {
+			ctx.EXPECT().RetryProcessing(err)
+			ctx.EXPECT().SetCondition(v1alpha1.Conditions().NotCollectionReady().Reason(collect.ErrorReadingServicesReason).Msg(err.Error()).Build())
+
+			collect.DirectSecretReference(ctx)
+		})
+
+	})
+})
```
pedjak, line 852: see above, this tests can be removed.

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -761,3 +761,110 @@ Feature: Bind values from a secret referred in backing service resource
...
+                kind: Secret
+                name: provisioned-secret-2
+                id: two
+              application:
+                name: myaop-provision-srv
+                group: apps
+                version: v1
+                resource: deployments
+          """
+        * The Secret is present
```
pedjak, line 832: In order to really validate this scenario, service binding must be detected as not ready before injecting secrets.

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -761,3 +761,110 @@ Feature: Bind values from a secret referred in backing service resource
...
+          metadata:
+              name: bind-provisioned-service-1
+          spec:
+              services:
+              - group: ""
+                version: v1
+                kind: Secret
+                name: provisioned-secret-1
+                id: one
+              - group: ""
```
pedjak, line 821: do we really need two services for this scenario? If not let's keep the setup minimal.

test/acceptance/features/steps/steps.py
```diff
@@ -249,7 +249,7 @@ def sbo_jq_is(context, jq_expression, sbr_name, json_value):
     openshift = Openshift()
     polling2.poll(lambda: json.loads(
         openshift.get_resource_info_by_jq("servicebinding", sbr_name, context.namespace.name, jq_expression,
-                                          wait=False)) == json_value, step=5, timeout=800,
+                                          wait=False)).startswith(json_value), step=5, timeout=800,
```
pedjak, line 252: this is going to result in a way more positive results - why is this change needed in this PR?

pedjak (changes requested):
we need  additional acceptance test demonstrating  additional bindings can be added on top of direct reference using `mappings`.

also, we support already using secret as reference, via annotations. Check this scenario:

https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToService.feature#L796

How we are going to differentiate? Should we treat a secret as direct reference only if it does not contain annotations?

baijum:
> How we are going to differentiate? Should we treat a secret as direct reference only if it does not contain annotations?

I am fine with this approach.  I can add check for the annotation.

test/acceptance/features/bindAppToServiceUsingSecret.feature
baijum: There was a comment from Igor about what happens if there are multiple Secret resources.
I tried to combine it. Looks like it's confusing, I will create a separate test case for that.

test/acceptance/features/steps/steps.py
baijum: I changed it to check the value of `.status.secret` when there are multiple Secret resources.
I will remove this change and create a separate step function for that requirement.

pedjak: Where there are multiple secret referred, then the binding secret will be generated as before, so no need to check anything new there.

pkg/reconcile/pipeline/handler/collect/impl_test.go
```diff
@@ -745,6 +746,111 @@ var _ = Describe("Collect From Provisioned Service", func() {
...
+		It("should retry processing if reading secret fails", func() {
+			_, content := defService()
+			secretName1 := "foo"
+			ns1 := "ns1"
+			content.SetName(secretName1)
+			content.SetNamespace(ns1)
+			content.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Secret"})
+			content.SetName(secretName1)
+
+			ctx.EXPECT().ReadSecret(ns1, secretName1).Return(content, err)
```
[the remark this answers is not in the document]
baijum: Done. 10cf41199e4e0a45796f32d33108c79969cac1f7

pkg/reconcile/pipeline/handler/collect/impl.go
baijum: Done. 10cf41199e4e0a45796f32d33108c79969cac1f7

pkg/reconcile/pipeline/handler/collect/impl_test.go
baijum: Done. 10cf41199e4e0a45796f32d33108c79969cac1f7

test/acceptance/features/bindAppToServiceUsingSecret.feature
baijum: Done. 10cf41199e4e0a45796f32d33108c79969cac1f7

test/acceptance/features/steps/steps.py
baijum: Yes, but there was no assertion for the `.status.secret` value which is dynamic in nature. Please take a look at the test that I added and let me know.

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -761,3 +761,202 @@ Feature: Bind values from a secret referred in backing service resource
...
+                name: myaop-provision-srv
+                group: apps
+                version: v1
+                resource: deployments
+              mappings:
+                - name: username_with_password
+                  value: '{{ .username }}:{{ .password }}'
+          """
+        Then Service Binding "bind-provisioned-service-1" is ready
+        And jq ".status.secret" of Service Binding "bind-provisioned-service-1" should starts with "bind-provisioned-service-1"
```
pedjak, line 842: we should not advertise this kind of info to users - it is not a part of api/contract - the spec only says that this field will be populated with the intermediate secret name, but does not say anything about naming convention.

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -761,3 +761,202 @@ Feature: Bind values from a secret referred in backing service resource
...
+            kind: Secret
+            metadata:
+                name: provisioned-secret-1
+            stringData:
+                username: foo
+                password: bar
+            """
+
+        Then Service Binding "bind-provisioned-service-4" is ready
+        And Content of file "/bindings/bind-provisioned-service-4/username" in application pod is
```
pedjak, line 889: we should have here the assert about the secret name in the status

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -761,3 +761,202 @@ Feature: Bind values from a secret referred in backing service resource
...
+            kind: Secret
+            metadata:
+                name: provisioned-secret-2
+            stringData:
+                username2: foo2
+                password2: bar2
+            """
+
+        Then Service Binding "bind-provisioned-service-1" is ready
+        And jq ".status.secret" of Service Binding "bind-provisioned-service-1" should starts with "bind-provisioned-service-1"
```
pedjak, line 946: we should not advertise this kind of info to users - it is not a part of api/contract - the spec only says that this field will be populated with the intermediate secret name, but does not say anything about naming convention.

baijum: See the last sentence in this section: https://github.com/k8s-service-bindings/spec#service-binding
It reads like this:
>  A Service Binding resource SHOULD reflect the secret projected into the application as `.status.binding.name`.

That's a recommendation.

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -761,3 +761,202 @@ Feature: Bind values from a secret referred in backing service resource
...
+                name: provisioned-secret-1
+                id: sec
+              application:
+                name: myaop-provision-srv
+                group: apps
+                version: v1
+                resource: deployments
+              mappings:
+                - name: username_with_password
+                  value: '{{ .username }}:{{ .password }}'
```
baijum, line 839: @pedjak Please let me know referring to Secret resource values like this is correct.
This seems to be working, but not sure the semantics is right.

When I tried to use the prefix with `data` (e.g., `.sec.data.username`) it gives the base64 encoded value. Is that also expected?

pedjak: Yes, both things work as expected.

test/acceptance/features/bindAppToServiceUsingSecret.feature
pedjak: I am not seeing any new line in this scenario checking that service binding is not ready.

test/acceptance/features/bindAppToServiceUsingSecret.feature
pedjak: it says only that the secret name should be set in `.status.binding.name`, but it does not force any particular naming convention.

test/acceptance/features/steps/steps.py
pedjak: in case when intermediate secret is generated and then injected, we should asset the following things:

* service binding Ready condition is true
* `.status.binding.name` is set, i.e. non empty
* the application sees injected env vars or files

The fact that intermediate secret name is made as `binding name` + `some hash` is an implementation detail and the user should not count on that - we can change that at any point in time.

pedjak:
@baijum https://github.com/redhat-developer/service-binding-operator/pull/984#pullrequestreview-691947110 is still not addressed.

test/acceptance/features/steps/steps.py
baijum: Done. 832c72fcec87d9f227ff5ea071ff759a112edc95

test/acceptance/features/bindAppToServiceUsingSecret.feature
baijum: Done. 832c72fcec87d9f227ff5ea071ff759a112edc95

baijum:
> also, we support already using secret as reference, via annotations. Check this scenario:
> 
> https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToService.feature#L796
> 
> How we are going to differentiate? Should we treat a secret as direct reference only if it does not contain annotations?

@pedjak I have implemented it here:
7903e02544667fd362e723947b55f9c5859131c8

test/acceptance/features/bindAppToServiceUsingSecret.feature
baijum: This was done. See line no. 876

test/acceptance/features/bindAppToServiceUsingSecret.feature
baijum: Done. 832c72fcec87d9f227ff5ea071ff759a112edc95

test/acceptance/features/bindAppToServiceUsingSecret.feature
baijum: Please check line no. 876
Let me know if that is sufficient.

baijum:
> > also, we support already using secret as reference, via annotations. Check this scenario:
> > https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/bindAppToService.feature#L796
> > How we are going to differentiate? Should we treat a secret as direct reference only if it does not contain annotations?
> 
> @pedjak I have implemented it here:
> [7903e02](https://github.com/redhat-developer/service-binding-operator/commit/7903e02544667fd362e723947b55f9c5859131c8)

I think no need to retry. Then annotations will not work. Changed the approach here:
07f2b439dd1f79bd91721855e7db192b8f0bd093

pkg/reconcile/pipeline/handler/collect/impl.go
```diff
@@ -131,6 +132,31 @@ func ProvisionedService(ctx pipeline.Context) {
...
+	// Error is ignored as this check is there in the PreFlight stage.
+	// That stage was created to perform common checks for all followup stages.
+	services, _ := ctx.Services()
+
+	for _, service := range services {
+		res := service.Resource()
+		if res.GetKind() == "Secret" && res.GetAPIVersion() == "v1" && res.GroupVersionKind().Group == "" {
+			annotations := res.GetAnnotations()
+			for k := range annotations {
+				if strings.HasPrefix(k, "service.binding") {
```
pedjak, line 145: please `AnnotationPrefix` constant: https://github.com/redhat-developer/service-binding-operator/blob/d680987ea0f2869557ada3f3dcc621981cb1d427/pkg/binding/annotationmapper.go#L34

pkg/reconcile/pipeline/handler/collect/impl_test.go
```diff
@@ -745,6 +746,105 @@ var _ = Describe("Collect From Provisioned Service", func() {
...
+			content.SetName(secretName1)
+
+			ctx.EXPECT().ReadSecret(ns1, secretName1).Return(nil, err)
+
+			ctx.EXPECT().RetryProcessing(err)
+			ctx.EXPECT().SetCondition(v1alpha1.Conditions().NotCollectionReady().Reason(collect.ErrorReadingSecret).Msg(err.Error()).Build())
+			collect.DirectSecretReference(ctx)
+		})
+
+		It("continue processing if secret has annotations", func() {
```
pedjak, line 832: stages are arranged in the pipeline, so not sure what do you mean by "continue processing"?  The case is here that a secret will not be considered as direct secret references if it has a service.binding annotation attached to it. Could you please rephrase the test name to reflect that?

baijum: How about something like this?
- `skip processing if the secret has annotations`
- `ensure secret with annotations are ignored`
Please suggest.

pkg/reconcile/pipeline/handler/collect/impl.go
baijum: Done. c7e29669fb69cbc23bbe07b4309aed4e98c11ba4

pkg/reconcile/pipeline/handler/collect/impl_test.go
pedjak: `Ignore secret having a secret binding annotation`

baijum: Done. 387c9cc3e38705018063de2dea18001a762885d7

test/acceptance/features/bindAppToServiceUsingSecret.feature
```diff
@@ -761,3 +761,202 @@ Feature: Bind values from a secret referred in backing service resource
...
+                kind: Secret
+                name: provisioned-secret-1
+              application:
+                name: myaop-provision-srv
+                group: apps
+                version: v1
+                resource: deployments
+          """
+        Then Service Binding "bind-provisioned-service-1" is ready
+        And jq ".status.secret" of Service Binding "bind-provisioned-service-1" should be changed to "provisioned-secret-1"
```
baijum, line 796: This needs to verified.

pedjak (approved):
please squash, otherwise LGTM

PR #985, opened 2021-07-02
add support for Spec API

### LINKED ISSUE #587: Support the spec  ServiceBinding CRD

pedjak:
Reconcile spec ServiceBinding resources as defined by the spec

* the added controller supports the core spec (minus `ClusterApplicationResourceMapping`) + annotation extension
as defined in:

https://github.com/k8s-service-bindings/spec/tree/04b99a3b9215af59e060359848de9622e909ff9e

* the project layout changed to `multi-group` to enable adding controllers for different API groups

* added Spec API based Context implementation and corresponding unit tests
* common API parts extracted into `/apis/common.go` and `/apis/conditions.go`
* common controller logic extracted into `/controllers/common.go`
* added Pipeline that contains only stages required by the spec
* `Pipeline.Process()` accept now any type to accommodate multiple service binding APIs
* `project.PreFlightCheck` can take a number of binding names to check if available and fails otherwise
* `project.BindingAsEnv` can now inject individual bindings into the application resource

Testing:

* added acceptances tests (tagged with `@spec`) demonstrating the usage of the spec
* created dedicated `bindAppToProvisionedService.feature` and moved relevant tests there
* python ServiceBinding class refactored to accommodate the spec and a number of service binding steps moved to `service_binding.py`
* added `jsonpath "{json_path}" on "{res_name}" should return no value` step
* `delete_service_binding()` moved from `openshift.py` to `service_binding.py`

Fixes #587


baijum:
Can we add a test case to demonstrate this mandate about `type` entry in the [Application Projection](https://github.com/k8s-service-bindings/spec#application-projection)?
> The projected binding **MUST** contain a `type` entry with a value that identifies the abstract classification of the binding.

If the Provisioned Service Secret resource doesn't have the `type` entry and there is no value for `.spec.type` field, the reconcile should set the `Ready` status with `False`.

pedjak:
> Can we add a test case to demonstrate this mandate about type entry in the Application Projection?

It is already a part of the PR: https://github.com/redhat-developer/service-binding-operator/pull/985/files#diff-51469be270f5f57f0ddd2ae8883984851ad2e04dd6493d56a79954ae7097a0d4R801

apis/spec/v1alpha2/servicebinding_types.go
```diff
@@ -0,0 +1,162 @@
...
+type ServiceBindingStatus struct {
+	// ObservedGeneration is the 'Generation' of the ServiceBinding that
+	// was last processed by the controller.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// Conditions are the conditions of this ServiceBinding
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// Binding exposes the projected secret for this ServiceBinding
+	Binding *ServiceBindingSecretReference `json:"binding,omitempty"`
```
baijum, line 96: I think we can directly use `*corev1.LocalObjectReference` here.

test/acceptance/features/injectBindingsAsFiles.feature
```diff
@@ -358,3 +358,52 @@ Feature: Bindings get injected as files in application
...
+                  apiVersion: stable.example.com/v1
+                  kind: Backend
+                  name: backend-demo-02
+
+                application:
+                    name: spec-generic-app-a-d-u-2
+                    apiVersion: apps/v1
+                    kind: Deployment
+            """
+        Then Service Binding "spec-binding-backend-vm-02" is ready
```
baijum, line 396: `.status.binding` attribute is mandatory for a Provisioned Service. See the first sentence here:
https://github.com/k8s-service-bindings/spec#provisioned-service

So, this testing behavior is not correct as per the spec.

baijum: Never mind, ServinceBinding allows referring other kinds of resources. It need not be a Provisioned Service. We are good here.

apis/spec/v1alpha2/servicebinding_types.go
pedjak: I literately copied these types from https://github.com/k8s-service-bindings/spec/blob/master/internal/service.binding/v1alpha2/service_binding.go and it would be good if consume the structures in the form as the spec publishes them.

baijum (approved):
LGTM

PR #988, opened 2021-07-15
Drop obsolete 'local' mode of running acceptance tests

pmacik:
### Motivation

Currently there are 2 ways (a.k.a. modes) to execute acceptance tests - `local` (default) and `remote`.
`local` mode means that the operator's binary is executed locally instead of being installed as operator pod inside a Kubernetes cluster. That is a default way of executing the acceptance tests via `make test-acceptance`.

On the other hand, `remote` mode assumes the operator to be installed in the cluster beforehand and just execute the aceptance tests.

There are changes to the operator (such as use of admission webhooks, etc.) comming soon that require the operator to run instide the cluster and the `local` mode would not work and so would be supported.

### Changes

This PR:
* Removes the `local` mode of running acceptance tests
* Makes the `remote` mode the only (and default) way of executing the acceptance tests.
* Updates the acceptance testing docs accordingly

All the PR checks and tests are already running in the `remote` mode.

### Testing

`make test-acceptance`


pedjak:
/hold

local mode has its advantages, and before dropping it, we should  assure how to debug the operator. We can make remote mode as default, but the local mode could be kept if for example webhooks are disabled. For the most of cases, webhooks are not in focus, and could be safely disabled while working on some bug/feature.

Let\s get on the same page before.

pmacik:
Ah, good point @pedjak, thanks. I didn’t know that the webhooks could be disabled. I was under the impression that those would be part of the core function and that the operator would not work without them.

pmacik:
This story assumed that SBO could not work without the webhooks, at all - which would make local mode obsolete. That assumption has shown as wrong... so the local mode is still valid for some debugging situations not affected by the webhooks (those could get disabled, actually).

So, closing this PR as obsolete.

PR #993, opened 2021-07-21
service-binding-operator-967

isutton:
### Motivation

When using the `sliceOfMaps` elementType in a service binding annotation, the operator panics when the specified sourceValue does not exist in the underlying maps, although the expected behaviour would be to inform the user the binding has an error through conditions.

### Changes

This PR changes the current behavior by verifying whether the value extracted is present, and set a condition in the service binding in the case the sourceValue key isn't present.

### Testing

<< How to test those changes, reference implementation of unit/acceptance tests. >>

### DIFF
--- a/pkg/reconcile/pipeline/handler/collect/impl.go
+++ b/pkg/reconcile/pipeline/handler/collect/impl.go
@@ -223,7 +223,15 @@ func collectItems(prefix string, ctx pipeline.Context, service pipeline.Service,
 			p = ""
 		}
 		for _, n := range v.MapKeys() {
-			collectItems(p, ctx, service, n, v.MapIndex(n).Interface())
+			if mapVal := v.MapIndex(n).Interface(); mapVal != nil {
+				collectItems(p, ctx, service, n, mapVal)
+			} else {
+				condition := v1alpha1.Conditions().NotCollectionReady().
+					Msg(fmt.Sprintf("Value for key %v_%v not found", prefix+k.String(), n.String())).
+					Reason("ValueNotFound").Build()
+				ctx.SetCondition(condition)
+				return
+			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/reconcile/pipeline/handler/collect/impl.go
```diff
@@ -223,7 +223,15 @@ func collectItems(prefix string, ctx pipeline.Context, service pipeline.Service,
...
 		}
 		for _, n := range v.MapKeys() {
-			collectItems(p, ctx, service, n, v.MapIndex(n).Interface())
+			if mapVal := v.MapIndex(n).Interface(); mapVal != nil {
+				collectItems(p, ctx, service, n, mapVal)
+			} else {
+				condition := v1alpha1.Conditions().NotCollectionReady().
+					Msg(fmt.Sprintf("Value for key %v_%v not found", prefix+k.String(), n.String())).
+					Reason("ValueNotFound").Build()
+				ctx.SetCondition(condition)
```
pedjak, line 232: apart from setting the condition, you need to do these as well:

```golang
ctx.Error(err)
ctx.StopProcessing()
```

Also, please extract `ValueNotFound` into a constant and put around other constant in the same file.

pkg/reconcile/pipeline/handler/collect/impl_test.go
```diff
@@ -928,6 +928,74 @@ var _ = Describe("Integration Collect definitions + items", func() {
 
 })
 
+// TODO: spec title should be rephrased: should we have one spec for regression tests?
```
pedjak, line 931: I would just add these tests cases under others for the existing specs.

pedjak (changes requested):

PR #997, opened 2021-07-26
Allow metadata updates on Ready bindings

### LINKED ISSUE #996: unable to modify metadata fields
0.9.0

  1. Create service binding resource
  2. try to add ownerReferece, label or annotations to service binding resource

It should be possible to modify metadata fields like ownerReferece, label or annotations even SB that has all conditions `True` 

```
error: servicebindings.binding.operators.coreos.com "busybox-postgresql" could not be patched: admission webhook "vservicebinding.kb.io" denied the request: cannot update Service Binding if 'Ready' condition is True. If you want to rebind to another service/application, remove this binding and create a new one.
```

pedjak:
Webhooks modifed to prevent updates only if `.spec` part of the binding is updated

Fixes #996

### DIFF
--- a/apis/binding/v1alpha1/servicebinding_types.go
+++ b/apis/binding/v1alpha1/servicebinding_types.go
@@ -223,6 +223,10 @@ func (sb *ServiceBinding) HasDeletionTimestamp() bool {
 	return !sb.DeletionTimestamp.IsZero()
 }
 
+func (sb *ServiceBinding) GetSpec() interface{} {
+	return &sb.Spec
+}
+
 // Returns GVR of reference if available, otherwise error
 func (ref *Ref) GroupVersionResource() (*schema.GroupVersionResource, error) {
 	if ref.Resource == "" {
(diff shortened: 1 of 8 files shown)


config/manager/kustomization.yaml
```diff
@@ -11,6 +11,6 @@ configMapGenerator:
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
-- digest: latest
-  name: controller
+- name: controller
   newName: quay.io/redhat-developer/servicebinding-operator
+  newTag: bc81d779
```
pmacik, line 16: @pedjak This file should probably updated with the PR

kadel:
I've just verified this using odo integration tests. 
This fixes the issue that odo is  facing with SBO v0.9.0 where it can't add `ownerReference` on already created ServiceBinding resource. 

Thank you @pedjak for quick fix.

PR #1020, opened 2021-09-13
Apply CSV best practices

pedjak:
* added annotation that indicates that the operator can be deployed in disconnected environments
* specified minimal supported kube version
* added containerImage annotation
* maturity set to 'candidate'
* explicit service account dropped from bundle
* common label renamed to service-binding-controller-manager


config/manifests/bases/service-binding-operator.clusterserviceversion.yaml
```diff
@@ -3,12 +3,14 @@ kind: ClusterServiceVersion
 metadata:
   annotations:
     alm-examples: '[]'
-    capabilities: Basic Install
+    capabilities: Seamless Upgrades
```
pmacik, line 6: `Seamless Upgrades` is something that we need to actually test before claiming it. Let's wait with that until GA.

PR #1023, opened 2021-09-16
Treat invalid "elementType" and "objectType" annotations as errors, not panics

sadlerap:
### Motivation

If a service has an invalid service binding annotation set, we shouldn't attempt to bind the service.  However, current behavior catches most cases (such as an invalid path), but errors currently are ignored, and when either `elementType` or `objectType` isn't valid, we panic.  This could cause issues for users.

### Changes

Rather than panicking, we should fail the binding request and report any errors that are thrown.

### Testing

Attempting to bind against a service with an garbage annotation should now fail and report errors in the binding status.  For example, ```service.binding/credentials: path={.status.dbCredentials},elementType=asdf``` should fail.  This currently implements unit and acceptance tests for both `elementType` and `objectType` errors.  It also adds an acceptance test for invalid paths for the sake of regression testing.

### DIFF
--- a/pkg/binding/annotationmapper.go
+++ b/pkg/binding/annotationmapper.go
@@ -2,8 +2,9 @@ package binding
 
 import (
 	"fmt"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	"github.com/pkg/errors"
 )
@@ -26,12 +27,13 @@ var _ DefinitionBuilder = (*annotationBackedDefinitionBuilder)(nil)
 type modelKey string
 
 const (
-	pathModelKey        modelKey = "path"
-	objectTypeModelKey  modelKey = "objectType"
-	sourceKeyModelKey   modelKey = "sourceKey"
-	sourceValueModelKey modelKey = "sourceValue"
-	elementTypeModelKey modelKey = "elementType"
-	AnnotationPrefix             = "service.binding"
+	pathModelKey                    modelKey = "path"
+	objectTypeModelKey              modelKey = "objectType"
+	sourceKeyModelKey               modelKey = "sourceKey"
+	sourceValueModelKey             modelKey = "sourceValue"
+	elementTypeModelKey             modelKey = "elementType"
+	AnnotationPrefix                         = "service.binding"
+	ProvisionedServiceAnnotationKey          = "servicebinding.io/provisioned-service"
 )
 
 func NewDefinitionBuilder(annotationName string, annotationValue string, configMapReader UnstructuredResourceReader, secretReader UnstructuredResourceReader) *annotationBackedDefinitionBuilder {
@@ -45,25 +47,43 @@ func NewDefinitionBuilder(annotationName string, annotationValue string, configM
 	}
 }
 
-func (m *annotationBackedDefinitionBuilder) outputName() (string, error) {
-	// bail out in the case the annotation name doesn't start with "service.binding"
-	if m.name != AnnotationPrefix && !strings.HasPrefix(m.name, AnnotationPrefix+"/") {
-		return "", fmt.Errorf("can't process annotation with name %q", m.name)
+func (m *annotationBackedDefinitionBuilder) isServiceBindingAnnotation() (bool, error) {
+	if m.name == ProvisionedServiceAnnotationKey {
+		return false, nil
 	}
 
-	if p := strings.SplitN(m.name, "/", 2); len(p) > 1 && len(p[1]) > 0 {
-		return p[1], nil
+	if m.name == AnnotationPrefix {
+		return true, nil
+	} else if strings.HasPrefix(m.name, AnnotationPrefix) {
+		if strings.HasPrefix(m.name, AnnotationPrefix+"/") {
+			return true, nil
+		}
+
+		// it starts with AnnotationPrefix, but has extra text at the end not
+		// separated by a /, so treat it as an error
+		return false, fmt.Errorf("can't process annotation with name %q", m.name)
 	}
 
-	return "", nil
+	// bail out when the annotation name doesn't start with "service.binding"
+	return false, nil
+}
+
+func (m *annotationBackedDefinitionBuilder) outputName() string {
+
+	if p := strings.SplitN(m.name, "/", 2); len(p) == 2 && len(p[1]) > 0 {
+		return p[1]
+	}
+
+	return ""
 }
 
 func (m *annotationBackedDefinitionBuilder) Build() (Definition, error) {
 
-	outputName, err := m.outputName()
-	if err != nil {
+	if valid, err := m.isServiceBindingAnnotation(); !valid || err != nil {
 		return nil, err
 	}
+
+	outputName := m.outputName()
 
 	mod, err := newModel(m.value)
 	if err != nil {
@@ -128,6 +148,5 @@ func (m *annotationBackedDefinitionBuilder) Build() (Definition, error) {
 			sourceValue: mod.sourceValue,
 		}, nil
 	}
-
-	panic(fmt.Sprintf("Annotation %s=%s not implemented!", m.name, m.value))
+	return nil, fmt.Errorf("Annotation %s: %s not implemented!", m.name, m.value)
 }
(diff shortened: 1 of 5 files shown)


pkg/reconcile/pipeline/handler/collect/impl.go
```diff
@@ -63,11 +64,22 @@ func BindingDefinitions(ctx pipeline.Context) {
...
 
 		for k, v := range anns {
+			if !strings.HasPrefix(k, binding.AnnotationPrefix) || k == ProvisionedServiceAnnotationKey {
+				continue
+			}
+
 			definition, err := makeBindingDefinition(k, v, ctx)
 			if err != nil {
-				continue
+				condition := apis.Conditions().NotCollectionReady().
```
pedjak, line 73: you can use here `notCollectionReadyCond` function available within the same package, for improved reading.

test/acceptance/features/bindAppToServiceAnnotations.feature
```diff
@@ -584,3 +584,125 @@ Feature: Bind an application to a service using annotations
...
+                    name: $scenario_id
+                application:
+                    name: $scenario_id
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        Then Service Binding CollectionReady.status is "False"
+        And Service Binding CollectionReady.reason is "InvalidAnnotation"
+        And Service Binding CollectionReady.message is "Failed to create binding definition"
```
pedjak, line 708: we should check here `And Service Binding Ready.message` right?

pedjak (commented):
overall looks good, a few things to improve only.

pkg/reconcile/pipeline/handler/collect/impl.go
```diff
@@ -63,11 +64,22 @@ func BindingDefinitions(ctx pipeline.Context) {
...
+					Msg("Failed to create binding definition").
```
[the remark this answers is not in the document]
sadlerap: I had that behavior in a previous version, but I felt that repeating the error message in multiple locations wasn't necessarily the best way to do things, so I took it out.  I can reimplement it.

test/acceptance/features/bindAppToServiceAnnotations.feature
pedjak: `""Failed to create binding definition"` cannot happen now, given the production code above?

test/acceptance/features/bindAppToServiceAnnotations.feature
```diff
@@ -584,3 +584,127 @@ Feature: Bind an application to a service using annotations
...
+                application:
+                    name: $scenario_id
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        Then Service Binding CollectionReady.status is "False"
+        And Service Binding CollectionReady.reason is "InvalidAnnotation"
+        And Service Binding CollectionReady.message is "Failed to create binding definition from "service.binding/credentials: path=asdf": could not create binding model for annotation key service.binding/credentials and value path=asdf: path has invalid syntax: "asdf""
+	And Service Binding Ready.message is "could not create binding model for annotation key service.binding/credentials and value path=asdf: path has invalid syntax: "asdf""
```
pedjak, line 710: please fix the indentation.

pkg/reconcile/pipeline/handler/collect/impl.go
sadlerap: For some reason that I really don't understand, using `notCollectionReadyCond` seems to cause this to break.  For reference, see the CI logs for commits 678672f and 44c2abf.

pkg/binding/annotationmapper.go
```diff
@@ -45,23 +46,32 @@ func NewDefinitionBuilder(annotationName string, annotationValue string, configM
 	}
 }
 
+var skip = errors.New("skip")
+
 func (m *annotationBackedDefinitionBuilder) outputName() (string, error) {
-	// bail out in the case the annotation name doesn't start with "service.binding"
-	if m.name != AnnotationPrefix && !strings.HasPrefix(m.name, AnnotationPrefix+"/") {
```
pedjak, line 50: we were accepting following annotations so far:

* `service.binding: value`
* `service.binding/xx: value`

Now, this is not the case anymore. We should revert back to that logic. We should only ignore `service.binding/provisioned-service` annotation

sadlerap: We also need to explicitly ignore non-service binding annotations here (such as `foo: bar`).  Previously, we were throwing an error when we found these annotations, but since we were also ignoring the errors that `annotationBackedDefinitionBuilder.Build` would throw, the result was effectively the same.

Since we're no longer ignoring these errors, we need to make these annotations not return errors (since they're not errors that we should stop binding for).  I think we should only return errors in the case that we need to either retry or stop processing the binding request, and an annotation not under the service binding namespace isn't such an error.

pkg/binding/annotationmapper.go
```diff
@@ -45,23 +46,32 @@ func NewDefinitionBuilder(annotationName string, annotationValue string, configM
...
+			// it starts with service.binding but doesn't have a trailing /, so it's not valid
+			return "", fmt.Errorf("can't process annotation with name %q", m.name)
+		} else if p := strings.SplitN(m.name, "/", 2); len(p) == 2 && len(p[1]) > 0 {
+			return p[1], nil
+		}
 	}
 
-	return "", nil
+	// bail out when the annotation name doesn't start with "service.binding"
+	return "", skip
```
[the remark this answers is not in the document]
sadlerap: fixed

pkg/binding/annotationmapper.go
```diff
@@ -45,26 +47,43 @@ func NewDefinitionBuilder(annotationName string, annotationValue string, configM
...
+
+	if m.name == AnnotationPrefix {
+		return true, nil
+	} else if strings.HasPrefix(m.name, AnnotationPrefix) {
+		if strings.HasPrefix(m.name, AnnotationPrefix+"/") {
+			return true, nil
+		}
+
+		// it starts with service.binding but doesn't have a trailing /, so it's not valid
+		return false, fmt.Errorf("can't process annotation with name %q", m.name)
```
pedjak, line 63: this was valid annotation until now, see https://github.com/redhat-developer/service-binding-operator/blob/3017eb443427396252959298b1b8e1ed923f29bd/test/acceptance/features/bindAppToService.feature#L31

sadlerap: It's just the comment that's wrong, the behavior of the code is correct.  I'll update the comment.

pkg/binding/annotationmapper.go
```diff
@@ -45,26 +47,44 @@ func NewDefinitionBuilder(annotationName string, annotationValue string, configM
...
+	if m.name == AnnotationPrefix {
+		return true, nil
+	} else if strings.HasPrefix(m.name, AnnotationPrefix) {
+		if strings.HasPrefix(m.name, AnnotationPrefix+"/") {
+			return true, nil
+		}
+
+		// it starts with AnnotationPrefix, but has extra text at the end not
+		// separated by a /, so treat it as an error
+		return false, fmt.Errorf("can't process annotation with name %q", m.name)
```
pedjak, line 64: I would not be so rigid, here is what should return `true`:
* `service.binding`
* `service.binding/xxx`
everything else should return `false`, and I do not think that we case when we should return error.

sadlerap: `service.binding` is handled explicitly in [line 55](https://github.com/redhat-developer/service-binding-operator/pull/1023/commits/a27aa032c20919ed62363578c9703c56bca3771d#diff-73e8f8345b43001c4d65bc139e25b3ed9b17a921772b50a3f0a0a125f0758a60R55), and `service.binding/xxx` is handled by [line 58](https://github.com/redhat-developer/service-binding-operator/pull/1023/commits/a27aa032c20919ed62363578c9703c56bca3771d#diff-73e8f8345b43001c4d65bc139e25b3ed9b17a921772b50a3f0a0a125f0758a60R58).  If we've reached here, then we've hit an annotation that looks like `service.bindingxxx`.  Previous behavior here was to throw an error, and I'm preserving that behavior.

pedjak: thanks for the clarification, I missed that.

Issue #983, opened 2021-06-22
SERVICE_BINDING_ROOT env variable is defined twice in the deployment


feloy:
### What is the environment (Minikube, Openshift)?

Kubernetes 1.20.6

### What is the SBO version used?

0.8.0

### What are the steps to reproduce this issue?

```
kubectl apply -f - <<EOF
apiVersion: v1
kind: Service
metadata:
  name: otherapp
spec:
  selector:
    name: otherapp
  ports:
  - port: 8080
    targetPort: 8080

---

apiVersion: apps/v1
kind: Deployment
metadata:
  name: myapp
spec:
  selector:
    matchLabels:
      app: myapp
  template:
    metadata:
      labels:
        app: myapp
    spec:
      containers:
      - name: myapp
        image: postgres
        resources:
          limits:
            memory: "128Mi"
            cpu: "500m"
        ports:
        - containerPort: 8080
        env:
        - name: POSTGRES_PASSWORD
          value: foobar
        volumeMounts:
          - mountPath: /var/lib/postgresql
            name: data
          - mountPath: /var/run/postgresql
            name: run
      volumes:
      - name: data
        emptyDir: {}
      - name: run
        emptyDir: {}
---

apiVersion: binding.operators.coreos.com/v1alpha1
kind: ServiceBinding
metadata:
  name: asbr
spec:
  application:
    group: apps
    version: v1
    resource: deployments
    name: myapp
  bindAsFiles: true
  detectBindingResources: true
  services:
  - group: ""
    version: v1
    kind: Service
    name: otherapp
    namespace: bug1
EOF
```

### What is the actual behaviour?

```
$ kubectl get deployment myapp -o yaml 
[...]
    spec:
      containers:
      - env:
        - name: SERVICE_BINDING_ROOT
          value: /bindings
        - name: SERVICE_BINDING_ROOT
          value: /bindings
```

### Additional Information (Screenshots, etc)
This can cause problem especially in `odo` when trying to update the deployment:

```
Failed to create manager for existing fields: failed to convert new object (apps/v1, Kind=Deployment) to smd typed: .spec.template.spec.containers[name="runtime"].env: duplicate entries for key [name="SERVICE_BINDING_ROOT"]
```

PR #1029, opened 2021-09-22
Remove duplicate setting of SERVICE_BINDING_ROOT env while projecting files

pmacik:
### Motivation

Issue: #983 

### Changes

This PR:
* Removes duplicate setting of `SERVICE_BINDING_ROOT` env variable
* Adds acceptance test related to `SERVICE_BINDING_ROOT` env variable for binding as files

### Testing

Related acceptance tests in PR checks.

### DIFF
--- a/test/acceptance/features/injectBindingsAsFiles.feature
+++ b/test/acceptance/features/injectBindingsAsFiles.feature
@@ -115,6 +115,7 @@ Feature: Bindings get injected as files in application
             """
             8080
             """
+        And The container declared in application resource contains env "SERVICE_BINDING_ROOT" set only once
 
     Scenario: Binding is injected as file into application at the location specified through mountPath
         Given Generic test application "generic-app-a-d-u-3" is running
@@ -363,3 +364,64 @@ Feature: Bindings get injected as files in application
             """
             mysql
             """
+
+    Scenario: SERVICE_BIDNING_ROOT is not defined twice in the deployment after binding two services
+        Given Generic test application is running
+        * The env var "SERVICE_BINDING_ROOT" is not available to the application
+        * The Service is present
+            """
+            apiVersion: v1
+            kind: Service
+            metadata:
+                name: $scenario_id-2
+                annotations:
+                    service.binding/service2: path={.metadata.name}
+            spec:
+                selector:
+                    name: $scenario_id
+                ports:
+                  - port: 80
+                    targetPort: 8080
+            """
+        * The Service is present
+            """
+            apiVersion: v1
+            kind: Service
+            metadata:
+                name: $scenario_id-3
+                annotations:
+                    service.binding/service3: path={.metadata.name}
+            spec:
+                selector:
+                    name: $scenario_id
+                ports:
+                  - port: 80
+                    targetPort: 8080
+            """
+        When Service Binding is applied
+            """
+            apiVersion: binding.operators.coreos.com/v1alpha1
+            kind: ServiceBinding
+            metadata:
+                name: $scenario_id
+            spec:
+                application:
+                    group: apps
+                    version: v1
+                    resource: deployments
+                    name: $scenario_id
+                bindAsFiles: true
+                detectBindingResources: true
+                services:
+                  - group: ""
+                    version: v1
+                    kind: Service
+                    name: $scenario_id-2
+                  - group: ""
+                    version: v1
+                    kind: Service
+                    name: $scenario_id-3
+            """
+        Then Service Binding is ready
+        * Service Binding has the binding secret name set in the status
+        * The container declared in application resource contains env "SERVICE_BINDING_ROOT" set only once
--- a/test/acceptance/features/steps/generic_testapp.py
+++ b/test/acceptance/features/steps/generic_testapp.py
@@ -3,6 +3,7 @@ import requests
 import json
 import polling2
 from behave import step
+from openshift import Openshift
 from util import scenario_id
 from string import Template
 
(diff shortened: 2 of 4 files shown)


test/acceptance/features/injectBindingsAsFiles.feature
```diff
@@ -323,3 +323,181 @@ Feature: Bindings get injected as files in application
...
+                  - group: ""
+                    version: v1
+                    kind: Service
+                    name: $scenario_id-3
+            """
+        Then Service Binding is ready
+        * Service Binding has the binding secret name set in the status
+        * The application "$scenario_id" container has only one env "SERVICE_BINDING_ROOT" set
+
+    Scenario: SERVICE_BIDNING_ROOT is not defined twice in the deployment after binding to service without bindable data
```
pedjak, line 430: as talked about this, no binding data should fail binding process, and that should be solved separatelly

test/acceptance/features/injectBindingsAsFiles.feature
```diff
@@ -323,3 +323,181 @@ Feature: Bindings get injected as files in application
             """
             mysql
             """
+
+    Scenario: SERVICE_BIDNING_ROOT is not defined twice in the deployment after binding
```
pedjak, line 327: we need also scenario when `SERVICE_BINDING_ROOT` is already defined on the app, but then no additional env var is added.

pedjak (commented):
I think we do not need to add new tests, we can add additional assertions for those in `test/acceptance/features/injectBindingsAsFiles.feature`

test/acceptance/features/injectBindingsAsFiles.feature
pmacik: +1 https://github.com/redhat-developer/service-binding-operator/pull/1031

test/acceptance/features/injectBindingsAsFiles.feature
pedjak: IMHO, we do not need a new test - we can add additional assertion to https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/injectBindingsAsFiles.feature#L71 because checking for `SERVICE_BINDING_ROOT` duplicate is one aspect of projecting binding as files.

test/acceptance/features/steps/generic_testapp.py
```diff
@@ -113,3 +115,11 @@ def set_common_label(context, label):
 def check_env_var_value_in_both_apps(context, name, value):
     polling2.poll(lambda: context.application1.get_env_var_value(name) == value, step=5, timeout=400)
     polling2.poll(lambda: context.application2.get_env_var_value(name) == value, step=5, timeout=400)
+
+
+@step(u'The application "{app_name}" container has only one env "{envVar}" set')
```
pedjak, line 120: The step is a bit unclear, because container at runtime has or does not have an env var set. Here we speak about env vars declared in application resource. While here, would you also like to make the step where we do not need to specify application name, like:

```
The container declared in application resource contains env "{envVar}" set only once'
```

test/acceptance/features/steps/service_binding.py
```diff
@@ -86,6 +86,8 @@ def invalid_sbr_is_applied(context):
 def sbo_is_ready(context, sbr_name=None):
     if sbr_name is None:
         sbr_name = list(context.bindings.values())[0].name
+    else:
+        sbr_name = Template(sbr_name).substitute(scenario_id=scenario_id(context))
```
pedjak, line 90: not seeing its usage somewhere in this PR?

test/acceptance/features/unbindAppToService.feature
```diff
@@ -247,3 +247,129 @@ Feature: Unbind an application from a service
         * File "/bindings/spec-remove-bindings-as-files-app-sb/port" is unavailable in application pod
         * File "/bindings/spec-remove-bindings-as-files-app-sb/type" is unavailable in application pod
         * Service Binding secret is not present
+
+    Scenario: Bind without SERVICE_BINDING_ROOT set, unbind and re-bind
```
pedjak, line 251: what is the purpose of this tests, i.e. expectations? How it is related to the submitted fix?

test/acceptance/features/unbindAppToService.feature
```diff
@@ -247,3 +247,129 @@ Feature: Unbind an application from a service
...
+        Then Service Binding is ready
+        * The env var "host" is not available to the application
+        * The application env var "SERVICE_BINDING_ROOT" has value "/bindings"
+        * Content of file "/bindings/$scenario_id/host" in application pod is
+            """
+            example.common
+            """
+        * The application "$scenario_id" container has only one env "SERVICE_BINDING_ROOT" set
+
+    Scenario: Bind with SERVICE_BIDNING_ROOT already set, unbind and re-bind
```
pedjak, line 314: same question as above?

pedjak (changes requested):

test/acceptance/features/steps/generic_testapp.py
pmacik: +1

test/acceptance/features/unbindAppToService.feature
pmacik: After deeper consideration, this is redundant test, will remove it, thx

test/acceptance/features/injectBindingsAsFiles.feature
pmacik: +1

test/acceptance/features/unbindAppToService.feature
pmacik: Those tests should verify various situations where SERVICE_BINDING_ROOT is or not set on the application. But after deeper consideration, I can see that those cases are covered by the other tests after adding the additional step for checking number of env occurencies... so removing this from PR

Thanks

test/acceptance/features/steps/service_binding.py
pmacik: not used, now - removing as not related to the PR

pedjak (approved):
/lgtm please squash

pmacik:
Rebased and squashed

Issue #1030, opened 2021-09-23
Invalid VolumeMount or secretRef to non-existing intermediate secret is attached to application when there is nothing to bind.


pmacik:
### What is the environment (Minikube, Openshift)?
Any

### What is the SBO version used?
0.10.0

### What are the steps to reproduce this issue?
  1. Create a service with no data exposed and no owned resources
  2. Create an application to bind with
  3. Create a service binding without any mapping to bind the service and application
 
### What is the expected behaviour?
Binding should fail at Injection phase complaining that there is nothing to bind

### What is the actual behaviour?
Service binding conditions says that binding is ready.
VolumeMount (bind as files) or `envFrom` secret referrence (bind as env) is created for the binding's intermediate secret that is not created because there is nothing to bind.

PR #1031, opened 2021-09-23
Do not bind if there is no binding item collected from the service

pmacik:
### Motivation

Issue #1030

### Changes

Do not bind as files if there is no binding data is collected from the service

### DIFF
--- a/pkg/reconcile/pipeline/handler/project/impl.go
+++ b/pkg/reconcile/pipeline/handler/project/impl.go
@@ -28,8 +28,14 @@ func PreFlightCheck(mandatoryBindingKeys ...string) func(pipeline.Context) {
 			ctx.StopProcessing()
 			return
 		}
+		items := ctx.BindingItems()
+		if len(items) == 0 {
+			err := errors.New("no binding data to project")
+			ctx.RetryProcessing(err)
+			ctx.SetCondition(apis.Conditions().NotInjectionReady().Reason(apis.NoBindingDataReason).Msg(err.Error()).Build())
+			return
+		}
 		if len(mandatoryBindingKeys) > 0 {
-			items := ctx.BindingItems()
 			itemMap := items.AsMap()
 			for _, bk := range mandatoryBindingKeys {
 				if _, found := itemMap[bk]; !found {
--- a/test/acceptance/features/bindAppToService.feature
+++ b/test/acceptance/features/bindAppToService.feature
@@ -570,6 +570,8 @@ Feature: Bind an application to a service
             kind: Backend
             metadata:
                 name: demo-backserv-cr-3
+                annotations:
+                    service.binding/name: path={.metadata.name}
             """
         * Generic test application "gen-app-a-s-e" is running
         * Service Binding is applied
@@ -646,6 +648,8 @@ Feature: Bind an application to a service
             kind: Backend
             metadata:
                 name: demo-backserv-cr-5
+                annotations:
+                    service.binding/name: path={.metadata.name}
             """
         * Generic test application "gen-app-a-s-g" is running
         * Service Binding is applied
@@ -696,6 +700,8 @@ Feature: Bind an application to a service
             kind: Backend
             metadata:
                 name: demo-backserv-cr-4
+                annotations:
+                    service.binding/name: path={.metadata.name}
             """
         * Generic test application "gen-app-a-s-h" is running
         * Service Binding is applied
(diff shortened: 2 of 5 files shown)


pkg/reconcile/pipeline/handler/project/impl.go
```diff
@@ -70,6 +70,11 @@ func BindingsAsEnv(ctx pipeline.Context) {
 		return
 	}
 
+	if len(ctx.BindingItems()) == 0 {
```
pedjak, line 73: we should put this check in `PreFlightCheck` function.

pkg/reconcile/pipeline/handler/project/impl.go
```diff
@@ -70,6 +70,11 @@ func BindingsAsEnv(ctx pipeline.Context) {
...
+		stop(ctx, errors.New("no binding item to project"))
```
pedjak, line 74: `stop` means that we are not going to retry, but we should actually.

pkg/reconcile/pipeline/handler/project/impl.go
```diff
@@ -160,6 +165,10 @@ func BindingsAsFiles(ctx pipeline.Context) {
 	if !ctx.BindAsFiles() {
 		return
 	}
+	if len(ctx.BindingItems()) == 0 {
+		stop(ctx, errors.New("no binding item to project"))
+		return
+	}
```
pedjak, line 171: see above comment on `PreFlightCheck` - once there, this check can be removed.

test/acceptance/features/bindAppToService.feature
```diff
@@ -842,3 +842,30 @@ Feature: Bind an application to a service
...
+            metadata:
+                name: $scenario_id
+            spec:
+                bindAsFiles: false
+                detectBindingResources: true
+                services:
+                  - group: ""
+                    version: v1
+                    kind: Service
+                    name: $scenario_id
```
pedjak, line 862: we did not create service earlier to bind to?

test/acceptance/features/injectBindingsAsFiles.feature
```diff
@@ -275,6 +275,33 @@ Feature: Bindings get injected as files in application
...
+            metadata:
+                name: $scenario_id
+            spec:
+                bindAsFiles: true
+                detectBindingResources: true
+                services:
+                  - group: ""
+                    version: v1
+                    kind: Service
+                    name: $scenario_id
```
pedjak, line 294: same as above

pedjak (changes requested):

test/acceptance/features/bindAppToService.feature
pmacik: The service is created along with the generic test application, but for clarity I can add a separate service so it is clear in the scenario that the service is present

pkg/reconcile/pipeline/handler/project/impl.go
pmacik: +1

pkg/reconcile/pipeline/handler/project/impl.go
pmacik: +1

test/acceptance/features/bindAppToService.feature
pmacik: Done

test/acceptance/features/injectBindingsAsFiles.feature
pmacik: Done, thx

pkg/reconcile/pipeline/handler/project/impl.go
pmacik: good point, thanks

pkg/reconcile/pipeline/handler/project/impl.go
```diff
@@ -40,6 +40,13 @@ func PreFlightCheck(mandatoryBindingKeys ...string) func(pipeline.Context) {
 					return
 				}
 			}
+		} else {
+			if len(items) == 0 {
+				err := errors.New("no binding item to project")
+				ctx.RetryProcessing(err)
+				ctx.SetCondition(apis.Conditions().NotInjectionReady().Reason(apis.NoBindingItemsReason).Msg(err.Error()).Build())
+				return
```
pedjak, line 48: go conventions try to avoid `else` blocks. do this check right after line 31 in a separate if then block.

test/acceptance/features/bindAppToService.feature
```diff
@@ -570,6 +570,8 @@ Feature: Bind an application to a service
             kind: Backend
             metadata:
                 name: demo-backserv-cr-3
+                annotations:
+                    service.binding/name: path={.metadata.name}
```
pedjak, line 574: how is this change related to PR change?

test/acceptance/features/bindAppToService.feature
```diff
@@ -842,3 +848,43 @@ Feature: Bind an application to a service
...
+                    name: $scenario_id-svc
+                application:
+                    name: $scenario_id
+                    group: apps
+                    version: v1
+                    resource: deployments
+            """
+        Then Service Binding CollectionReady.status is "True"
+        * Service Binding InjectionReady.status is "False"
+        * Service Binding InjectionReady.message is "no binding item to project"
```
pedjak, line 890: I would perhaps change the message to `no binding data to project`

pedjak (changes requested):

test/acceptance/features/bindAppToService.feature
pmacik: The PR makes it necessary to have some binding data in order to finish binding successfully and to make it 'Ready'. Those 3 scenarios did not have anything bindable and so with introducing this PR it failed at first SB being checked for Readiness. So I have to introduce some binding data for those scenarios to work.

PR #1034, opened 2021-09-27
Remove obsolete examples in favour of SBO docs

pmacik:
### Motivation

Currently, this repo contains a set of examples under `examples` directory along with couple of acceptance tests tagged with`@examples`.

Those examples are quite old and obsolete and the acceptance tests meant for verifying it are disabled by default.

### Changes

This PR:
* Removes `examples` directory completely
* Removes all acceptance tests tagged with `@example`


.github/workflows/pr-checks.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
     runs-on: ubuntu-20.04
 
     env:
-      EXTRA_BEHAVE_ARGS: "--tags=~@knative --tags=~@openshift --tags=~@examples"
+      EXTRA_BEHAVE_ARGS: "--tags=~@knative --tags=~@openshift"
```
pedjak, line 71: I would revert this line, so that when we examples, we do not need to touch this line again.

Makefile
```diff
@@ -52,9 +52,9 @@ TEST_ACCEPTANCE_CLI ?= oc
 TEST_ACCEPTANCE_TAGS ?=
 
 ifdef TEST_ACCEPTANCE_TAGS
-TEST_ACCEPTANCE_TAGS_ARG ?= --tags="~@disabled" --tags="~@examples" --tags="$(TEST_ACCEPTANCE_TAGS)"
+TEST_ACCEPTANCE_TAGS_ARG ?= --tags="~@disabled" --tags="$(TEST_ACCEPTANCE_TAGS)"
```
pedjak, line 55: same as above

Makefile
```diff
@@ -52,9 +52,9 @@ TEST_ACCEPTANCE_CLI ?= oc
...
 else
-TEST_ACCEPTANCE_TAGS_ARG ?= --tags="~@disabled" --tags="~@examples"
+TEST_ACCEPTANCE_TAGS_ARG ?= --tags="~@disabled"
```
pedjak, line 57: same as above

.github/workflows/pr-checks.yaml
pmacik: Good point, thx - reverted

Makefile
pmacik: Good point, thx - reverted

Makefile
pmacik: Good point, thx - reverted

pedjak (approved):

PR #1036, opened 2021-09-28
Detect bindable kinds and expose them for client discovery

pedjak:
The operator detects `CustomResourceDefinitions` exposing binding data
and reports bindable resource types (group, version, kind) within
cluster-scoped instance of `BindableKinds`, under `.status` field:

```
kind: BindableKinds
metadata:
  name: bindable-kinds
status:
- group: stable.example.com
  kind: BindableBackend
  version: v1
.
.
.
```

There is only one instance of `BindableKinds`, named `bindable-kinds`


pmacik:
@pedjak The acceptance test `Discover bindable service when defined by v1beta1 CRD API` for 4.9 fail because CRD API `v1beta1` is not supported in 4.9+, only `v1` is.

test/acceptance/features/discoverBindableResources.feature
```diff
@@ -0,0 +1,20 @@
...
+        And Service Binding Operator is running
+    Scenario: Discover the list of bindable kinds by reading status part of cluster-scoped BindableKinds resource
+        Given OLM Operator "provisioned_backend_with_annotations" is running
+        And OLM Operator "backend_with_annotations" is running
+        Then bindablekinds/bindable-kinds is available in the cluster
+#        And User acceptance-tests-dev can read resource bindablekinds/bindable-kinds
+        And Kind ProvisionedBackend with apiVersion stable.example.com/v1 is listed in bindable kinds
+        And Kind BindableBackend with apiVersion stable.example.com/v1 is listed in bindable kinds
+
+    Scenario: Discover bindable service when defined by v1beta1 CRD API
```
pmacik, line 17: This scenario will fail in OpenShift v4.9+ because `v1beta1` CRD API was removed in 4.9 and is not supported.. only `v1` is.

So this scenarios should be disabled for OpenShift 4.9

pmacik (changes requested):

test/acceptance/features/discoverBindableResources.feature
pedjak: > So this scenarios should be disabled for OpenShift 4.9

Yes, submitted https://github.com/openshift/release/pull/22338 please review it.

PR #1040, opened 2021-10-04
Add support for Opstree Redis operator-backed services

pedjak:
Provisioned `redis.redis.opstreelabs.in/v1beta1`/`Redis` instances are bindable out of the box, exposing following binding data:
* type (always set to 'redis')
* host
* password

In order to expose `type`, support for binding constant has been added. Namely, if a service binding annotation value does not specify `path`,
then the annotation value is used as the binding value. With that it is possible to state something like this:

```
service.binding/type: redis
```

Unit and acceptance tests have been added to verify the introduced behavior.

Solves indirectly #843

### DIFF
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -10,6 +10,8 @@ resources:
 - bindablekinds_viewer_rolebinding.yaml
 - servicebinding_controller_role.yaml
 - servicebinding_controller_rolebinding.yaml
+# operators supproted out of the box
+- opstree_redis_clusterrole.yaml
 # Comment the following 4 lines if you want to disable
 # the auth proxy (https://github.com/brancz/kube-rbac-proxy)
 # which protects your /metrics endpoint.
--- a/controllers/crd/crd_controller.go
+++ b/controllers/crd/crd_controller.go
@@ -24,6 +24,7 @@ import (
 	"github.com/redhat-developer/service-binding-operator/pkg/client/kubernetes"
 	"github.com/redhat-developer/service-binding-operator/pkg/converter"
 	"github.com/redhat-developer/service-binding-operator/pkg/reconcile/pipeline/context/service"
+	"github.com/redhat-developer/service-binding-operator/pkg/util"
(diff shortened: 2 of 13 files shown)


config/rbac/kustomization.yaml
```diff
@@ -10,6 +10,9 @@ resources:
 - bindablekinds_viewer_rolebinding.yaml
 - servicebinding_controller_role.yaml
 - servicebinding_controller_rolebinding.yaml
+# operators supproted out of the box
+- opstree_redis_clusterrole.yaml
+#- opstree_redis_viewer_rolebinding.yaml
```
baijum, line 15: This line is commented. If not required, remove this and the corresponding YAML file.

controllers/crd/crd_controller.go
```diff
@@ -96,10 +108,27 @@ func (r *CrdReconciler) Reconcile(req ctrl.Request) (reconcileResult ctrl.Result
...
+					log.Error(err, "Error updating CRD")
+					return ctrl.Result{}, err
+				}
+				log.Info("Annotations applied")
+			}
 		}
 	}
 
+	if !toPersist {
+		log.Info("Done")
```
baijum, line 129: A descriptive log message would be better.

baijum (approved):
/lgtm

pkg/binding/definition.go
```diff
@@ -65,6 +67,13 @@ func (d *stringDefinition) Apply(u *unstructured.Unstructured) (Value, error) {
 	if d.outputName == "" {
 		return nil, fmt.Errorf("cannot use generic service.binding annotation for string elements, need to specify binding key like service.binding/foo")
 	}
+	if d.value != "" {
+		return &value{
+			v: map[string]interface{}{
+				d.outputName: reflect.ValueOf(d.value),
+			},
+		}, nil
+	}
```
sadlerap, line 76: Can we get a test for this?  `pkg/binding/definition_test.go` doesn't cover this, since it doesn't have a test for a set `value` getting set in a `stringDefinition`.

sadlerap (changes requested):
Looks mostly good.

sadlerap (approved):

PR #1043, opened 2021-10-06
Add support for CrunchyData Postgres operator

pedjak:
In order to read binding data from the right secret, the value of `path` key of a service binding annotation
is treated as a template:
* multiple json paths can specified and the results of their evaluation are rendered in the final binding value
* the template can contain arbitrary additional strings that are rendered 'as-is' in the final binding value

Hence, it is possible to generate proper secret reference and read binding data from it. In the case of
CrunchyData Postgres Operator, this is the right annotations:

```
"service.binding": "path={.metadata.name}-pguser-{.metadata.name},objectType=Secret",
```

On top of that, `user` is exposed as `username` as well, and `dbname` and `database`, so that
more applications can find all the binding data they need.

### DIFF
--- a/test/acceptance/features/steps/crunchypostgresoperator.py
+++ b/test/acceptance/features/steps/crunchypostgresoperator.py
@@ -1,0 +1,24 @@
+from olm import Operator
+from environment import ctx
+from behave import given
+
+
+class CrunchyPostgresOperator(Operator):
+
+    def __init__(self, name="pgo"):
+        self.name = name
+        if ctx.cli == "oc":
+            self.operator_catalog_source_name = "community-operators"
+        else:
+            self.operator_catalog_source_name = "operatorhubio-catalog"
+        self.operator_catalog_channel = "v5"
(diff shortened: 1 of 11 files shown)


test/acceptance/features/steps/crunchypostgresoperator.py
```diff
@@ -0,0 +1,24 @@
...
+        self.name = name
+        if ctx.cli == "oc":
+            self.operator_catalog_source_name = "community-operators"
+        else:
+            self.operator_catalog_source_name = "operatorhubio-catalog"
+        self.operator_catalog_channel = "v5"
+        self.package_name = "postgresql"
+
+
+@given(u'Crunchy Data Postgres Redis operator is running')
```
baijum, line 18:
```suggestion
@given(u'Crunchy Data Postgres operator is running')
```

baijum (changes requested):
Few typos. Otherwise looks good to me.

baijum (approved):
/lgtm

Issue #656, opened 2020-08-31
binding:env:attribute being missed/ignored currently while fetching info from x-descriptors


Avni-Sharma:
If we try to export the backing service values via just the x-descriptors(i.e CSV in OLM) then `binding:env:attribute` or `binding:volumemount:attribute` is missed currently here https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/controller/servicebindingrequest/service.go#L82 .  Need to tackle it as https://gist.githubusercontent.com/isutton/61e20b4034207ef065821246df6ab3a4/raw/2764b977396f4ac9148b7dd4ea5b2e234bddbbe4/loadDescriptor.go as suggested by @isutton .

pedjak:
Got obsolete after merging https://github.com/redhat-developer/service-binding-operator/issues/403

PR #1070, opened 2021-11-04
test(acceptance): remove obsolete scenario

pmacik:
### Motivation

Currently, acceptance tests contain disabled scenario for a feature that didn't work (https://github.com/redhat-developer/service-binding-operator/issues/656). However, the related issue was closed as obsolete and the actual feature (exposing a resource field as string via OLM x-descriptors) is actually already covered by a [different scenario](https://github.com/redhat-developer/service-binding-operator/blob/31886b30b2cc101a74d10359d864ea6fdb8357fb/test/acceptance/features/bindAppToService.feature#L528).

While being disabled it always registers as a skipped scenario in the test stats.

### Changes

This PR:
* Removes obsolete scenario `Create binding secret using specDescriptors definitions managed in OLM operator descriptors` which is `@disabled` anyway.

### Testing

PR checks should do.


pedjak (approved):
/lgtm
/approve

Issue #888, opened 2021-03-02
Sample integration tests application should provide more advanced use cases


wtrocki:
Sample application used in tests should add more use cases like:
https://github.com/redhat-developer/service-binding-operator/tree/master/test

Needed:
- Nested arguments
annotations:
    service.binding/bootstrapServer: 'path={.status.bootstrapServer.host}'
- Mapping secret values

pedjak:
@pmacik can you take a look and see if some tests can be added?

pmacik:
The nested arguments scenario as suggested in the description was not covered so I added it in #888.
However, mapping secret values is already covered by:

wtrocki:
Thx

PR #1072, opened 2021-11-05
test(acceptance): add annotations with nested attributes into scenario for CRD annotations

pmacik:
### Motivation

#888 Suggests some scenarios to be covered in acceptance tests. The nested arguments scenario as suggested is not covered by any current acceptance test.

### Changes

This PR:
* Enhances one of the scenarios with binding annotations for nested arguments as suggested in #888

### Testing

PR checks should do

### DIFF
--- a/test/acceptance/features/bindAppToServiceAnnotations.feature
+++ b/test/acceptance/features/bindAppToServiceAnnotations.feature
@@ -20,6 +20,8 @@ Feature: Bind an application to a service using annotations
                 annotations:
                     service.binding/host: path={.spec.host}
                     service.binding/ready: path={.status.ready}
+                    service.binding/environment: path={.spec.userLabels.environment}
+                    service.binding/dataType: path={.status.data.type}
             spec:
                 group: stable.example.com
                 versions:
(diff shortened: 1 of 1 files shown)


test/acceptance/features/bindAppToServiceAnnotations.feature
```diff
@@ -518,8 +518,15 @@ Feature: Bind an application to a service using annotations
                 name: binding-app-via-gvk-service
                 annotations:
                     service.binding/host_internal_db: path={.spec.host_internal_db}
+                    service.binding/environment: path={.spec.userLabels.environment}
+                    service.binding/dbCredentials: path={.status.data.dbCredentials}
             spec:
                 host_internal_db: internal.db.stable.example.com
+                userLabels:
```
pedjak, line 525: this scenario has a focus on something fully different, and hence I would not add broader its focus. Instead, let's add a separate test or find another which focus is on exposing binding data using annotations.

pedjak (changes requested):
please rephrase commit message to state exactly what kind of test was added and please do not refer GitHub issue in it.

test/acceptance/features/bindAppToServiceAnnotations.feature
pmacik: Good point, thx @pedjak 
Updated a different scenario

pedjak (approved):
/lgtm
/approve

PR #1076, opened 2021-11-30
build: add recipe for generating mocks

sadlerap:
### Motivation

When making changes to mocked interfaces, I'd like a way for our build system to automatically regenerate our mocks.

### Changes

`mockgen` is now placed in `bin/`, so installing it globally is no longer required.

I don't have the new `gen-mocks` recipe as a dependency for running unit tests, since you're not likely going to need to run it very often.  Ideally, I'd have it as a dependency and have it run only when mocked interfaces have changed in some way, but I'm not sure how to teach `make` to do that kind of introspection.

### Testing

Run `make gen-mocks`; mocks should be regenerated.

### DIFF
--- a/hack/common.mk
+++ b/hack/common.mk
@@ -79,6 +79,9 @@ endif
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+gen-mocks: mockgen
+	PATH=$(shell pwd)/bin:$(shell printenv PATH) $(GO) generate $(V_FLAG) ./...
+
 # Download controller-gen locally if necessary
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 controller-gen:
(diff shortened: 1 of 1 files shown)


controllers/common.go
```diff
@@ -120,4 +120,5 @@ func (r *BindingReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	}
 	return result, nil
 }
-// end::reconcile-func[]
\ No newline at end of file
+
+// end::reconcile-func[]
```
pedjak, line 124: whitespace change, please revert.

hack/common.mk
```diff
@@ -79,6 +79,9 @@ endif
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+gen-mocks: mockgen
+	PATH=$(PATH):$(shell pwd)/bin $(GO) generate $(V_FLAG) ./...
```
pedjak, line 83: I would put `bin` folder as the first thing in the path, so that we can be sure the tight `mockgen` version is used.

pedjak (changes requested):

controllers/common.go
sadlerap: Reverted.

hack/common.mk
sadlerap: Fixed.

pedjak (approved):
/lgtm
/approve

PR #1079, opened 2021-12-07
Make sure to return nil value when there was an error

kadel:
### Motivation
Saving service crd (`s.crd`) when there was an error while retrieving it (`s.CrdReader(s.groupVersionResource)`) can lead to `nil` pointer problems.

This was causing problem for odo when it tried to use SBO as a library.

### Changes

Return error and don't save crd to `s.crd`.

### Testing

Added a test that covers this scenario.

### DIFF
--- a/pkg/reconcile/pipeline/context/service/service_test.go
+++ b/pkg/reconcile/pipeline/context/service/service_test.go
@@ -2,6 +2,8 @@ package service
 
 import (
 	"errors"
+	"strings"
+
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
@@ -17,7 +19,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/dynamic/fake"
 	"k8s.io/client-go/testing"
-	"strings"
 )
 
 var _ = Describe("Service", func() {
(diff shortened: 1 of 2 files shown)


pkg/reconcile/pipeline/context/service/service_test.go
```diff
@@ -90,6 +91,22 @@ var _ = Describe("Service", func() {
...
+		crdResource := &unstructured.Unstructured{}
+		crdResource.SetName("crdfoo1")
+		builer := NewBuilder(typeLookup).WithClient(client).WithCrdReader(func(gvk *schema.GroupVersionResource) (*unstructured.Unstructured, error) {
+			return crdResource, errors.New("some error")
+		})
+		service, err := builer.Build(&unstructured.Unstructured{})
+		Expect(err).NotTo(HaveOccurred())
+
+		res, err := service.CustomResourceDefinition()
+		Expect(err).To(HaveOccurred())
```
pedjak, line 106: the error should occur and should be the very same that get return in line 100.

pedjak (changes requested):
looks good - we need just to tighten error assertion in the new test.

pedjak (approved):
/lgtm please squash commits

kadel:
> /lgtm please squash commits

done. thx for review.

pedjak (approved):
/lgtm
/approve

PR #1080, opened 2021-12-09
Introduce sample application (spring-petclinic + postgresql) to be used in quickstart

pmacik:
### Motivation

Current quickstart uses only REST API server part of the Spring Petclinic as an application and Crunchy Postgresql Operator backed DB. It has couple of UX issues.

The current application:
* does not have UI so user must use `kubectl port-forward` and `curl` command to send requests to the application's API server and analyze responses for accessing and verifying that the application works as expected
* user has to manually run a script to initialize database with application schema and data (as one of the quickstart's steps)

The current database:
* requires installation of Crunchy Postgresql Operator (via OLM) that wouldn't be possible for user with kubernetes cluster without OLM
* uses an operator that is not available on other then `x86_64` architectures, which excludes all users on alternative architectures

### Changes

This PR:
* Introduces an alternative Sprint Petclinic application that improves UX:
    * has a UI for user to interact with
    * initializes database schema and data automatically, so user does not to do it manually - thus simlify the quickstart scenario
    * is based on https://github.com/spring-projects/spring-petclinic application enhanced by https://github.com/spring-cloud/spring-cloud-bindings library.
    * is multi-arch
* Introduces an alternative PostgreSQL deployment that:
    * is in a for of plain `Deployment` based on `postgres:13` image that is multi-arch
    * does not require OLM
* Both app and db including binding is tested on `minikube` and `openshift` clusters
* Is followed-up by https://github.com/redhat-developer/service-binding-operator/pull/1085
    * These app and db deployments are supposed to replace the ones used in SBO's quickstart eventually.

### Testing

The following steps involve running some convenient `Makefile` targets that serve here only to ease testing and will be replaced by the actual `kubectl` or `oc` commands and explained in the quickstart, eventually.

0. Set your environment
    * with minikube (default)
    `export KUBERNETES_RUNTIME=minikube`
    * with OpenShift
    `export KUBERNETES_RUNTIME=openshift`

1. Go to `samples/app/spring-petclinic` directory
`cd samples/app/spring-petclinic`

2. Create user namespace (`my-postgresql` by default)
`make namespace`

3. Deploy database Deployment
`make deploy-database`

4. Expose DB binding data by setting annotations on database Deployment
`make annotate-database`

5. Deploy application Deployment
`make deploy-app`
URL for the application UI is printed out so open it to access the application, but application's pod is stuck in `Init` state due unavailable DB.

6. Bind application to db
`make bind`

7. After the application pod is restarted, the application's pod is `Running` UI should show the web interface.

### DIFF
--- a/samples/apps/spring-petclinic/Dockerfile.app
+++ b/samples/apps/spring-petclinic/Dockerfile.app
@@ -1,0 +1,26 @@
+# ---
+FROM --platform=amd64 maven:3.8.4-openjdk-11 as builder
+
+ARG PETCLINIC_REPO=https://github.com/spring-projects/spring-petclinic.git
+ARG PETCLINIC_COMMIT=a7439c74ea718c4f59fe6c7c643c4afe59d7e718
+ENV PETCLINIC_DIR=/petclinic
+
+RUN git clone "${PETCLINIC_REPO}" ${PETCLINIC_DIR} && \
+    git --git-dir=${PETCLINIC_DIR}/.git reset --hard "${PETCLINIC_COMMIT}"
+
+COPY patch /patch
+
+RUN for i in /patch/*.patch; do echo " -> Applying $i"; git -C ${PETCLINIC_DIR} apply --verbose $i; done
+
+WORKDIR ${PETCLINIC_DIR}
+
+RUN mvn package -DskipTests -Dmaven.artifact.threads=4
+
+# ---
+FROM registry.redhat.io/ubi8/openjdk-11-runtime as runtime
+
+COPY --from=builder /petclinic/target/spring-petclinic-*.jar /tmp/petclinic.jar
+
+EXPOSE 8080
+
+CMD ["java", "-Dorg.springframework.cloud.bindings.boot.enable=true", "-jar", "/tmp/petclinic.jar"]
--- a/samples/apps/spring-petclinic/Makefile
+++ b/samples/apps/spring-petclinic/Makefile
@@ -1,0 +1,60 @@
+PETCLINIC_REPO ?= quay.io/service-binding/spring-petclinic
+PETCLINIC_APP_IMAGE ?= $(PETCLINIC_REPO):latest
+
+NAMESPACE ?= my-postgresql
+
+MULTIARCH_PLATFORMS ?= linux/amd64,linux/arm64,linux/s390x,linux/ppc64le
+
+# one of "openshift", "minikube"
+KUBERNETES_RUNTIME ?= minikube
+
+.PHONY: namespace
+namespace:
+ifeq ($(KUBERNETES_RUNTIME), openshift)
+	-oc new-project $(NAMESPACE)
+else ifeq ($(KUBERNETES_RUNTIME), minikube)
+	-kubectl create namespace $(NAMESPACE)
+	kubectl config set-context --current --namespace=$(NAMESPACE)
+endif
+
+.PHONY: deploy-database
+deploy-database:
+	kubectl apply -f db-deployment.yaml -n $(NAMESPACE)
+
+.PHONY: annotate-database
+annotate-database:
+	kubectl annotate --overwrite=true deployment -n ${NAMESPACE} spring-petclinic-postgresql \
+	service.binding/type="postgresql" \
+    service.binding/host="path={.metadata.name}.{.metadata.namespace}" \
+    service.binding/port="path={.spec.template.spec.containers[0].ports[0].containerPort}" \
+    service.binding="path={.spec.template.spec.volumes[0].secret.secretName},objectType=Secret"
+
+.PHONY: build-app
+build-app:
+	docker build -f Dockerfile.app -t $(PETCLINIC_APP_IMAGE) .
+
+.PHONY: push-app
+push-app:
+	docker push $(PETCLINIC_APP_IMAGE)
+
+.PHONY: build-app-multiarch
+build-app-multiarch:
+	docker buildx build --platform "$(MULTIARCH_PLATFORMS)" -f Dockerfile.app -t $(PETCLINIC_APP_IMAGE) .
+
+.PHONY: push-app-multiarch
+push-app-multiarch:
+	docker buildx build --push --platform "$(MULTIARCH_PLATFORMS)" -f Dockerfile.app -t $(PETCLINIC_APP_IMAGE) .
+
+.PHONY: deploy-app
+deploy-app:
+	PETCLINIC_APP_IMAGE=$(PETCLINIC_APP_IMAGE) \
+	NAMESPACE=$(NAMESPACE) \
+	$(KUBERNETES_RUNTIME)/deploy-app.sh
+
+.PHONY: bind
+bind:
+	kubectl apply -f service-binding.yaml -n $(NAMESPACE)
+
+.PHONY: unbind
+unbind:
+	kubectl delete -f service-binding.yaml -n $(NAMESPACE)
--- a/samples/apps/spring-petclinic/db-deployment.yaml
+++ b/samples/apps/spring-petclinic/db-deployment.yaml
@@ -1,0 +1,70 @@
+---
+# tag::db-secret[]
+apiVersion: v1
+kind: Secret
+metadata:
+  name: spring-petclinic-postgresql
+stringData:
+  database: spring-petclinic-postgresql-db
+  username: spring-petclinic-postgresql-user
+  password: spring-petclinic-postgresql-passwd
+# end::db-secret[]
+---
+# tag::db-deployment[]
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: spring-petclinic-postgresql
+  labels:
+    app: spring-petclinic-postgresql
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: spring-petclinic-postgresql
+  template:
+    metadata:
+      labels:
+        app: spring-petclinic-postgresql
+    spec:
+      containers:
+        - name: postgresql
+          image: postgres:13
+          imagePullPolicy: IfNotPresent
+          env:
+            - name: POSTGRES_DB_FILE
+              value: /secrets/database
+            - name: POSTGRES_PASSWORD_FILE
+              value: /secrets/password
+            - name: POSTGRES_USER_FILE
+              value: /secrets/username
+            - name: PGDATA
+              value: /tmp/data
+          volumeMounts:
+            - name: spring-petclinic-postgresql
+              mountPath: "/secrets"
+              readOnly: true
+          ports:
+            - name: postgresql
+              containerPort: 5432
+      volumes:
+        - name: spring-petclinic-postgresql
+          secret:
+            secretName: spring-petclinic-postgresql
+# end::db-deployment[]
+---
+# tag::db-service[]
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: spring-petclinic-postgresql
+  name: spring-petclinic-postgresql
+spec:
+  ports:
+    - port: 5432
+      protocol: TCP
+      targetPort: 5432
+  selector:
+    app: spring-petclinic-postgresql
+# end::db-service[]
--- a/samples/apps/spring-petclinic/app-deployment.yaml
+++ b/samples/apps/spring-petclinic/app-deployment.yaml
@@ -1,0 +1,46 @@
+---
+# tag::app-deployment[]
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: spring-petclinic
+  labels:
+    app: spring-petclinic
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: spring-petclinic
+  template:
+    metadata:
+      labels:
+        app: spring-petclinic
+    spec:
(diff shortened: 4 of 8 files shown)


pmacik:
@yselkowitz could you please take a look in terms of multi-arch? thx

samples/apps/spring-petclinic/.patch/application-postgresql.properties.patch
```diff
@@ -0,0 +1,31 @@
...
+index 80bd3e1..0000000
+--- a/src/main/resources/application-postgres.properties
++++ /dev/null
+@@ -1,6 +0,0 @@
+-database=postgres
+-spring.datasource.url=${POSTGRES_URL:jdbc:postgresql://localhost/petclinic}
+-spring.datasource.username=${POSTGRES_USER:petclinic}
+-spring.datasource.password=${POSTGRES_PASS:petclinic}
+-# SQL is written to be idempotent so this is safe
+-spring.datasource.initialization-mode=always
```
pedjak, line 12: why did we drop these lines? If I understand correctly here: https://docs.spring.io/spring-boot/docs/2.1.0.M1/reference/html/howto-database-initialization.html#howto-database-initialization 

> Spring Boot automatically creates the schema of an embedded DataSource. This behaviour can be customized by using the spring.datasource.initialization-mode property. For instance, if you want to always initialize the DataSource regardless of its type:
>
> spring.datasource.initialization-mode=always

Is this something what we want actually to happen and with that we can drop that initiContainer?

samples/apps/spring-petclinic/.patch/pom.xml.patch
```diff
@@ -0,0 +1,69 @@
...
+ 
+   <dependencies>
+@@ -46,6 +48,10 @@
+       <groupId>org.springframework.boot</groupId>
+       <artifactId>spring-boot-starter-data-jpa</artifactId>
+     </dependency>
++    <dependency>
++      <groupId>org.springframework.boot</groupId>
++      <artifactId>spring-boot-starter-jdbc</artifactId>
++    </dependency>
```
pedjak, line 21: I thought that we need only to add `spring-cloud-bindings` lib? Where do we need this?

samples/apps/spring-petclinic/.patch/pom.xml.patch
```diff
@@ -0,0 +1,69 @@
...
+     </dependency>
++    <dependency>
++      <groupId>org.springframework.data</groupId>
++      <artifactId>spring-data-jdbc-core</artifactId>
++      <version>${spring-data-jdbc.version}</version>
++      <exclusions>
++        <exclusion>
++          <groupId>org.springframework</groupId>
++          <artifactId>*</artifactId>
++        </exclusion>
```
pedjak, line 49: same question as above?

samples/apps/spring-petclinic/.patch/pom.xml.patch
```diff
@@ -0,0 +1,69 @@
...
+       </snapshots>
+     </repository>
++    <repository>
++      <snapshots>
++        <enabled>false</enabled>
++      </snapshots>
++      <id>spring-libs-private-release</id>
++      <name>libs-private-release</name>
++      <url>https://repo.spring.io/artifactory/libs-private-release</url>
++    </repository>
```
pedjak, line 66: why do we need this?

samples/apps/spring-petclinic/Dockerfile.app
```diff
@@ -0,0 +1,25 @@
+FROM maven:3.8.4-openjdk-11 as builder
+
+ARG PETCLINIC_REPO=https://github.com/spring-projects/spring-petclinic.git
+ARG PETCLINIC_COMMIT=ce626da705302e71771b7e22429d1cc3209e7ebb
+
+ENV PETCLINIC_DIR /workspace/petclinic.git
+
+WORKDIR ${PETCLINIC_DIR}
+
+RUN git clone ${PETCLINIC_REPO} ${PETCLINIC_DIR} && \
```
pedjak, line 10: It seems to me that the code is checked out in `${PETCLINIC_DIR}/${PETCLINIC_DIR}` ?

samples/apps/spring-petclinic/db-deployment.yaml
```diff
@@ -0,0 +1,64 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: spring-petclinic-postgresql
+stringData:
+  database: petclinic
+  username: petclinic
```
pedjak, line 8: since these values are same as those set in `application-postgres.properties` file, how do we know that they are really consumed from binding projection?

samples/apps/spring-petclinic/init-database.sh
```diff
@@ -0,0 +1,28 @@
...
+        sleep infinity
+    else
+        export PG_HOST=${PG_HOST:-$(cat $BINDING/host)}
+        export PG_PORT=${PG_PORT:-$(cat $BINDING/port)}
+        export PG_NAME=${PG_NAME:-$(cat $BINDING/database)}
+        export PG_USER=${PG_USER:-$(cat $BINDING/username)}
+        export PGPASSWORD=${PGPASSWORD:-$(cat $BINDING/password)}
+
+        if [[ -z "$(psql -h $PG_HOST -p $PG_PORT -U "$PG_USER" "$PG_NAME" -c "SELECT tablename FROM pg_catalog.pg_tables" | grep pets)" ]]; then
+            curl -LO https://raw.githubusercontent.com/spring-petclinic/spring-petclinic-rest/master/src/main/resources/db/postgresql/initDB.sql
```
pedjak, line 20: if we really need to keep this init container (see my comment about auto schema creation above), we should download `.sql` files during container build process and have them in container ready, or even better commit them into our repo.. because the version on `spring-petclinic-rest` master branch can change over time and break us at the runtime.

samples/apps/spring-petclinic/.minikube/deploy-app.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+
+NAMESPACE=${NAMESPACE:-my-postgresql}
+
+sed -e 's,REPLACE_PETCLINIC_APP_IMAGE,'${PETCLINIC_APP_IMAGE}',g' app-deployment.yaml | \
+sed -e 's,REPLACE_PETCLINIC_APP_DBINIT_IMAGE,'${PETCLINIC_APP_DBINIT_IMAGE}',g' | \
```
pedjak, line 6: the rest of the project uses `kustomize`, so if really needed we could use that instead of `sed`

samples/apps/spring-petclinic/.minikube/deploy-app.sh
```diff
@@ -0,0 +1,13 @@
...
+kubectl apply -f - -n ${NAMESPACE} --wait
+kubectl delete service spring-petclinic --ignore-not-found=true -n ${NAMESPACE}
```
pedjak, line 8: why is this needed?

samples/apps/spring-petclinic/Makefile
```diff
@@ -0,0 +1,72 @@
...
+namespace:
+ifeq ($(KUBERNETES_RUNTIME), openshift)
+	-oc new-project $(NAMESPACE)
+else ifeq ($(KUBERNETES_RUNTIME), minikube)
+	-kubectl create namespace $(NAMESPACE)
+	kubectl config set-context --current --namespace=$(NAMESPACE)
+endif
+
+.PHONY: install-sbo
+install-sbo:
```
pedjak, line 21: do we really need to have here a target where we install SBO? If needed, we can refer directly `install.sh` from the repo itself, and not download it from `master` branch.

pedjak (changes requested):
additionally, please avoid folder names starting with a dot, e.g. `.minikube`, `.openshift`.. they are hidden by default from shell and this can create some confusions.

samples/apps/spring-petclinic/Dockerfile.app
sadlerap: Since `${PETCLINIC_DIR}` is an absolute path (it starts with `/`), `git clone` will clone into `${PETCLINIC_DIR}` (`/workspace/petclinic.git`), not `${PETCLINIC_DIR}/${PETCLINIC_DIR}`.

samples/apps/spring-petclinic/.patch/pom.xml.patch
pmacik: This maven repo is where `spring-cloud-binding` library is available

pedjak: hm, it sounds like some private repo, not official one.. how about that we use `https://repo.spring.io/releases` instead?

samples/apps/spring-petclinic/.patch/pom.xml.patch
pmacik: this was needed for `spring-data-jpa` profile used - not needed anymore

samples/apps/spring-petclinic/.patch/pom.xml.patch
pmacik: Good catch! I missed that, I assumed the release repo was already there so i looked for it elsewhere. But it wasn't. I'll use the release repo instead. thx

samples/apps/spring-petclinic/db-deployment.yaml
pmacik: good point, I've change changed it to someting different to be sure

samples/apps/spring-petclinic/init-database.sh
pmacik: the init container is not needed, after all. The only reason for it was that the standard way of initializing db via JPA failed when combined with the autoconfiguration via spring-cloud-bindings and when I wanted the application to run even without the DB... I dropped the idea based on the feedback (that the application is expected to crash with no db) and so now the standard way from spring works and therefore the init container is not necessary

samples/apps/spring-petclinic/.minikube/deploy-app.sh
pmacik: this was to be able to run that multiple times for the developement... if you try to expose a deployement as a service that already exists, it would fail... but I changed that - I included the service in the app-deployment.yaml where it is applied, rather then created.

pmacik: so no need for the deletion - removed

samples/apps/spring-petclinic/Makefile
pmacik: this makefile target is for convenience, no need to have it here, thx

samples/apps/spring-petclinic/.patch/pom.xml.patch
pmacik: you are right, when I tried the things around I used `spring-data-jpa` and that required additional jdbc dependencies... but it now works without it... no need for these dependencies - removed. Thx.

pmacik:
Sent a pull-request https://github.com/spring-projects/spring-petclinic/pull/887 to enable spring cloud bindings in the petclinic application to avoid the need of patching the app source while building the app container.

samples/apps/spring-petclinic/build-app.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
pedjak, line 1: should not we wrap this logic inside of a `Dockerfile`, so that we do not need to prepare build environment locally. I think you had this previously?

pmacik: indeed, I'll do it eventually, but this builds much faster so it allows me going forward faster while I'm still working on it.

samples/apps/spring-petclinic/patch/application-postgresql.properties.patch
```diff
@@ -0,0 +1,4 @@
+diff --git a/src/main/resources/application-postgres.properties b/src/main/resources/application-postgresql.properties
+similarity index 100%
+rename from src/main/resources/application-postgres.properties
+rename to src/main/resources/application-postgresql.properties
```
pedjak, line 4: is renaming really needed?

pmacik: yes, the spring-cloud-binding requires the type to be `postgresql` and currently it is just `postgres`

pmacik: well it didn't work before, but it was some number of attempts ago... let me check at the current state

pmacik: Ok, no need for that, after all... reverting

pmacik: Thx @pedjak!

samples/apps/spring-petclinic/build-app.sh
pmacik: done

pedjak (approved):
/lgtm

baijum (approved):
/lgtm

pmacik:
The `Bind test application to Postgres instance provisioned by Cloud Native Postgres operator` test fails for OpenShift 4.10  because the `Cloud Native Postgres` operator is not available in 4.10 catalog.

PR #1081, opened 2021-12-10
Fixed exposed host binding item for databases provisioned by Cloud Native Postgres Operator

### LINKED ISSUE #1077: Bad data with edg postgress operator created databse

OpenShift

1.1.0

  1. Install EDB operator
  2. Created cluster called edb-cluster
  3. Bind DeploymentConfig to cluster
 ```shell
spec:
  application:
    group: apps.openshift.io
    kind: DeploymentConfig
    name: nodejs-rest-http-crud
    version: v1
  bindAsFiles: true
  services:
    - group: postgresql.k8s.enterprisedb.io
      kind: Cluster
      name: edb-cluster
      version: v1
```
  4. Look at data that is bound in:

The contents of the `host` file is `edb-cluster`, however that does not seem to be reachable while `edb-cluster-rw` is.  

The contents of the `host` file can be passed to the client as a `host` that can be used to connect.

The contents of the `host` file cannot be used to connect.

pedjak:
The appropriate database host name is stored in `.status.writeService` field, and it is NOT
equal to `Postgres` resource name.

Changes:
* servicebinding host annotation got updated
* added `provider` binding, the value is set to `enterprisedb`
* generic test application is extended with `/postgres-ready` endpoint that
returns `OK` if the connection to Postgres database can be established using the provided bindings
* New version of the test application got pushed to the registry for all supported arches
* Cloud Native Postgres acceptance test got update to assert the database connection

Fixes #1077

### DIFF
--- a/pkg/binding/registry/registry.go
+++ b/pkg/binding/registry/registry.go
@@ -59,7 +59,8 @@ func New() Registry {
 			},
 			schema.GroupVersionKind{Group: "postgresql.k8s.enterprisedb.io", Version: "v1", Kind: "Cluster"}: {
 				"service.binding/type":     "postgresql",
-				"service.binding/host":     "path={.metadata.name}",
+				"service.binding/host":     "path={.status.writeService}",
+				"service.binding/provider": "enterprisedb",
 				"service.binding":          "path={.metadata.name}-{.spec.bootstrap.initdb.owner},objectType=Secret",
 				"service.binding/database": "path={.spec.bootstrap.initdb.database}",
 			},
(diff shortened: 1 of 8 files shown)


pedjak:
@pmacik could provsion an ARM based cluster and the generic test application works there as well?

baijum (approved):
/lgtm

pmacik:
> @pmacik could provsion an ARM based cluster and the generic test application works there as well?

yes, I'll do that

pmacik:
I can confirm, that new generic test app works on `arm64` cluster

Issue #1078, opened 2021-12-01
Missing setup with crunchy postgresQL database created with cruchy operator


mhdawson:
### What is the environment (Minikube, Openshift)?
OpenShift

### What is the SBO version used?
1.0.0

### What are the steps to reproduce this issue?
  1. Install Crunch PostgresQL operator
  2. Create database with operator in application namespace
  3. Drag from DeploymentConfig to create service binding
  4. Connect with the data provided in the service binding files

### What is the expected behaviour?

Connection succeeds.

### What is the actual behaviour?

Connection fails with:

``` shell
[1638397659933] INFO (24 on nodejs-rest-http-crud-17-8fqt4): Listening on port 8080
error: no pg_hba.conf entry for host "10.217.0.194", user "crunchy2", database "crunchy2", SSL off
    at Parser.parseErrorMessage (/opt/app-root/src/node_modules/pg-protocol/dist/parser.js:287:98)
    at Parser.handlePacket (/opt/app-root/src/node_modules/pg-protocol/dist/parser.js:126:29)
    at Parser.parse (/opt/app-root/src/node_modules/pg-protocol/dist/parser.js:39:38)
    at Socket.<anonymous> (/opt/app-root/src/node_modules/pg-protocol/dist/index.js:11:42)
    at Socket.emit (events.js:400:28)
    at addChunk (internal/streams/readable.js:290:12)
    at readableAddChunk (internal/streams/readable.js:265:9)
    at Socket.Readable.push (internal/streams/readable.js:204:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:188:23) {
  length: 154,
  severity: 'FATAL',
  code: '28000',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'auth.c',
  line: '503',
  routine: 'ClientAuthentication'
}
```

I expected the service binding operator to do all required setup so that the database can be accessed.

pedjak:
> I expected the service binding operator to do all required setup so that the database can be accessed.

SBO just projects as files a number of bindings exposed by Crunchy Postgres operator and the application/runtime now needs to know where to find and to read them to construct database connection objects. Here is the acceptance tests that verifies what is accessible to the app:

[...]

PR #1084, opened 2021-12-17
Project certificates together with other Crunchy Postgres bindings

pedjak:
Crunchy operator provisions Postgres instances and creates at the same certificates that might be required for some clients.

* added service binding annotation for locating and exposing certificates
* acceptance test updated to assert if postgres connection can be really established
* acceptance test update to assert if the certifactes are projected
* added `provider` binding with value `crunchydata`
* fixed the bug in `test/acceptance/resources/apps/sbo-generic-test-app/postgres-ready.sh` - password can contain special characters and hence needs to be urlencoded before constructing Postgres connection url.
* `polling2` library updated to 0.5.0 so that polling decorator can be used to implement `assert_file_exist` method in `test/acceptance/features/steps/generic_testapp.py`

Related to #1078


baijum (approved):
/lgtm

PR #1089, opened 2022-01-10
Upgrade cert manager from v1.1.0 to v1.6.0

sadlerap:
### Motivation

We need this to enable local testing on `minikube` clusters; cert-manager's `v1beta1` resources were deprecated.

### Changes

Updates the `deploy-cert-manager` make rule to use cert-manager v1.6.0.

### Testing

Deploy a cluster using `make deploy`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,7 +128,7 @@ uninstall: manifests kustomize
 
 .PHONY: deploy-cert-manager
 deploy-cert-manager:
-	kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.1.0/cert-manager.yaml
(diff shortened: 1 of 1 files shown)


pmacik:
The `Bind test application to Postgres instance provisioned by Cloud Native Postgres operator` test fails for OpenShift 4.10  because the `Cloud Native Postgres` operator is not available in 4.10 catalog.

pedjak:
> The Bind test application to Postgres instance provisioned by Cloud Native Postgres operator test fails for OpenShift 4.10 because the Cloud Native Postgres operator is not available in 4.10 catalog.

Merged manually

PR #1093, opened 2022-01-21
Avoid expression evaluation in the script

baijum:
Ref.
https://stackoverflow.com/questions/22697688/how-to-cat-eof-a-file-containing-code

The `stable` text is getting evaluted and producing an error.


pedjak (approved):
/lgtm
/approve

PR #1100, opened 2022-02-10
Remove unused CRDs and mock objects

baijum:
Signed-off-by: Baiju Muthukadan <baiju.m.mail@gmail.com>


pedjak:
this files are not used anywhere at all, they are leftovers from some early code versions, and IMHO it is safe to remove them fully instead. /cc @pmacik

pmacik:
@pedjak is correct. Those are a remnance of the past not used any longer and can be deleted.
Additionally the following pieces that used to depend on it can also be removed:
* `test/acceptance/features/steps/etcdoperator.py` file
* `test/acceptance/features/steps/etcdcluster.py` file
* the following step definitions in `test/acceptance/features/steps/steps.py`:
  * `Etcd operator running`
  * `Etcd cluster "{etcd_name}" is running`
* `test/acceptance/features/steps/postgres_db.py` file
* all related references/functions in `test/mocks/mocks.go`
* all related references/functions in `test/mocks/fake.go`

baijum:
Thanks @pedjak !

@pedjak I have removed the unused CRDs and mock objects. Also updated the PR title.

test/acceptance/features/steps/steps.py
```diff
@@ -13,13 +13,10 @@
 
 from behave import given, register_type, then, when, step
 from dboperator import DbOperator
```
pmacik, line 15: @baijum This should be also removed along with the class and step:
see the https://github.com/baijum/service-binding-operator/pull/2

pmacik (changes requested):

test/acceptance/features/steps/steps.py
baijum: Thanks for the PR. I have merged it.

PR #1102, opened 2022-02-15
Remove the last usage of obsolete package

baijum:
The github.com/pkg/errors package is no more maintaned:
https://github.com/pkg/errors/issues/245

Signed-off-by: Baiju Muthukadan <baiju.m.mail@gmail.com>

# Changes

# Submitter Checklist

### DIFF
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -128,7 +128,6 @@ github.com/operator-framework/api/pkg/lib/version
 github.com/operator-framework/api/pkg/operators
 github.com/operator-framework/api/pkg/operators/v1alpha1
 # github.com/pkg/errors v0.9.1
-## explicit
 github.com/pkg/errors
 # github.com/pmezard/go-difflib v1.0.0
 github.com/pmezard/go-difflib/difflib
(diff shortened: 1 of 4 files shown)


vendor/modules.txt
```diff
@@ -128,7 +128,6 @@ github.com/operator-framework/api/pkg/lib/version
 github.com/operator-framework/api/pkg/operators
 github.com/operator-framework/api/pkg/operators/v1alpha1
 # github.com/pkg/errors v0.9.1
-## explicit
 github.com/pkg/errors
```
pmacik, line 131: Shouldn't these be also removed?

baijum: @pmacik That package is transitive dependency. For example `github.com/evanphx/json-patch` and `go.uber.org/zap` use it. Again these packages are also transitive dependencies. So, only the explicit dependency has been removed.

pmacik:
@baijum please rebase on top of the current master to rebuild-the images for acceptance tests to match the acceptance tests scenarios. Otherwise, the acceptance tests would keep failing.

PR #1109, opened 2022-02-24
Disable scenarios for architectures other then amd64

pmacik:
Signed-off-by: Pavel Macík <pavel.macik@gmail.com>

# Changes

Currently, all acceptance tests assume that SBO to be installed in Kubernetes or OpenShift clusters deployed on `amd64` machines.

While SBO itself can run on the alternative architectures, some of the acceptance tests scenarios cannot.

This PR:
* Introduces tags to diable scenarios/feature for various alternative architectures (`@disable.arch.ppc64le`, `@disable.arch.s390x`, `@disable.arch.arm64`)
* Marks scenarios that cannot run on respective architectures with those tags.

To disable those scenarios, append the following into `EXTRA_BEHAVE_ARGS` env variable for that acceptance test run: 
```
EXTRA_BEHAVE_ARGS="... --tags=~@disable.arch.<arch> ..."
```

# Submitter Checklist

### DIFF
--- a/test/acceptance/features/gettingStartedGuide.feature
+++ b/test/acceptance/features/gettingStartedGuide.feature
@@ -21,6 +21,9 @@ Feature: Getting Started Guide
     * Service Binding is deleted
 
   @olm
+  @disable.arch.ppc64le
+  @disable.arch.s390x
+  @disable.arch.arm64
   Scenario: Connecting PetClinic application to an Operator-backed PostgreSQL database
     Given Crunchy Data Postgres operator is running
     * PetClinic sample application is installed
--- a/test/acceptance/features/supportExistingOperatorBackedServices.feature
+++ b/test/acceptance/features/supportExistingOperatorBackedServices.feature
@@ -1,5 +1,8 @@
 @olm
 @supported-operator
+@disable.arch.ppc64le
+@disable.arch.s390x
+@disable.arch.arm64
 Feature: Support a number of existing operator-backed services out of the box
 
   As a user of Service Binding operator


baijum:
@pmacik 
Instead of using these kinds of deny-lists. Can we use allow-lists? For example `@arch.x86_64`.
So, this way no need to change test feature files again when a new unsupported platform comes up.

I am not should how to specify the command-line arguments with allow-list. If that is possible, please consider the above suggestion.

sadlerap:
Do we need to have this for v1.0.1?

pmacik:
> @pmacik Instead of using these kinds of deny-lists. Can we use allow-lists? For example `@arch.x86_64`. So, this way no need to change test feature files again when a new unsupported platform comes up.
> 
> I am not should how to specify the command-line arguments with allow-list. If that is possible, please consider the above suggestion.

@baijum we could use allow-list approach but then we would need to tag each and every one of the feature with the `@arch.amd64`, `@arch.ppc64le`, `@arch.s309x` and `@arch.arm64` tags (including any new feature or scenario that would be added) and only those that are not supported for the alternative architectures would lack the respective arch tags.

That way we could set the `--tags=@arch.<arch>` to run those on the particular architecture such as `ppc64le`.

But that approach would me much more verbose and would require more maintanence.

The current idea and way of the acceptance tests is that we want to run as much of them as possible (ideall all of them) by default to cover the as much of the scenarios/behavior as possible and only if we need to skip some, we would mark them as `@disabled...` and exclude it by that tag for that particular run.

We use the same approach for [excluding tests for a particular versions of openshift](https://github.com/redhat-developer/service-binding-operator/pull/1090)

pmacik:
> Do we need to have this for v1.0.1?

@sadlerap Ideally, yes so we can have a matching set of scenarios and operator, and since there is still some time before, I'd say let's include it if approved and merged.

This update can be used by IBM Z and P users to run the tests on those platforms

PR #1111, opened 2022-03-08
This is our PostgreSQL Service Endpoint Definition (SED)

dperaza4dustbit:
# Changes

This SED can be used in cases where the service was provisioned
either manually or via a Helm Chart. It will allow developer to
test their binding data in a standard manner regardless of how the
PostgreSQL database was provisioned or the backend cloud. The
assumption here is that there is no CR representing the service in
kubernetes.

/kind enhancement

# Submitter Checklist

- [x] [Docs](https://github.com/redhat-developer/service-binding-operator/blob/master/CONTRIBUTING.md#docs) 

### DIFF
--- a/charts/seds/psql-sed/Chart.yaml
+++ b/charts/seds/psql-sed/Chart.yaml
@@ -1,0 +1,19 @@
+apiVersion: v2
+name: psql-sed
+description: A Helm chart for PostgreSQL Service Endpoint Definition (SED)
+
+
+type: application
+
+version: 1.0.0
+
+
+appVersion: "1.0.0"
+
+kubeVersion: ">=1.20.0"
+
+annotations:
+  charts.openshift.io/provider: RedHat
+  charts.openshift.io/name: PostgreSQL Service Endpoint Definition (SED)
+  charts.openshift.io/supportURL: https://github.com/dperaza4dustbit/helm-chart
+  charts.openshift.io/archs: x86_64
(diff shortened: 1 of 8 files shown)


charts/seds/psql-sed/Chart.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+
+appVersion: "1.0.0"
+
+kubeVersion: ">=1.20.0"
+
+annotations:
+  charts.openshift.io/provider: RedHat
+  charts.openshift.io/name: PostgreSQL Service Endpoint Definition (SED)
+  charts.openshift.io/supportURL: https://github.com/dperaza4dustbit/helm-chart
```
dperaza4dustbit, line 18: What should we put for this URL on SBO?

pmacik: So, it would be something like `https://redhat-developer.github.io/service-binding-operator/userguide/service-endpoint-definition/psql.html`

pmacik: This PR should also include adding a section `Service Endpoint Definition` in the User's guide explaining the concept of SED itself.

dperaza4dustbit: Agree

baijum:
@dperaza4dustbit This PR can be closed, right?

dperaza4dustbit:
Created new project for SEDs https://github.com/redhat-developer/service-endpoint-definition/tree/main/seds/psql-sed

Issue #1083, opened 2021-12-15
Wish to bind to 'sso' / 'keycloak'


sbose78:
Do we have samples which show how to bind an application(preferably a java application)  to keycloak ?

growi:
I was playing around with keycloak and created a sample binding for the petclinic application. 
The example can be found [here](https://github.com/growi/service-binding-operator/tree/feature/keycloak-sample/samples/apps/spring-petclinic) .  To get it to work you need to build the petclinic image or use [mine](https://hub.docker.com/r/growi83/spring-petclinic-oauth2) (I extended the patch to include the oauth2 starter). The app will auto-config if the servicebinding is bound.  Otherwise a password is generated on startup. 
For the keycloak login flow to work it is REQUIRED that the keycloak ingress is accessible through http://keycloak.local .
The keycloak deployment creates a realm "Test" and the confidential client "petclinic". No realm users will be created.

If there is interest I can create a clean PR to integrate the example.

baijum:
This looks good to me. Please send PR.

growi:
I created PR #1122 that integrates the sample.

filariow:
Closed based on community discussion for December the 1st

PR #1122, opened 2022-03-26
Add keycloak sample for spring-petclinic app 

growi:
related issue: #1083 

# Changes
/kind enhancement

This sample provides a binding of the spring-petclinic application to a backing Keycloak as an Identity Provider. After the successful projection of the binding values the application will be auto-configured for user authentication via the OpenId Connect authorization code flow. The sample includes a deployment for the Keycloak server that creates a realm 'Test' on startup. Users need to be created manually. The necessary admin credentials are defined in the deployment. The sample also includes an ingess definition for the Keycloak server. The ingress MUST be DNS resolvable for 'keycloak.local' for the sample to work.

# Side Effects

## Samples using the petclinic app will require authentication
Since the petclinic application has been extended to utilize the spring-boot-starter-oauth2-client that in turn includes Spring Security the auto-configuration in absence of the binding values will result in a random password being generated on startup. This password will be written to the log and provides an alternative means of authentication for the also auto-generated user 'user'. This affects all other samples that use the extended petclinic application.
This could be resolved by restructuring the samples. Otherwise it needs to be documented [here](https://redhat-developer.github.io/service-binding-operator/userguide/getting-started/petclinic-postgresql.html).  

##  Change to Dockerfile.app (petclinic app)
I made the following substitution :
```
- FROM registry.redhat.io/ubi8/openjdk-11-runtime as runtime
+ FROM registry.access.redhat.com/ubi8/openjdk-11-runtime as runtime
```
Since I have no access to registry.redhat.io . 

# Possible Enhancements

- Right now the auto-configuration fails fatally if the binding values get projected but the projected issuer-url is not resolvable. It should be possible to extend the Keycloak config with external vs. internal urls to guarantee that the issuer-url is always accessed through the clusters internal dns resolution.
- By utilizing the Keycloak Operator it should be possible to map all necessary binding values without annotating the deployment. Only realm and clientId  are probably still necessary even if it is conceivable that for those some kind of convention could be established. 

# Submitter Checklist

### DIFF
--- a/samples/apps/spring-petclinic/Dockerfile.app
+++ b/samples/apps/spring-petclinic/Dockerfile.app
@@ -17,7 +17,7 @@ WORKDIR ${PETCLINIC_DIR}
 RUN mvn package -DskipTests -Dmaven.artifact.threads=4
 
 # ---
-FROM registry.redhat.io/ubi8/openjdk-11-runtime as runtime
+FROM registry.access.redhat.com/ubi8/openjdk-11-runtime as runtime
 
 COPY --from=builder /petclinic/target/spring-petclinic-*.jar /tmp/petclinic.jar
 
--- a/samples/apps/spring-petclinic/Makefile
+++ b/samples/apps/spring-petclinic/Makefile
@@ -17,6 +17,10 @@ else ifeq ($(KUBERNETES_RUNTIME), minikube)
 	kubectl config set-context --current --namespace=$(NAMESPACE)
 endif
 
+.PHONY: deploy-keycloak
+deploy-keycloak:
+	kubectl apply -f keycloak-deployment.yaml -n $(NAMESPACE)
+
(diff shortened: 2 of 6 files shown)


baijum:
> This affects all other samples that use the extended petclinic application.
> This could be resolved by restructuring the samples.

@growi Please create a separate sample app project without modifying the existing PetClinic application. You can create a new directory, say "samples/apps/spring-petclinic-keycloak" and place all these files there.

baijum:
> > This affects all other samples that use the extended petclinic application.
> > This could be resolved by restructuring the samples.
> 
> @growi Please create a separate sample app project without modifying the existing PetClinic application. You can create a new directory, say "samples/apps/spring-petclinic-keycloak" and place all these files there.

@growi I am closing the PR for now. Feel free to reopen it when you are working on it.

PR #1130, opened 2022-04-13
bind labeled workloads created after sb reconcile

sadlerap:
# Changes

/kind enhancement

Workloads that we bound with label selectors won't get bound if they
were created after the service binding was first reconciled.  As a
workaround, periodically reconcile service bindings with label selectors
to pick up any new workloads.

This uses an exponential backoff strategy, which is the default for when
a duration before re-reconciling is not specified.  A fixed interval was
considered, but was decided against out of concerns for thundering herd
problems in a cluster.

# Submitter Checklist

- [x] All acceptance test scenarios included in the PR which verifies a bugfix or a requested feature reported by a non-member are tagged with `@external-feedback` tag.

### DIFF
--- a/docs/userguide/modules/creating-service-bindings/pages/binding-options.adoc
+++ b/docs/userguide/modules/creating-service-bindings/pages/binding-options.adoc
@@ -183,3 +183,58 @@ spec:
 <1>  Database service
 <2> Optional "id" field
 <3>  Translation service
+
+== Binding workloads from a label selector
+
+Sometimes, it may be useful to specify the workload being bound by using a label selector.  For
+instance, you may want to bind a service to every workload with the label `environment: production`
+set.  SBO is able to support this kind of binding by being able to bind with label selectors.  As an
+example, to project a single secret into multiple workloads, set up your application field like so:
+[source,yaml]
+----
+apiVersion: binding.operators.coreos.com/v1alpha1
+kind: ServiceBinding
+metadata:
+  name: multi-application-binding
+  namespace: service-binding-demo
+spec:
+  application:
+    labelSelector:
+      matchLabels:
+        environment: production
+    group: apps
+    version: v1
+    resource: deployments
+  services:
+    group: ""
+    version: v1
+    kind: Secret
+    name: super-secret-data
+----
+
+Or, using the spec API:
+[source,yaml]
+----
+apiVersion: servicebindings.io/v1alpha3
+kind: ServiceBinding
+metadata:
+  name: multi-application-binding
+  namespace: service-binding-demo
+spec:
+  workload:
+    selector:
+      matchLabels:
+        environment: production
+    apiVersion: app/v1
+    kind: Deployment
+  service:
+    apiVersion: v1
+    kind: Secret
+    name: super-secret-data
+----
+
+If a service binding is declared using these label selectors to pick up workloads, SBO will
+periodically attempt to find and bind new workloads that match the given label selector.
+
+As a note: it is currently forbidden to attempt a binding with both `name:` and `labelSelector`
+defined (or `selector` in the spec API).  Any attempt to do so will result in a error.
--- a/pkg/reconcile/pipeline/builder/pipeline.go
+++ b/pkg/reconcile/pipeline/builder/pipeline.go
@@ -2,6 +2,8 @@ package builder
 
 import (
 	"fmt"
+	"time"
+
 	"github.com/redhat-developer/service-binding-operator/pkg/reconcile/pipeline"
 	"github.com/redhat-developer/service-binding-operator/pkg/reconcile/pipeline/handler/collect"
 	"github.com/redhat-developer/service-binding-operator/pkg/reconcile/pipeline/handler/mapping"
@@ -16,7 +18,7 @@ type impl struct {
 	handlers    []pipeline.Handler
 }
 
-func (i *impl) Process(binding interface{}) (retry bool, err error) {
+func (i *impl) Process(binding interface{}) (retry bool, delay time.Duration, err error) {
 	defer func() {
 		if perr := recover(); perr != nil {
 			retry = true
@@ -25,7 +27,7 @@ func (i *impl) Process(binding interface{}) (retry bool, err error) {
 	}()
 	ctx, err := i.ctxProvider.Get(binding)
 	if err != nil {
-		return false, err
+		return false, time.Duration(0), err
 	}
 	var status pipeline.FlowStatus
 	for _, h := range i.handlers {
@@ -37,9 +39,9 @@ func (i *impl) Process(binding interface{}) (retry bool, err error) {
 	}
 	err = ctx.Close()
 	if err != nil {
-		return true, err
+		return true, time.Duration(0), err
 	}
-	return status.Retry, status.Err
+	return status.Retry, status.Delay, status.Err
 }
 
 type builder struct {
--- a/pkg/reconcile/pipeline/handler/project/impl.go
+++ b/pkg/reconcile/pipeline/handler/project/impl.go
@@ -49,6 +49,10 @@ func PreFlightCheck(mandatoryBindingKeys ...string) func(pipeline.Context) {
 
 func PostFlightCheck(ctx pipeline.Context) {
 	ctx.SetCondition(apis.Conditions().InjectionReady().Reason("ApplicationUpdated").Build())
+	if ctx.HasLabelSelector() {
+		// Force periodic reprocessing of this service binding to catch new workloads
+		ctx.DelayReprocessing(nil)
+	}
 }
 
 func InjectSecretRef(ctx pipeline.Context) {
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"flag"
+
 	"github.com/redhat-developer/service-binding-operator/pkg/client/kubernetes"
 
 	"github.com/go-logr/logr"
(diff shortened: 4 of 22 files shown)


pkg/reconcile/pipeline/handler/project/impl.go
```diff
@@ -52,6 +52,10 @@ func PreFlightCheck(mandatoryBindingKeys ...string) func(pipeline.Context) {
 
 func PostFlightCheck(ctx pipeline.Context) {
 	ctx.SetCondition(apis.Conditions().InjectionReady().Reason("ApplicationUpdated").Build())
+	if ctx.HasLabelSelector() {
```
pedjak, line 55: iirc, `context.Applications()` return error in case when label selector is used and no workload is found: 

https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/reconcile/pipeline/context/impl.go#L255 

That error is checked in `PreFlightCheck`: 

https://github.com/redhat-developer/service-binding-operator/blob/master/pkg/reconcile/pipeline/context/impl.go#L255

iirc, retrying should happen even with the current code?

sadlerap: The intent of this patch is to pick up workloads that have labels *after* we've reconciled the service binding.  We're triggering a retry here _not_ because we failed to find any workloads, but because we succeeded and want to periodically rebind to get informed of new workloads with the labels we're interested in.  This is why I've left the error condition blank and put this in `PostFlightCheck` (i.e. the end of our pipeline).

As far as I know, operator-sdk doesn't let us install new reconcilers (especially on arbitrary resource types), so ideally, we'd be using a watch of some kind in a background task of some kind to get these resources and debounce them as they come in.  This patch is more of a workaround until we can properly do these kind of watches.

> iirc, retrying should happen even with the current code?

Correct, it does.  That addresses the error case when no workloads were found.

This patch is handling a separate case: we succeeded, and we've bound workloads that match a label selector.  However, we weren't binding new workloads that also match, which is required by the spec for us to support.

pedjak: > We're triggering a retry here not because we failed to find any workloads, but because we succeeded and want to periodically rebind to get informed of new workloads with the labels we're interested in. 

I understand the intention, but not sure if this is going to work as wished in practice. Check this excellent [blog post](https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/) describing in details how backoff strategies work. Simply, very quickly, reconciliation periods will become very large and there is also upper bound for number of tries, so at some point the reconciliation will happen after a number of hours, and eventually stop, and newly created workloads will not be bound.

sadlerap: > Simply, very quickly, reconciliation periods will become very large and there is also upper bound for number of tries, so at some point the reconciliation will happen after a number of hours, and eventually stop, and newly created workloads will not be bound.

The way I understand it, the worst we'll see for a reconciliation delay by default is 16:40, which is long but not _super_ unreasonable.  If we used separate parameters on the rate-limiting scheme, we could tailor this down to be more frequent (with slight performance penalty).  Something like this:

```go
// SetupWithManager sets up the controller with the Manager.
func (r *BindingReconciler) SetupWithManager(mgr ctrl.Manager) error {
	pipeline, err := r.PipelineProvider(mgr.GetConfig(), kubernetes.ResourceLookup(mgr.GetRESTMapper()))
	if err != nil {
		return err
	}
	r.pipeline = pipeline
	return ctrl.NewControllerManagedBy(mgr).
		For(r.ReconcilingObject()).
		WithEventFilter(predicate.GenerationChangedPredicate{}).
		WithOptions(
			controller.Options{
				MaxConcurrentReconciles: MaxConcurrentReconciles,
				RateLimiter: workqueue.NewMaxOfRateLimiter(
					&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(10), 100)},
					// set max timeout to 5 minutes (down from default of 1000 seconds)
					workqueue.NewItemExponentialFailureRateLimiter(5*time.Millisecond, 5*time.Minute),
				),
			}).
		Complete(r)
}
```

I'm also not convinced we're going to run into the upper bound.  I think that bound is implemented [here](https://github.com/kubernetes/client-go/blob/20732a1bc198ab57de644af498fa75e73fa44c08/examples/workqueue/main.go#L103), and we only run into that bound if we were reporting errors.  I explicitly set `err = nil` on those cases, so I don't think that this will be a problem.

pedjak: > The way I understand it, the worst we'll see for a reconciliation delay by default is 16:40, which is long but not super unreasonable.

Well, we speak here about 16mins, and IMHO from user perspective this is very long, at least user need to be aware that this is the time when the newly added workload will get the bindings projected.

sadlerap: I've configured re-reconciliation times as a tunable parameter with a default of 2 minutes (chosen rather arbitrarily).  Since re-reconciliation has a performance impact, making this tunable by a cluster administrator seemed to be the right way to go.

pedjak: I am not sure if cluster admin has any way to discover and tune that parameter at the installation time. OLM does not support that. Also, even if admins want to update the deployment later on, the changes will be reverted by OLM. IMHO validate this change against a large cluster with a lot of service binding (both those that can be bound and others that cannot) /cc @pmacik

pedjak: BTW, you can implement your own `RateLimiter` where you can decide based on service binding item how long to wait for the next reconciliation. This way service bindings that declare workloads via labels can have constant retry times and do not interfere with other faulty service bindings.

baijum (approved):
/lgtm

pkg/reconcile/pipeline/handler/project/impl.go
pedjak: Even if you want to make the re-reconciliation time as tunable parameter, I would not bake this into a controller CLI arg - how about that you can control that per service binding, by applying an annotation on a given binding at the creation time, something like:

```yaml
kind: ServiceBinding
metadata:
  annotations:
     servicebinding.io/re-reconcilation-delays: 2m
.
.
.
```

Anyhow, this PR should contain as well an update to UserDoc describing the new behavior.

pkg/reconcile/pipeline/builder/pipeline.go
```diff
@@ -25,10 +27,11 @@ func (i *impl) Process(binding interface{}) (retry bool, err error) {
 	}()
 	ctx, err := i.ctxProvider.Get(binding)
 	if err != nil {
-		return false, err
+		return false, time.Duration(0), err
 	}
 	var status pipeline.FlowStatus
-	for _, h := range i.handlers {
+	for i, h := range i.handlers {
+		fmt.Printf("%v\n", i)
```
baijum, line 34: Forgot to remove it? I guess this is added for debugging.

sadlerap: Yep, added this for debugging.  Should be fixed.

pmacik:
/lgtm

I run performance evaluation and it does not show any serious issues with performance.

sadlerap:
I think something is up with prow/4.8 acceptance testing.  It looks like there are some timeouts, but not from anything this PR has touched.

pmacik:
@sadlerap @dperaza4dustbit There is a glitch in the OpenShift CI currently causing the job to fail. However, the acceptance test suite for 4.8 itself actually passed. If the the `4.8-acceptance` PR check job is the only thing that prevents mergin this PR, the QE approves to merge it (with administrator permissions).

docs/userguide/modules/creating-service-bindings/pages/binding-options.adoc
```diff
@@ -182,4 +182,59 @@ spec:
...
+    version: v1
+    resource: deployments
+  services:
+    group: ""
+    version: v1
+    kind: Secret
+    name: super-secret-data
+----
+
+Or, using the spec API:
```
Srivaralakshmi, line 215: Replace L215 with this:
.Example of `ServiceBinding` CR in the `servicebinding.io` API:

docs/userguide/modules/creating-service-bindings/pages/binding-options.adoc
```diff
@@ -182,4 +182,59 @@ spec:
...
-<3>  Translation service
\ No newline at end of file
+<3>  Translation service
+
+== Binding workloads from a label selector
+
+Sometimes, it may be useful to specify the workload being bound by using a label selector.  For
+instance, you may want to bind a service to every workload with the label `environment: production`
+set.  SBO is able to support this kind of binding by being able to bind with label selectors.  As an
+example, to project a single secret into multiple workloads, set up your application field like so:
```
Srivaralakshmi, line 192: Replace L187 to L192 with this:
== Binding workloads using a label selector
You can use a label selector to specify the workload that is being bound. If you declare a service binding using these label selectors to pick up workloads, the {servicebinding-title} periodically attempts to find and bind new workloads that match the given label selector.

For example, you may want to bind a service to every application with the `environment: production` label set in the `application` field. Doing so projects a single secret into multiple workloads.

.Example of `ServiceBinding` CR in the `binding.operators.coreos.com/v1alpha1` API:

docs/userguide/modules/creating-service-bindings/pages/binding-options.adoc
```diff
@@ -182,4 +182,59 @@ spec:
...
+    apiVersion: v1
+    kind: Secret
+    name: super-secret-data
+----
+
+If a service binding is declared using these label selectors to pick up workloads, SBO will
+periodically attempt to find and bind new workloads that match the given label selector.
+
+As a note: it is currently forbidden to attempt a binding with both `name:` and `labelSelector`
+defined (or `selector` in the spec API).  Any attempt to do so will result in a error.
```
Srivaralakshmi, line 240: Replace L236 to L240 with this:

[IMPORTANT]
====
Currently, it is forbidden to attempt a binding with the following fields defined:

* `name:` and `labelSelector` in the `binding.operators.coreos.com/v1alpha1` API
* `name:` and `selector` in the `servicebinding.io` API (Spec API)

Any attempt to do so will result in an error.
====

Srivaralakshmi (commented):
@sadlerap Left my suggestions. Please incorporate them by raising another PR. Thanks!

PR #1132, opened 2022-04-25
Service Binding Operator Helm chart

Kartikey-star:
This PR defines the Service Binding Operator Helm chart.

### DIFF
--- a/charts/service-binding-operator/Chart.yaml
+++ b/charts/service-binding-operator/Chart.yaml
@@ -1,0 +1,25 @@
+apiVersion: v2
+name: service-binding-operator
+description: A Helm chart to deploy service binding operator
+type: application
+version: 1.0.0
+appVersion: "1.0.1"
+kubeVersion: ">= 1.19.2-0"
+keywords:
+  - ""
+links:
+  - name: Documentation
+    url: https://redhat-developer.github.io/service-binding-operator
+maintainers:
+  - name: Kartikey-star
+    email: kmamgain@redhat.com
+  - name: pmacik
+    email: pmacik@redhat.com
+  - name: jasperchui
+    email: jchui@redhat.com
+  - name: dperaza4dustbit
+    email: dperaza@redhat.com
+annotations:
+  charts.openshift.io/name: service-binding-operator
+  charts.openshift.io/provider: RedHat
+  charts.openshift.io/supportURL: https://github.com/redhat-developer/service-binding-operator.git
--- a/charts/service-binding-operator/README.md
+++ b/charts/service-binding-operator/README.md
@@ -1,0 +1,105 @@
+
+# Service Binding Operator Helm Chart
+
+This Helm chart defines the Service Binding Operator. You can install Service Binding Operators using this Helm chart. 
+
+Installing the Service Binding Operator Helm chart creates the following custom resource definitions (CRDs):
+- bindablekinds.binding.operators.coreos.com
+- servicebindings.binding.operators.coreos.com
+- servicebindings.servicebinding.io
+
+The resources required for the Service Binding Operator will also be installed.
+
+## Introduction
+
+The values.yaml file contains the following values that can be customized when installing the chart:
+
+- `image.pullPolicy`
+- `image.repository`
+- `image.testRepository`
+- `keepTestResources`
+
+A user can define values for the image  PullPolicy. 
+A user can define values for `image.repository` and `image.testRepository`. If user is not able to pull image from quay.io registry, they can copy the image  to their own container registry.
+As part of Helm test we delete the deploymemt,service binding resource and secret used for testing. If a user is interested to view them, then he has to install the chart with keepTestResources set to `true`.
+
+
+## Helm Chart Installation 
+
+The Helm chart installation involves the following steps:
+1. Adding the `service-binding-operator-helm-chart` repository.
+2. Installing the Service Binding Operator chart.
+3. Running a Helm test.
+
+**Note:** If you are not installing the Service Binding Operator through Operator Lifecycle Manager (OLM), you must install cert-manager on the cluster. Installing the cert-manager automates TLS certificates for Kubernetes and OpenShift workloads. Cert-manager ensures that the certificates are valid and up-to-date, and attempts to renew certificates at a configured time before expiry. You can install cert-manager by running the following command:
+
+kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.6.0/cert-manager.yaml
+
+### Adding the Helm chart repository
+You need to add our helm repository to your local repository. Name the repository as per your convenience.  
+
+```
+helm repo add service-binding-operator-helm-chart https://redhat-developer.github.io/service-binding-operator-helm-chart/
+```
+
+### Installing the Helm chart
+In order to install the chart you need to search the repository, with the following command:
+
+```
+helm search repo service-binding-operator-helm-chart
+```
+```
+helm install service-binding-operator-release \
+service-binding-operator-helm-chart/service-binding-operator \
+--namespace service-binding-operator --create-namespace
+```
+Remove --namespace and --create-namespace flag if you wish to install the chart on default namespace.
+
+In order to view the resources created on helm test , set the keepTestResources to true. 
+
+```
+helm install service-binding-operator-release \
+service-binding-operator-helm-chart/service-binding-operator \
+--namespace service-binding-operator --create-namespace \
+--set keepTestResources=true
+```
+
+You can check whether the chart is succesfully installed by running the following command
+
+```
+kubectl get pods --namespace service-binding-operator
+```
+
+### Helm test
+
+In order to test the chart the user is expected to create a secret (specify the namespace if applicable), named my-k-config from his kubeconfig .
+
+**NOTE**:
+In case you are installing the chart on AWS eks cluster then you need to modify the aws-auth configmap.
+```
+kubectl edit -n kube-system cm/aws-auth 
+```
+Please add -system:masters to mapRoles and save.
+After editing the config map you need to update the eks kubeconfig
+```
+aws eks update-kubeconfig --name <cluster-name>
+```
+Then Continue with the following steps.
+
+```
+kubectl create secret generic my-k-config --from-file=kubeconfig=<PATH TO YOUR KUBECONFIG> -namespace service-binding-operator
+```
+
+Run the Helm test (specify the namespace if applicable) using :
+
+```
+helm test service-binding-operator-release --namespace service-binding-operator
+```
+
+Please ensure to delete the secret (specify the namespace if applicable) created :
+```
+kubectl delete secret my-k-config --namespace service-binding-operator
+```
+
+## Additional Help
+Please reach out to us for any additional queries by creating an issue on https://github.com/redhat-developer/service-binding-operator/issues.
--- a/charts/service-binding-operator/templates/clusterrole.yaml
+++ b/charts/service-binding-operator/templates/clusterrole.yaml
@@ -1,0 +1,343 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+  name: service-binding-bindablekinds-viewer-role
+rules:
+- apiGroups:
+  - binding.operators.coreos.com
+  resources:
+  - bindablekinds
+  verbs:
+  - get
+  - list
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    servicebinding.io/controller: "true"
+  name: service-binding-cloud-native-postgres-viewer-role
+rules:
+- apiGroups:
+  - postgresql.k8s.enterprisedb.io
+  resources:
+  - clusters
+  verbs:
+  - get
+  - list
+---
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
+      servicebinding.io/controller: "true"
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: service-binding-controller-role
+rules: []
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    servicebinding.io/controller: "true"
+  name: service-binding-crunchy-postgres-viewer-role
+rules:
+- apiGroups:
+  - postgres-operator.crunchydata.com
+  resources:
+  - postgresclusters
+  verbs:
+  - get
+  - list
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+  name: service-binding-editor-role
+rules:
+- apiGroups:
+  - binding.operators.coreos.com
+  - servicebinding.io
+  resources:
+  - servicebindings
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - binding.operators.coreos.com
+  - servicebinding.io
+  resources:
+  - servicebindings/status
+  verbs:
+  - get
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: service-binding-manager-role
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - endpoints
+  - pods
+  - secrets
+  - services
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  - secrets
+  verbs:
+  - patch
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - create
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - apps
+  resources:
+  - daemonsets
+  - deployments
+  - replicasets
+  - statefulsets
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+- apiGroups:
+  - apps.openshift.io
+  resources:
+  - deploymentconfigs
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+- apiGroups:
+  - authentication.k8s.io
+  resources:
+  - tokenreviews
+  verbs:
+  - create
+- apiGroups:
+  - authorization.k8s.io
+  resources:
+  - selfsubjectaccessreviews
+  verbs:
+  - create
+- apiGroups:
+  - authorization.k8s.io
+  resources:
+  - subjectaccessreviews
+  verbs:
+  - create
+- apiGroups:
+  - binding.operators.coreos.com
+  resources:
+  - bindablekinds
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - binding.operators.coreos.com
+  resources:
+  - bindablekinds/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - binding.operators.coreos.com
+  resources:
+  - bindablekinds/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
+  - binding.operators.coreos.com
+  resources:
+  - servicebindings
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - binding.operators.coreos.com
+  resources:
+  - servicebindings/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - binding.operators.coreos.com
+  resources:
+  - servicebindings/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
+  - operators.coreos.com
+  resources:
+  - clusterserviceversions
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - route.openshift.io
+  resources:
+  - routes
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - servicebinding.io
+  resources:
+  - servicebindings
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - servicebinding.io
+  resources:
+  - servicebindings/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - servicebinding.io
+  resources:
+  - servicebindings/status
+  verbs:
+  - get
+  - patch
+  - update
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    servicebinding.io/controller: "true"
+  name: service-binding-opstree-redis-viewer-role
+rules:
+- apiGroups:
+  - redis.redis.opstreelabs.in
+  resources:
+  - redis
+  verbs:
+  - get
+  - list
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    servicebinding.io/controller: "true"
+  name: service-binding-percona-mongodb-view
+rules:
+- apiGroups:
+  - psmdb.percona.com
+  resources:
+  - perconaservermongodbs
+  - perconaservermongodbs/status
+  verbs:
+  - get
+  - list
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    servicebinding.io/controller: "true"
+  name: service-binding-percona-mysql-viewer-role
+rules:
+- apiGroups:
+  - pxc.percona.com
+  resources:
+  - perconaxtradbclusters
+  verbs:
+  - get
+  - list
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    servicebinding.io/controller: "true"
+  name: service-binding-rabbitmq-operator-viewer-role
+rules:
+- apiGroups:
+  - rabbitmq.com
+  resources:
+  - rabbitmqclusters
+  verbs:
+  - get
+  - list
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+  name: service-binding-viewer-role
+rules:
+- apiGroups:
+  - binding.operators.coreos.com
+  - servicebinding.io
+  resources:
+  - servicebindings
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - binding.operators.coreos.com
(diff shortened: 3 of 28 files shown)


charts/service-binding-operator/templates/tests/test-service-binding-operator-connection.yaml
```diff
@@ -0,0 +1,45 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ .Release.Name }}-test"
+  annotations:
+    "helm.sh/hook": test-success
+spec:
+  containers:
+    - name: "{{ .Release.Name }}-test"
+      image: "quay.io/kmamgain/sboat:latest"
```
pmacik, line 11: Please include the `Dockerfile` for this image

Kartikey-star: The Dockerfile and shell script cannot be kept in the charts directory . Can u suggest some other directory to place those.

charts/service-binding-operator/templates/clusterrole.yaml
```diff
@@ -0,0 +1,343 @@
...
+  verbs:
+  - get
+  - list
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+  name: service-binding-editor-role
```
baijum, line 63: I don't see this role used in role binding. Also there is aggregation role for the given label: `rbac.authorization.k8s.io/aggregate-to-edit`.

charts/service-binding-operator/templates/clusterrole.yaml
```diff
@@ -0,0 +1,343 @@
...
+  - servicebindings/status
+  verbs:
+  - get
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: service-binding-manager-role
+rules:
```
baijum, line 91: How about using an aggregation rule label instead of creating a separate role and role binding?

charts/service-binding-operator/templates/clusterrole.yaml
```diff
@@ -0,0 +1,343 @@
...
+  verbs:
+  - get
+  - list
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+  name: service-binding-viewer-role
```
baijum, line 325: I don't see this role used in role binding. Also there is aggregation role for the given label: `rbac.authorization.k8s.io/aggregate-to-view`.

charts/service-binding-operator/templates/clusterrole.yaml
Kartikey-star: It is present in release yaml for v1.0.1 too so i guess we can keep it as it is.

charts/service-binding-operator/templates/clusterrole.yaml
Kartikey-star: It is present in release yaml for v1.0.1 too so i guess we can keep it as it is.

charts/service-binding-operator/templates/clusterrole.yaml
Kartikey-star: It is present in release yaml for v1.0.1 too so i guess we can keep it as it is.

baijum: ok

baijum (approved):
/lgtm

charts/service-binding-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,66 @@
...
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+      - args:
+        - --leader-elect
+        - --zap-encoder=json
+        - --zap-log-level=info
+        command:
+        - /manager
+        image: quay.io/redhat-developer/servicebinding-operator:31151ab8
```
pedjak, line 32: is this the published 1.0.1 version? It should referred via sha, and there should be an option to pass another value from values.yaml

pedjak (commented):
IMHO, the PR needs significant improvements:
* Chart template files should be generated from config resources on demand - similar to how operator bundle is generated and `release.yaml`. Files under `/config` are source of truth  - and if there is no automation around that - Updating the chart will be hard over the time
* Chart version should be the same as the operator version, because they are both published from the same repo
* There is no CI job that verify if this chart install the operator properly. This PR should have that as well - similar to how both OLM based and release.yaml based installation mechanisms are tests cc @pmacik 
* The chart itself should offer more customization options, similar to other good charts.

charts/service-binding-operator/README.md
```diff
@@ -0,0 +1,88 @@
...
+## Helm Chart Installation 
+
+The helm chart installation involves the following steps:
+- Add the helm chart repository
+- Install the chart
+- Helm test
+
+Note: If service binding operator is not installed through OLM, the operator requires that cert-manager is available on the cluster. It can be installed by:
+
+kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.6.0/cert-manager.yaml
```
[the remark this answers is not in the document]
Kartikey-star: The cert manager documentation refers not to embed cert-manager as a sub-chart of other Helm charts.We intended do it but dropped the idea because of that reason

charts/service-binding-operator/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Add the helm chart repository
+You need to add our helm repository to your local repository. Name the repository as per your convenience.  
+
+```
+helm repo add service-binding-operator https://redhat-developer.github.io/service-binding-operator-helm-chart/
```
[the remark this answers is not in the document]
Kartikey-star: I agree but we do have other PR in place which we intend to merge simultaneously so that no incorrect information is provided.

charts/service-binding-operator/templates/deployment.yaml
Kartikey-star: Yes this is the published 1.0.1 version image. 
The ability to change the image tag is not provided because we will be rolling out different charts as and when there is a change in app version of sbo itself. So that way the need is not there.

pedjak: well then, please refer it via sha, not via tag. Also, enabling user to refer an alternative image is always super userful, due to various reasons - for example user is not able to access quay directly, so he/she needs to copy the image to some other internal registry, etc.

charts/service-binding-operator/README.md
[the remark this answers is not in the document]
pedjak: ok, but when speaking about Helm way to install, it would be good that all components are installed using Helm?

charts/service-binding-operator/README.md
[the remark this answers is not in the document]
pedjak: There is really no need to add a whole Helm chart repository for serving just SBO charts. You can publish charts as GitHub releases and install them

```
helm install sbo https://github.com/redhat-developer/service-binding-operator/releases/download/v1.0.1/chart.tgz
```

pedjak:
IMHO, chart files should not be a part of SBO repo at all, they should be generated on demand with something like `make chart`

cc @pmacik @dperaza4dustbit

charts/service-binding-operator/README.md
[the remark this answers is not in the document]
dperaza4dustbit: I agree it would be good to have as a dependency and use helm dependency update to get latest version etc... However, I also understand the concern from cert manager team, if multiple charts make cert manager a dependency there could be a scenario where more than one cert manager is installed. I think it is worth exploring with them, cause I think our chart will fail if cert manager is already installed, which then will make our chart the bad actor. Either way, something to explore on future Epics, I would say.

charts/service-binding-operator/README.md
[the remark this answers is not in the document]
dperaza4dustbit: The idea @pedjak is to add this chart to artifact hub. Which will need a repo I think, correct me if I'm wrong @Kartikey-star. We could also just add to an existing registry, helm now supports OCI, but most people are not on helm 3.8 where OCI was officially supported. So in the future the path will be to OCI and drop this repo.

charts/service-binding-operator/README.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+### Install the helm chart
+```
+helm install service-binding-operator https://redhat-developer.github.io/service-binding-operator-helm-chart/service-binding-operator
```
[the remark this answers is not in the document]
dperaza4dustbit: there is not need to do this at all, since the repo was added in previous step. I should be calling the install referencing the repo in the chart. I would add a step to call: helm search repo service-binding-operator to list the chart first, then you know how to reference it during the install

charts/service-binding-operator/templates/deployment.yaml
dperaza4dustbit: good usecase @pedjak, so, maybe we could provide the registry as optional for internal network scenario, but the image name and sha should be hardcoded since it belongs to the app version we are putting in the Chart.yaml

dperaza4dustbit: @Kartikey-star ^^^^

charts/service-binding-operator/templates/namespace.yaml
```diff
@@ -0,0 +1,7 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: service-binding-controller-manager
+  name: service-binding-operator
```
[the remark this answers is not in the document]
dperaza4dustbit: hum... I do think it is strange that this works right now, the namespace resource happens to come before all others. I would prefer we could hardcore the namespace, but I agree with @pedjak lets do it the helm way. User should have the option to install in current namespace or create one at install time. No need to put in values file though.

dperaza4dustbit:
> IMHO, chart files should not be a part of SBO repo at all, they should be generated on demand with something like `make chart`
> 
> cc @pmacik @dperaza4dustbit

I would say, it is fine for 1.0.1 to be done manually like this, and then add a CI integration EPIC that would generate at release time. But we don need the source in the repo, in case there is an issue with the chart itself we can patch it and publish right away, no need to wait for a new SBO release. In fact @isutton is working on a tool that would extract charts from running apps give labels. This could be use here as part of a make extract as chart or something like that, after we install the operator. Other ideas welcomed as well.

dperaza4dustbit (commented):
A few things to address from @pedjak review, I added my comments there for some context

charts/service-binding-operator/README.md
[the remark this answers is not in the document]
pedjak: The line is wrong because the snippet:

```
helm install service-binding-operator https://redhat-developer.github.io/service-binding-operator-helm-chart/service-binding-operator

```

states that the chart is installed from the given url - and that url should point to the chart tarball, not repo.

charts/service-binding-operator/templates/deployment.yaml
pedjak: There should be a good default, and this comes into `values.yaml`, not here - so that users can override it.

pedjak:
> I would say, it is fine for 1.0.1 to be done manually like this, and then add a CI integration EPIC that would generate at release time. 

@pmacik did you run acceptance tests against this chart?

charts/service-binding-operator/README.md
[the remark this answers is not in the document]
Kartikey-star: Yes we only need to have index.yaml and artifacthub-repo.yaml in our source repo so as to deliver it as a helm repository. We did create a separate repository for ideas like using chart-releaser and keep the publishing separate from chart source .

charts/service-binding-operator/README.md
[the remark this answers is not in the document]
Kartikey-star: No actually service-binding-operator is the name of our chart. I will rename the chart repo ,chart release so as to give better clarity.In fact since i have added repo with name service-binding-operator i can even install it with:
helm install <release-name> service-binding-operator/service-binding-operator.

charts/service-binding-operator/templates/deployment.yaml
Kartikey-star: Yes will do so.Will refer the image tag with the sha and image.repositpry will be a key in values.yaml with default as quay.io/redhat-developer.

charts/service-binding-operator/templates/namespace.yaml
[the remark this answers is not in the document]
Kartikey-star: It is working as of now because we are creating that namespace and have all resources in that namespace. Actually the helm release is happening in a different namespace (where ever the user has context set) ,but the chart workloads happen to be in service-binding-operator ns .
As a workaround i will be installing the chart as :
helm install sbo ./Projects/service-binding-operator/charts/service-binding-operator  --namespace service-binding-operator --create-namespace.
The --create-namespace param seems to be a bool flag instead of string and i believe then i can refernce the namespace field in the template with {{.Release.Nmaespace}}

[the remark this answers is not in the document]
pedjak: > It is working as of now because we are creating that namespace and have all resources in that namespace. Actually the helm release is happening in a different namespace (where ever the user has context set) ,but the chart workloads happen to be in service-binding-operator ns .

The helm release (actually the secret keeping all information about the installed chart) should be placed in the same namespace as the operator - otherwise, the user experience will become very strange - how you would ever find where your Helm release is - if it is just based on the fact what was your context at the time when the chart was installed?

docs/userguide/modules/getting-started/pages/installing-service-binding.adoc
```diff
@@ -59,3 +60,94 @@ image:sbo_install_options.png[sbo_install_options]
...
+
+```
+kubectl create secret generic my-k-config --from-file=kubeconfig=<PATH to your kubeconfig>
+```
+
+Run the helm test (specify the namespace if applicable) using :
+
+```
+helm test service-binding-operator
+```
```
Srivaralakshmi:
```suggestion
. Run the Helm test and specify the namespace if applicable:
+
[source,bash]
----
helm test service-binding-operator
----
+
.Example output
[source,terminal]
----
W0509 13:59:25.509390 1434368 warnings.go:70] would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "sbo-test" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "sbo-test" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "sbo-test" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "sbo-test" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
NAME: sbo
LAST DEPLOYED: Mon May  9 13:50:16 2022
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE:     sbo-test
Last Started:   Mon May  9 13:59:24 2022
Last Completed: Mon May  9 14:00:57 2022
Phase:          Succeeded
----
It is safe to ignore this warning now.
+
The `Succeeded` phase from the warning output indicates that the Helm test has run successfully. 

. Verify that the Helm test has run successfully:
+
[source,bash]
----
kubectl get pods
----
+
.Example output
[source,terminal]
----
NAME                                              READY   STATUS      RESTARTS   AGE
bindapptoprovisionedservice-246-d84dccdf4-s5xtv   1/1     Running     0          8m10s
sbo-test                                          0/1     Completed   0          4m28s
----
+
This implies that you have successfully installed the {servicebinding-title} using a Helm chart and are able to bind your workload to backing services.
```

Srivaralakshmi (changes requested):
@Kartikey-star Left some suggestions. PTAL. Thanks!

docs/userguide/modules/getting-started/pages/installing-service-binding.adoc
```diff
@@ -59,3 +60,94 @@ image:sbo_install_options.png[sbo_install_options]
...
+Please add -system:masters to mapRoles and save.
+After editing the config map you need to update the eks kubeconfig
+```
+aws eks update-kubeconfig --name <cluster-name>
+```
+Then Continue with the following steps.
+
+```
+kubectl create secret generic my-k-config --from-file=kubeconfig=<PATH to your kubeconfig>
+```
```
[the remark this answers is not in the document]
Kartikey-star: This verifies that the Helm chart is working successfully. This is incorrect. We only need to mount the secret to test the chart.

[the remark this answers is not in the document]
Kartikey-star: This is in a essential secret which will be containing the kubeconfig, without this we cannot test our chart.

docs/userguide/modules/getting-started/pages/installing-service-binding.adoc
```diff
@@ -59,3 +60,94 @@ image:sbo_install_options.png[sbo_install_options]
...
+helm install service-binding-operator https://redhat-developer.github.io/service-binding-operator-helm-chart/service-binding-operator
+....
+As part of helm test we clear up the deploymemt,service binding resource and secret created. If a user is interested to view them , then he 
+has to install the chart with keepTestResources set to true.
+
+[source,bash]
+....
+helm install service-binding-operator https://redhat-developer.github.io/service-binding-operator-helm-chart/service-binding-operator --set keepTestResources=true
+....
+
```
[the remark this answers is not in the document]
Kartikey-star: NAME                             	CHART VERSION	APP VERSION	DESCRIPTION                                    
kartikey/service-binding-operator	1.0.0        	1.0.1      	A Helm chart to deploy service binding operator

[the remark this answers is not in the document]
Kartikey-star: I will add sbo-helm-chart repo add output in the PR

charts/service-binding-operator/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+name: service-binding-operator
+description: A Helm chart to deploy service binding operator
+type: application
+version: 1.0.0
+appVersion: "1.0.1"
```
pedjak, line 6: similar to OLM bundle, the chart needs to specify what is minimal supported kube version, add same set of keywords and other metadata available in OLM bundle.

/hold

dperaza4dustbit: kubeVersion: '>= 1.21.0' for OCP 4.8 right? https://access.redhat.com/solutions/4870701

pedjak: CI tests with 1.19: https://github.com/redhat-developer/service-binding-operator/blob/master/.github/workflows/pr-checks.yaml#L13 so why putting the bar higher here? Also, there are plenty of useful metadata https://github.com/redhat-developer/service-binding-operator/blob/master/config/manifests/bases/service-binding-operator.clusterserviceversion.yaml that could be reused here as well, at least this is what I would expect from a chart carrying 1.0 version.

Kartikey-star: Majority of data looks like field used by OLM ,as the target for our chart is non OLM i think it is fine just to mention the kubeversion.

dperaza4dustbit:
@Kartikey-star for license we can use this one: https://github.com/redhat-developer/service-binding-operator/blob/master/LICENSE place it in an md file like this: https://github.com/openshift-helm-charts/charts/blob/main/charts/partners/hashicorp/vault/0.19.0/src/LICENSE.md

charts/service-binding-operator/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+The following are the values that can be customized when the chart is installed:
+
+- image.pullPolicy
+- image.tag
+- is_openshift
+
+The is_openshift value signifies whether installing the operator on openshift or non openshift cluster. The value is set to true for openshift and false for non openshift.
+
+A user can define values for the Tag and PullPolicy. A  User can obtain tag from https://github.com/redhat-developer/service-binding-operator/tags to get the desired version of service binding operator. The image tag can be obtained from quay.io/redhat-developer/servicebinding-operator using the tag mentioned above. For example , search quay.io/redhat-developer/servicebinding-operator using 31151ab (release tag of v1.0.1) to obtain image tag  31151ab8. By default the chart points out to the latest version.
```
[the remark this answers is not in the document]
Kartikey-star: This is not required now.

charts/service-binding-operator/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Helm Chart Installation 
+
+The helm chart installation involves the following steps:
+- Add the helm chart repository
+- Install the chart
+- Helm test
```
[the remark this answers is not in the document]
Kartikey-star: Done

charts/service-binding-operator/README.md
```diff
@@ -0,0 +1,88 @@
...
+The resources required for service binding operator will also be installed. The chart creates a service-binding-operator namespace and installs the required resources in that namespace. 
+
+## Introduction
+
+The following are the values that can be customized when the chart is installed:
+
+- image.pullPolicy
+- keepTestResources
+
+A user can define values for the image  PullPolicy. As part of helm test we clear up the deploymemt,service binding resource and secret created. If a user is interested to view them , then he has to install the chart with keepTestResources set to true.
```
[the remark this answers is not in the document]
Kartikey-star: As part of Helm test we delete the deploymemt,service binding resource and secret used for testing

docs/userguide/modules/getting-started/pages/installing-service-binding.adoc
```diff
@@ -59,3 +60,94 @@ image:sbo_install_options.png[sbo_install_options]
...
+. Test the chart
+
+Note: If service binding operator is not installed through OLM, the operator requires that cert-manager is available on the cluster. It can be installed by:
+[source,bash]
+....
+kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.6.0/cert-manager.yaml
+....
+
+[## Add the helm chart repository]
+== Add the helm chart repository
```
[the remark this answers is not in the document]
Kartikey-star: We already mention that in prerequisite stated in l36 so don't see a reason for mentioning again.

docs/userguide/modules/getting-started/pages/installing-service-binding.adoc
```diff
@@ -59,3 +60,94 @@ image:sbo_install_options.png[sbo_install_options]
...
+....
+helm install service-binding-operator https://redhat-developer.github.io/service-binding-operator-helm-chart/service-binding-operator
+....
+As part of helm test we clear up the deploymemt,service binding resource and secret created. If a user is interested to view them , then he 
+has to install the chart with keepTestResources set to true.
+
+[source,bash]
+....
+helm install service-binding-operator https://redhat-developer.github.io/service-binding-operator-helm-chart/service-binding-operator --set keepTestResources=true
+....
```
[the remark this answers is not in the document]
Kartikey-star: As part of the Helm test, the deployment is removed, and the `ServiceBinding` and `Secret` resources are created. 
. Optional: To view the `ServiceBinding` and `Secret` resources, install the chart with the `keepTestResources` flag value set to `true`:
This is incorrect

docs/userguide/modules/getting-started/pages/installing-service-binding.adoc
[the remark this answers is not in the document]
Kartikey-star: I have added :. You have access to a Kubernetes or an OpenShift Container Platform cluster using an account with cluster-admin permissions.
. You have the cert-manager installed on the cluster if you are not installing the Service Binding Operator through Operator Lifecycle Manager (OLM).
. You have Helm CLI installed.

docs/userguide/modules/getting-started/pages/installing-service-binding.adoc
```diff
@@ -59,3 +61,180 @@ image:sbo_install_options.png[sbo_install_options]
...
+
+.Example output
+[source,terminal]
+----
+secret/my-k-config created
+----
+
+The previous output verifies that the `my-k-config` secret is created.
+
+Run the Helm test (specify the namespace if applicable) using :
```
Srivaralakshmi, line 201: . Run the Helm test and specify the namespace if applicable:

docs/userguide/modules/getting-started/pages/installing-service-binding.adoc
```diff
@@ -59,3 +61,180 @@ image:sbo_install_options.png[sbo_install_options]
...
+STATUS: deployed
+REVISION: 1
+TEST SUITE:     service-binding-operator-release-test
+Last Started:   Mon May 16 11:01:10 2022
+Last Completed: Mon May 16 11:01:22 2022
+Phase:          Succeeded
+----
+The `Succeeded` phase from the output indicates that the Helm test has run successfully. 
+
+Verify that the Helm test has run successfully:
```
Srivaralakshmi, line 221: . Verify that the Helm test has run successfully:

docs/userguide/modules/getting-started/pages/installing-service-binding.adoc
```diff
@@ -59,3 +61,180 @@ image:sbo_install_options.png[sbo_install_options]
...
+----
+
+.Example output
+[source,terminal]
+----
+NAME                                              READY   STATUS      RESTARTS   AGE
+service-binding-operator-release-test             0/1     Completed   0          4m28s
+----
+
+This implies that you have successfully installed the service binding operator using a Helm chart and are able to bind your workload to backing services.
```
Srivaralakshmi, line 235: The output verifies that you have successfully installed the {servicebinding-title} using a Helm chart and are able to bind your workload to backing services.

docs/userguide/modules/getting-started/pages/installing-service-binding.adoc
```diff
@@ -59,3 +61,180 @@ image:sbo_install_options.png[sbo_install_options]
...
+
+Please ensure to delete the secret (specify the namespace if applicable) created :
```
Srivaralakshmi, line 237: . As a safety measure, delete the secret created and specify the namespace if applicable:

docs/userguide/modules/getting-started/pages/installing-service-binding.adoc
```diff
@@ -59,3 +61,180 @@ image:sbo_install_options.png[sbo_install_options]
...
+----
+kubectl delete secret my-k-config  --namespace service-binding-operator
+----
```
Srivaralakshmi, line 240: After L240, add this:

.Example output
[source,terminal]
----
secret "my-k-config" deleted
----

The output verifies that the secret that you had created is now deleted. 

Deleting the secret avoids exposing the secret credentials of the cluster to which you are connected.

charts/service-binding-operator/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+The values.yaml file contains the following values that can be customized when installing the chart:
+
+- `image.pullPolicy`
+- `image.repository`
+- `image.testRepository`
+- `keepTestResources`
+
+A user can define values for the image  PullPolicy. 
+A user can define values for `image.repository` and `image.testRepository`. If user is not able to pull image from quay.io registry, they can copy the image  to their own container registry.
```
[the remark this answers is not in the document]
Kartikey-star: looks good to change.

charts/service-binding-operator/README.md
```diff
@@ -0,0 +1,105 @@
...
+```
+kubectl create secret generic my-k-config --from-file=kubeconfig=<PATH TO YOUR KUBECONFIG> -namespace service-binding-operator
+```
+
+Run the Helm test (specify the namespace if applicable) using :
+
+```
+helm test service-binding-operator-release --namespace service-binding-operator
+```
+
```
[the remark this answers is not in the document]
Kartikey-star: This has to be done before deleting secret then.

PR #1192, opened 2022-08-01
Delete namespace after every scenario

shruthihub:
Signed-off-by: shruthihub <shruthi.shruthz@gmail.com>

# Changes

# Submitter Checklist

- [x] All acceptance test scenarios included in the PR which verifies a bugfix or a requested feature reported by a non-member are tagged with `@external-feedback` tag.

### DIFF
--- a/test/acceptance/features/environment.py
+++ b/test/acceptance/features/environment.py
@@ -14,11 +14,17 @@ before_all(context), after_all(context)
 
 from steps.command import Command
 from steps.environment import ctx
+from behave.model_core import Status
 
 import os
 import semver
 
 cmd = Command()
+
+if os.getenv("DELETE_NAMESPACE") in ["always", "never", "keepwhenfailed"]:
+    delete_namespace = os.getenv("DELETE_NAMESPACE")
+else:
+    delete_namespace = "keepwhenfailed"
 
 
 def before_all(_context):
@@ -49,9 +55,36 @@ def before_all(_context):
         _context.sbo_namespace = output
     else:
         assert False, f"TEST_ACCEPTANCE_START_SBO={start_sbo} is currently unsupported."
+    ctx.no_scenarios_failed = True
 
 
 def before_scenario(_context, _scenario):
     _context.bindings = dict()
     output, code = cmd.run(f'{ctx.cli} get ns default -o jsonpath="{{.metadata.name}}"')
     assert code == 0, f"Checking connection to OS cluster by getting the 'default' project failed: {output}"
+
+
+def after_scenario(_context, scenario):
+    if "namespace" in _context:
+        namespace = _context.namespace.name
+    elif os.getenv("TEST_NAMESPACE"):
+        namespace = os.getenv("TEST_NAMESPACE")
+    else:
+        print("No namespace set in context nor TEST_NAMESPACE env variable is set, skipping deletion")
+        return
+    if delete_namespace == "always":
+        delete_current_namespace(namespace)
+    elif (delete_namespace == "keepwhenfailed"):
+        if scenario.status == Status.passed and (ctx.no_scenarios_failed):
+            delete_current_namespace(namespace)
+        else:
+            ctx.no_scenarios_failed = False
+            print(f"Deleting namespace {namespace} skipped, since one of the scenarios failed.")
+    elif delete_namespace == "never":
+        print(f"Namespace {namespace} deletion skipped.")
+
+
+def delete_current_namespace(name):
+    output, code = cmd.run(f"{ctx.cli} delete namespace {name} --ignore-not-found --timeout=1800s")
+    assert code == 0, f"Deletion of namespace failed: {output}"
+    print(f"Namespace {name} deleted.")


shruthihub:
to add arguments to make command DELETE_NAMESPACE=True

test/acceptance/features/environment.py
```diff
@@ -55,3 +56,13 @@ def before_scenario(_context, _scenario):
     _context.bindings = dict()
     output, code = cmd.run(f'{ctx.cli} get ns default -o jsonpath="{{.metadata.name}}"')
     assert code == 0, f"Checking connection to OS cluster by getting the 'default' project failed: {output}"
+
+
+def after_scenario(context, scenario):
+    delete_namespace=True
+    delete_namespace=os.getenv("DELETE_NAMESPACE")
+    namespace = os.getenv("TEST_NAMESPACE")
+    if (scenario.status == Status.passed) and (delete_namespace):
```
pmacik, line 65: Please consider a situation when the current scenarios passed, but one of the previously executed scenarios failed. In such case we don't want to delete the whole namespace because we would loose the resources for investigating or debugging the failed scenario.

The framework needs to remember that any scenario executed before the current one failed along with checking the current scenario's status and base the decision about the cleaning on that.

test/acceptance/features/environment.py
```diff
@@ -55,3 +56,13 @@ def before_scenario(_context, _scenario):
...
+        output, code = cmd.run(f"{ctx.cli} delete namespace {namespace}")
+    else:
+        assert False, f"Deleting namespace {namespace} failed since the scenario has {scenario.status}."
```
pmacik, line 68: This should not fail if the current scenario passed, just skip the cleanup with a log message.

pmacik: it causes all the tests to fail in CI

test/acceptance/features/environment.py
```diff
@@ -55,3 +62,26 @@ def before_scenario(_context, _scenario):
...
     output, code = cmd.run(f'{ctx.cli} get ns default -o jsonpath="{{.metadata.name}}"')
     assert code == 0, f"Checking connection to OS cluster by getting the 'default' project failed: {output}"
+
+
+def after_scenario(context, scenario):
+    global scenario_status
+    context.scenario_status = scenario_status
+    namespace = os.getenv("TEST_NAMESPACE")
+    if delete_namespace == "alwaysdelete":
+        if scenario.status == Status.passed:
```
pmacik, line 72: If you want to "always delete", it does not make sense to check for scenario status... in this configuration the namespace must be deleted no matter what (even when scenario failed).

test/acceptance/features/environment.py
```diff
@@ -55,3 +62,26 @@ def before_scenario(_context, _scenario):
...
+    elif (delete_namespace == "keepwhenfailed") and (context.scenario_status):
+        if scenario.status == Status.passed:
+            output, code = cmd.run(f"{ctx.cli} delete namespace {namespace}.")
+            print(f"Namespace {namespace} deleted.")
+        else:
+            scenario_status = False
+            print(f"Deleting namespace {namespace} skipped hereafter, since the scenario has failed.")
+
+    elif delete_namespace == "neverdelete":
+        print(f"Namespace {namespace} deletion not supported.")
```
pmacik, line 87: The message is confusing... it should say something like "namespace deletion skipped". 

"deletion is not supported" does not make sense in our context... the deletion is indeed supported (that is the point of this PR), but is disabled by configuration.

test/acceptance/features/environment.py
```diff
@@ -55,3 +62,26 @@ def before_scenario(_context, _scenario):
...
+            print(f"Namespace {namespace} deleted.")
+        else:
+            print(f"Deleting namespace {namespace} skipped since the scenario has failed.")
+
+    elif (delete_namespace == "keepwhenfailed") and (context.scenario_status):
+        if scenario.status == Status.passed:
+            output, code = cmd.run(f"{ctx.cli} delete namespace {namespace}.")
+            print(f"Namespace {namespace} deleted.")
+        else:
+            scenario_status = False
```
pmacik, line 83: You are setting the global variable but not using it in any condition or decision making code :)

test/acceptance/features/environment.py
```diff
@@ -14,12 +14,19 @@
...
 import os
 import semver
 
 cmd = Command()
 
+if os.getenv("DELETE_NAMESPACE") in ["alwaysdelete", "neverdelete", "keepwhenfailed"]:
+    delete_namespace = os.getenv("DELETE_NAMESPACE")
+else:
+    delete_namespace = "alwaysdelete"
+scenario_status = True
```
pmacik, line 28: Instead of using a plain global variable, please use the `Environment` class (`test/acceptance/features/steps/environment.py`) and add it there... you could then access it via global `ctx` similarly as `ctx.cli`

test/acceptance/features/environment.py
```diff
@@ -55,3 +62,26 @@ def before_scenario(_context, _scenario):
...
     assert code == 0, f"Checking connection to OS cluster by getting the 'default' project failed: {output}"
+
+
+def after_scenario(context, scenario):
+    global scenario_status
+    context.scenario_status = scenario_status
+    namespace = os.getenv("TEST_NAMESPACE")
+    if delete_namespace == "alwaysdelete":
+        if scenario.status == Status.passed:
+            output, code = cmd.run(f"{ctx.cli} delete namespace {namespace}.")
```
pmacik, line 73: Please check the exit code of the `cmd` to handle potential errors - like in here https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/environment.py#L27

test/acceptance/features/environment.py
```diff
@@ -55,3 +62,26 @@ def before_scenario(_context, _scenario):
...
+            print(f"Namespace {namespace} deleted.")
+        else:
+            print(f"Deleting namespace {namespace} skipped since the scenario has failed.")
+
+    elif (delete_namespace == "keepwhenfailed") and (context.scenario_status):
+        if scenario.status == Status.passed:
+            output, code = cmd.run(f"{ctx.cli} delete namespace {namespace}.")
```
pmacik, line 80: Please check the exit code of the `cmd` to handle potential errors - like in here https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/environment.py#L27

test/acceptance/features/environment.py
shruthihub: since all scenarios are independent im assinging the global variable to context.scenario_status  like context.scenario_status = scenario_status.
so im using the context.scenario_status to check for the keepwhenfailed scenario.
if i do not use context then there is linting errors too.

pmacik: Right, I missed that... however, the use of this global variable is redundant and not neccessary, in that case. You can use `context.scenario_status` (or `context.all_scenarios_passed` as suggested in another comment) directly. In such case you can set the initial value `True` in the `before_all` hook function.

pmacik: or use the `Environment` class as suggested in the other comment instead.

test/acceptance/features/environment.py
pmacik: Or better `no_scenarios_failed` which is more accurate

test/acceptance/features/environment.py
```diff
@@ -49,9 +55,28 @@ def before_all(_context):
...
+
+def after_scenario(_context, scenario):
+    namespace = os.getenv("TEST_NAMESPACE")
+    if delete_namespace == "alwaysdelete":
+        output, code = cmd.run(f"{ctx.cli} delete namespace {namespace}.")
+        assert code == 0, f"Deletion of namespace failed: {output}"
+        print(f"Namespace {namespace} deleted.")
+    elif (delete_namespace == "keepwhenfailed"):
+        if scenario.status == Status.passed and (_context.no_scenarios_failed):
+            output, code = cmd.run(f"{ctx.cli} delete namespace {namespace}.")
```
pmacik, line 75: Typo - (the `.` dot) - at the end of the namespace name

test/acceptance/features/environment.py
```diff
@@ -49,9 +55,28 @@ def before_all(_context):
...
 def before_scenario(_context, _scenario):
     _context.bindings = dict()
     output, code = cmd.run(f'{ctx.cli} get ns default -o jsonpath="{{.metadata.name}}"')
     assert code == 0, f"Checking connection to OS cluster by getting the 'default' project failed: {output}"
+
+
+def after_scenario(_context, scenario):
+    namespace = os.getenv("TEST_NAMESPACE")
+    if delete_namespace == "alwaysdelete":
+        output, code = cmd.run(f"{ctx.cli} delete namespace {namespace}.")
```
pmacik, line 70: Typo - (the `.` dot) - at the end of the namespace name

test/acceptance/features/environment.py
```diff
@@ -49,9 +55,28 @@ def before_all(_context):
...
 
 
 def before_scenario(_context, _scenario):
     _context.bindings = dict()
     output, code = cmd.run(f'{ctx.cli} get ns default -o jsonpath="{{.metadata.name}}"')
     assert code == 0, f"Checking connection to OS cluster by getting the 'default' project failed: {output}"
+
+
+def after_scenario(_context, scenario):
+    namespace = os.getenv("TEST_NAMESPACE")
```
pmacik, line 68: The `TEST_NAMESPACE` env variable is not always the namespace in which the scenario is executed.

All scenarios use the `Namespace "name" is used` or `Namespace [ENV] is used` step in the background which sets the `context.namespace` attribute.

You should first check if that is set (by `if "namespace" in _context:`) and use that value, otherwise use the `TEST_NAMESPACE` env variable as the default.

Thx

shruthihub: hi 
do u mean something like this

shruthihub: if(os.getenv("TEST_NAMESPACE")):
        namespace = os.getenv("TEST_NAMESPACE")
    elif "namespace" in _context):
        namespace = ****** ->> to be filled
    else: 
        print(f"No namespace set in context skipping deletion")

test/acceptance/features/environment.py
```diff
@@ -49,9 +55,28 @@ def before_all(_context):
...
+    if delete_namespace == "alwaysdelete":
+        output, code = cmd.run(f"{ctx.cli} delete namespace {namespace}.")
+        assert code == 0, f"Deletion of namespace failed: {output}"
+        print(f"Namespace {namespace} deleted.")
+    elif (delete_namespace == "keepwhenfailed"):
+        if scenario.status == Status.passed and (_context.no_scenarios_failed):
+            output, code = cmd.run(f"{ctx.cli} delete namespace {namespace}.")
+            assert code == 0, f"Deletion of namespace failed: {output}"
+            print(f"Namespace {namespace} deleted.")
```
pmacik, line 77: Please do not repeat the code for deleting the namespace. You can extract it to a separate function and then call it instead, for instance.

test/acceptance/features/environment.py
pmacik: Rather something like:
```python
    if "namespace" in _context:
        namespace = _context.namespace.name
    elif os.getenv("TEST_NAMESPACE"):
        namespace = os.getenv("TEST_NAMESPACE")
    else:
        print(f"No namespace set in context nor TEST_NAMESPACE env variable is set, skipping deletion")
        return
```
The namespace in context should have precedence - after all it is supposed to be set by the "Namespace ... is used" step in the scenario, so it should be the one cleaned.

test/acceptance/features/environment.py
pmacik: Additionally, if by any chance the namespace is not there, it should not fail the test. After all, the point of this is to ensure that the namespace is not there after each scenario.

```suggestion
            output, code = cmd.run(f"{ctx.cli} delete namespace {namespace} --ignore-not-found")
            assert code == 0, f"Deletion of namespace failed: {output}"
            print(f"Namespace {namespace} deleted.")
```

test/acceptance/features/environment.py
pmacik: FYI, the `_context.namespace` is not a `string` but an instance of the `Namespace` class ([`test/acceptance/features/steps/namespace.py`](https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/steps/namespace.py)) - See https://github.com/redhat-developer/service-binding-operator/blob/master/test/acceptance/features/steps/steps.py#L40

test/acceptance/features/environment.py
```diff
@@ -49,9 +55,35 @@ def before_all(_context):
...
+    if "namespace" in _context:
+        namespace = _context.namespace.name
+    elif os.getenv("TEST_NAMESPACE"):
+        namespace = os.getenv("TEST_NAMESPACE")
+    else:
+        print(f"No namespace set in context nor TEST_NAMESPACE env variable is set, skipping deletion")
+        return
+    
+    if delete_namespace == "alwaysdelete":
+        output, code = cmd.run(f"{ctx.cli} delete namespace {namespace} --ignore-not-found.")
```
pmacik, line 77: Typo (the `.`) in the command
```suggestion
        output, code = cmd.run(f"{ctx.cli} delete namespace {namespace} --ignore-not-found")
```

test/acceptance/features/environment.py
```diff
@@ -14,12 +14,18 @@
...
 
 import os
 import semver
 
 cmd = Command()
 
+if os.getenv("DELETE_NAMESPACE") in ["alwaysdelete", "neverdelete", "keepwhenfailed"]:
+    delete_namespace = os.getenv("DELETE_NAMESPACE")
+else:
+    delete_namespace = "alwaysdelete"
```
pmacik, line 27: @shruthihub please make the `keepwhenfailed` the default value.

pmacik:
# Observation
Since an overhead (in terms of time) is expected to be added to finish running acceptance tests (after all, cleaning the namespace after each senarios takes some amount of time) I compared the time of respective CI steps to run the acceptance tests  before this PR (I used #1204 PR as baseline) and after this PR. The follwing table shows the results (time is in `H:M:S` format):

Acceptance Testing Job | Environment | Before (with #1204) | After (with this PR) | Overhead
--- | --- | --- | --- | ---
kubernetes-with-olm | minikube |  00:31:08 | 00:52:50 | +69,7%
kubernetes-without-olm | minikube | 00:28:35 | 00:54:21 | +79.7%
optional-annotations | minikube | 00:02:04 | 00:03:18 | +59.7%
supported-operators | minikube | 00:07:46 | 00:08:54 | +14.6%
workload-resource-mapping | minikube | 00:03:08 | 00:04:58 | +58.5%
4.10-acceptance | OpenShift | 01:27:00 | 01:38:40 | +13.4%
4.11-acceptance | OpenShift | 01:01:50 | 01:22:00 | +32.6%
4.12-acceptance | OpenShift | 01:01:40 | 01:42:50 | +66.8%

sadlerap:
Could we do like what we do in test cleanup and unset the finalizer for each service binding resource?  That might cut down on the time it takes to clean up a namespace.

test/acceptance/features/environment.py
```diff
@@ -49,9 +55,35 @@ def before_all(_context):
...
     assert code == 0, f"Checking connection to OS cluster by getting the 'default' project failed: {output}"
+
+
+def after_scenario(_context, scenario):
+    if "namespace" in _context:
+        namespace = _context.namespace.name
+    elif os.getenv("TEST_NAMESPACE"):
+        namespace = os.getenv("TEST_NAMESPACE")
+    else:
+        print("No namespace set in context nor TEST_NAMESPACE env variable is set, skipping deletion")
```
sadlerap, line 73: Q: If there's no namespace, this print statement makes it sound like we won't delete anything.  Shouldn't we do an early return here?

Since there's no namespace, the variable `namespace` never gets set, so each of the cases below will throw a `NameError` (since `namespace` never got set).

test/acceptance/features/environment.py
```diff
@@ -14,12 +14,18 @@
...
 from steps.command import Command
 from steps.environment import ctx
+from behave.model_core import Status
 
 import os
 import semver
 
 cmd = Command()
 
+if os.getenv("DELETE_NAMESPACE") in ["alwaysdelete", "neverdelete", "keepwhenfailed"]:
```
sadlerap, line 24: Nit: I like this better:
```suggestion
if os.getenv("DELETE_NAMESPACE") in ["always", "never", "keepwhenfailed"]:
```
This would also require lines 74 and 76 to be changed.

shruthihub:
> 

i understand that we remove finalizers like https://github.com/redhat-developer/service-binding-operator/blob/master/hack/remove-sbr-finalizers.sh @pmacik , can i get your opinion on the same.

test/acceptance/features/environment.py
```diff
@@ -49,9 +55,35 @@ def before_all(_context):
...
+            delete_current_namespace(namespace)
+        else:
+            ctx.no_scenarios_failed = False
+            print(f"Deleting namespace {namespace} skipped, since one of the scenarios failed.")
+    elif delete_namespace == "neverdelete":
+        print(f"Namespace {namespace} deletion skipped.")
+
+
+def delete_current_namespace(name):
+    output, code = cmd.run(f"{ctx.cli} delete namespace {name} --ignore-not-found")
```
pmacik, line 87: Please consider adding a timeout to the command so that the execution is not prolonged unnecesarily in case of failures to delete the namespace.

pmacik:
> i understand that we remove finalizers like https://github.com/redhat-developer/service-binding-operator/blob/master/hack/remove-sbr-finalizers.sh @pmacik , can i get your opinion on the same.

@shruthihub yes, you can just call the same script from your `delete_current_namespace` function (prior the `delete` command) providing the `namespace` parameter as the `NAMESPACE` env variable for the script.

pmacik:
> Could we do like what we do in test cleanup and unset the finalizer for each service binding resource? That might cut down on the time it takes to clean up a namespace.

@sadlerap That was a good suggestion however the recent test runs show the opposite - running cleanup for the finalizers before the namespace deletion actually adds additional overhead in general as seen in the following table:

Acceptance Testing Job | Environment | Before (with #1204) | After (with this PR) | Overhead - current (original)
--- | --- | --- | --- | ---
kubernetes-with-olm | minikube |  00:31:08 | 00:57:02 | +83.2% (+69,7%)
kubernetes-without-olm | minikube | 00:28:35 | 00:55:30 | +94.2% (+79.7%)
optional-annotations | minikube | 00:02:04 | 00:03:41 | +78.2% (+59.7%)
supported-operators | minikube | 00:07:46 | 00:10:35 | +36.3% (+14.6%)
workload-resource-mapping | minikube | 00:03:08 | 00:05:28 | +74.4% (+58.5%)
4.10-acceptance | OpenShift | 01:27:00 | 01:41:00 | +16.1% (+13.4%)
4.11-acceptance | OpenShift | 01:01:50 | 01:33:50 | +51.6% (+32.6%)
4.12-acceptance | OpenShift | 01:01:40 | 01:23:50 | +36.0% (+66.8%)

make/acceptance.mk
```diff
@@ -62,6 +62,7 @@ test-acceptance: test-acceptance-setup
 	$(Q)TEST_ACCEPTANCE_START_SBO=$(TEST_ACCEPTANCE_START_SBO) \
 		TEST_ACCEPTANCE_SBO_STARTED=$(TEST_ACCEPTANCE_SBO_STARTED) \
 		TEST_NAMESPACE=$(TEST_NAMESPACE) \
+		HACK_DIR=$(HACK_DIR) \
```
pmacik, line 65: This change is not necessary now that you removed the finalizers script. It should not be there

test/acceptance/features/environment.py
```diff
@@ -14,12 +14,19 @@
 
 from steps.command import Command
 from steps.environment import ctx
+from behave.model_core import Status
 
 import os
 import semver
 
+
```
pmacik, line 22: Also unnecessary change. It should not be in the PR

pmacik (changes requested):

sadlerap (approved):
With Pavel's comments, looks good to me.

sadlerap (approved):
/lgtm

pmacik:
@shruthihub @sadlerap The tests are failing because the PR branch is updated by the CI by the recent PR #1206 so the new scenarios were introduced, however the PR images are original, (not rebuilt automatically.

Please rebase your PR branch on top on the current master to fix that

PR #1199, opened 2022-08-10
Fix hidden dependency between scenarios in acceptance tests

pmacik:
Signed-off-by: Pavel Macík <pavel.macik@gmail.com>

Currently, the `Operator` class uses class attributes such as `Openshift.operators_namespace` and `Openshift.olm_namespace` to define what namespace to install the operator in and the namespace for catalog source.

That information is supposed to be local to the operator class but the `Operator` class uses the 
 `Openshift` class attributes (which is shared between all instances of that class) for setting those namespaces for the instances of the operator. That is fine when the namespaces are the default ones (`operators` or `openshift-operators` and `olm` or `openshift-marketplace` respectively).

In order to change the namespace for the operator or the catalog source the operator currently changes the `Openshift.operators_namespace` or the `Openshift.olm_namespace` attributes directly. However, since those are the class attributes they are global and shared among all the operator instances.

So in case there is a sequence of scenarios executed in which one scenario uses an operator to set the namespaces to a non-default values (such as `PerconaMysqlOperator`) that namespace is set globally and so any operator used in any following scenarios that assumes the default namespaces would actually use the ones set by the non-default operator.

Also, most of the other attributes of the `Operator` class are class attributes and so shared between instances.

That causes a hidden dependency between scenarios.

# Changes

This PR:
* Changes the class attributes of `Operator` class to instance attributes so that it belongs only to the particular instance. 
* Introduces `operator_namespace` and `operator_catalog_namespace` attributes to the `Operator` class
* Updates the `Operator` class to use the above instance attributes instead of the global ones.

# Submitter Checklist

### DIFF
--- a/test/acceptance/features/steps/cloudnativepostgresoperator.py
+++ b/test/acceptance/features/steps/cloudnativepostgresoperator.py
@@ -6,12 +6,13 @@ from behave import given
 class CloudNativePostgresOperator(Operator):
 
     def __init__(self, name="cloud-native-postgresql"):
-        self.name = name
-        self.pod_name_pattern = "postgresql-operator-controller-manager.*"
-        self.operator_catalog_source_name = "operatorhubio-catalog"
-        self.operator_catalog_channel = "stable"
-        self.operator_catalog_image = "quay.io/operatorhubio/catalog:latest"
-        self.package_name = name
+        super().__init__(
+            name=name,
+            pod_name_pattern="postgresql-operator-controller-manager.*",
+            operator_catalog_source_name="operatorhubio-catalog",
+            operator_catalog_channel="stable",
+            operator_catalog_image="quay.io/operatorhubio/catalog:latest",
+            package_name=name)
 
 
 @given(u'Cloud Native Postgres operator is running')
--- a/test/acceptance/features/steps/crunchypostgresoperator.py
+++ b/test/acceptance/features/steps/crunchypostgresoperator.py
@@ -7,15 +7,19 @@ from behave import given
 class CrunchyPostgresOperator(Operator):
 
     def __init__(self, name="pgo"):
-        self.name = name
         if ctx.cli == "oc":
-            self.operator_catalog_source_name = "certified-operators"
-            self.package_name = "crunchy-postgres-operator"
+            super().__init__(
+                name=name,
+                package_name="crunchy-postgres-operator",
+                operator_catalog_source_name="certified-operators",
+                operator_catalog_channel="v5")
         else:
-            self.operator_catalog_source_name = "operatorhubio-catalog"
-            self.package_name = "postgresql"
-            self.operator_subscription_csv_version = "postgresoperator.v5.0.5"
-        self.operator_catalog_channel = "v5"
+            super().__init__(
+                name=name,
+                package_name="postgresql",
+                operator_catalog_source_name="operatorhubio-catalog",
+                operator_subscription_csv_version="postgresoperator.v5.0.5",
+                operator_catalog_channel="v5")
 
 
 @given(u'Crunchy Data Postgres operator is running')
--- a/test/acceptance/features/steps/percona_mysql_operator.py
+++ b/test/acceptance/features/steps/percona_mysql_operator.py
@@ -6,19 +6,18 @@ from behave import given
 class PerconaMysqlOperator(Operator):
 
     def __init__(self, name="percona-xtradb-cluster-operator"):
-        self.name = name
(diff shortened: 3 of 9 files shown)


shruthihub (approved):
looks good to me!

baijum (approved):
/lgtm

test/acceptance/features/steps/cloudnativepostgresoperator.py
```diff
@@ -6,12 +6,14 @@
 class CloudNativePostgresOperator(Operator):
 
     def __init__(self, name="cloud-native-postgresql"):
-        self.name = name
-        self.pod_name_pattern = "postgresql-operator-controller-manager.*"
-        self.operator_catalog_source_name = "operatorhubio-catalog"
-        self.operator_catalog_channel = "stable"
-        self.operator_catalog_image = "quay.io/operatorhubio/catalog:latest"
-        self.package_name = name
+        Operator.__init__(
```
sadlerap, line 9:
```suggestion
        super().__init__(
```

baijum:
> /approved /lgtm

@shruthihub These commands should be in separate lines.

test/acceptance/features/steps/crunchypostgresoperator.py
```diff
@@ -7,15 +7,21 @@
 class CrunchyPostgresOperator(Operator):
 
     def __init__(self, name="pgo"):
-        self.name = name
         if ctx.cli == "oc":
-            self.operator_catalog_source_name = "certified-operators"
-            self.package_name = "crunchy-postgres-operator"
+            Operator.__init__(
```
sadlerap, line 11:
```suggestion
            super().__init__(
```

test/acceptance/features/steps/crunchypostgresoperator.py
```diff
@@ -7,15 +7,21 @@
...
+                name=name,
+                package_name="crunchy-postgres-operator",
+                operator_catalog_source_name="certified-operators",
+                operator_catalog_channel="v5")
         else:
-            self.operator_catalog_source_name = "operatorhubio-catalog"
-            self.package_name = "postgresql"
-            self.operator_subscription_csv_version = "postgresoperator.v5.0.5"
-        self.operator_catalog_channel = "v5"
+            Operator.__init__(
```
sadlerap, line 18:
```suggestion
            super().__init__(
```

test/acceptance/features/steps/percona_mysql_operator.py
```diff
@@ -6,44 +6,45 @@
...
 
     def __init__(self, name="percona-xtradb-cluster-operator"):
-        self.name = name
-        if ctx.cli == "oc":
-            self.operator_catalog_source_name = "community-operators"
-        else:
-            self.operator_catalog_source_name = "operatorhubio-catalog"
-        self.operator_catalog_channel = "stable"
-        self.package_name = "percona-xtradb-cluster-operator"
+        Operator.__init__(
```
sadlerap, line 9:
```suggestion
        super().__init__(
```

test/acceptance/features/steps/rabbitmq_operator.py
```diff
@@ -5,13 +5,13 @@
 class RabbitMqOperator(Operator):
 
     def __init__(self, name="rabbitmq-cluster-operator"):
-        self.name = name
+        Operator.__init__(self, name=name)
```
sadlerap, line 8:
```suggestion
        super().__init__(self, name=name)
```

test/acceptance/features/steps/perconamongodboperator.py
```diff
@@ -6,44 +6,45 @@
...
 
     def __init__(self, name="percona-server-mongodb-operator"):
-        self.name = name
-        if ctx.cli == "oc":
-            self.operator_catalog_source_name = "community-operators"
-        else:
-            self.operator_catalog_source_name = "operatorhubio-catalog"
-        self.operator_catalog_channel = "stable"
-        self.package_name = name
+        Operator.__init__(
```
sadlerap, line 9:
```suggestion
        super().__init__(
```

sadlerap (changes requested):
AIUI, if you want to call methods on a base class, you should be going through `super()` to preserve method resolution order.

pmacik:
> AIUI, if you want to call methods on a base class, you should be going through `super()` to preserve method resolution order.
> 
> Other than this, looks good.

Thx @sadlerap, good point. I've used your suggestions, PTAL

PR #1222, opened 2022-09-08
Update chart directory  with service binding operator 1.2.0

Kartikey-star:
Signed-off-by: Kartikey Mamgain <mamgainkartikey@gmail.com>

This PR updates the  chart  directory with service binding operator 1.2.0 source chart.

### DIFF
--- a/charts/service-binding-operator/README.md
+++ b/charts/service-binding-operator/README.md
@@ -19,9 +19,11 @@ The `values.yaml` file contains the following values that can be customized when
 - `image.repository`
 - `image.testRepository`
 - `keepTestResources`
+- `kubernetes.platform`
 
 You can define values for the `image.pullPolicy`, `image.repository`, and `image.testRepository` values. If you are not able to pull an image from the quay.io registry, then copy the image to your own container registry.
 As part of the Helm test, objects such as deployment, service binding resources, and secrets used for testing are deleted. To view them, you can install the chart with the `keepTestResources` flag value set to `true`.
+`kubernetes.platform` is by default set to `vanilla`. The user should specify the platform while installing the chart. The kubernetes platform supported for now are `eks`,`vanilla` and `openshift`.
 
 
 ## Helm Chart Installation 
--- a/charts/service-binding-operator/templates/tests/test-service-binding-operator-connection.yaml
+++ b/charts/service-binding-operator/templates/tests/test-service-binding-operator-connection.yaml
@@ -8,7 +8,7 @@ metadata:
     "helm.sh/hook": test-success
 spec:
   securityContext:
-    runAsNonRoot: false
+    runAsNonRoot: !!bool '{{if eq .Values.kubernetes.platform "openshift"}}true{{else}}false{{end}}'
     seccompProfile:
       type: RuntimeDefault
   containers:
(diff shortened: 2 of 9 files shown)


dperaza4dustbit (approved):
lgtm

Kartikey-star:
@dperaza4dustbit needed to change the test file because we need to set securitycontext.runAsNonRoot to true for openshift and false for vanilla clusters. In order to so the test file,value.yaml and Readme has been updated. Hence i  made the chart version as 1.1.0.

Kartikey-star:
Chart tested on aws eks and openshift cluster.

charts/service-binding-operator/README.md
```diff
@@ -19,9 +19,10 @@ The `values.yaml` file contains the following values that can be customized when
 - `image.repository`
 - `image.testRepository`
 - `keepTestResources`
+- `isOpenshift`
```
pmacik, line 22: In stead of a boolean like "is that openshift?" wouln't it be better to use a form of an enum like `kubernets.runtime` or `kubernetes.platform` with values like `vanilla`, `openshift` and `eks`?

pmacik: That way you could use the conditions like `{{ if eq .Values.kubernetes.platform openshift}}`

Kartikey-star: I get the point but the reason i did it using isOpenshift bool is that it automatically will have just two values true or false and i would prefer installing chart with just a bool flag instead of a string flag. Let me know if it makes sense else will make the change.

pmacik: Ok, but what if you have (at some point in the future) some change that is only for EKS... you would have introduced additional boolean variable `isEks` and then allowed users to set `isOpenshift=true` and `isEks=true` at the same time which would create a potential for errors. Having it mutually excluded decreases the possibility of misuse because this way you have garanteed that the platform flag is IMHO, it doesn't matter if users need to set `isOpenshift=true` (literal `true` is a string in the shell anyway) or `platform=openshift` (`platform`, `runtime`,... whatever) , as long as they need to set it in the CLI.

sadlerap: I'm inclined to agree: if we need to introduce platform-specific logic for multiple platforms, it's best if we use an enum-style type here, instead of a set of bools.

charts/service-binding-operator/templates/tests/test-service-binding-operator-connection.yaml
```diff
@@ -8,7 +8,11 @@ metadata:
     "helm.sh/hook": test-success
 spec:
   securityContext:
+    {{if eq .Values.kubernetes.platform "openshift"}}
+    runAsNonRoot: true
+    {{- else if or (eq .Values.kubernetes.platform "vanilla") (eq .Values.kubernetes.platform "eks")}}
     runAsNonRoot: false
+    {{- end}}
```
pmacik, line 15: @Kartikey-star Could you please try this?

```suggestion
    runAsNonRoot: !!bool '{{if eq .Values.kubernetes.platform "openshift"}}true{{else}}false{{end}}'
```

Kartikey-star:
Chart tested on aws eks and openshift cluster with latest changes

charts/service-binding-operator/templates/tests/test-service-binding-operator-connection.yaml
Kartikey-star: Chart tested on aws eks and openshift cluster with suggested 
changes.

PR #1232, opened 2022-09-23
crds: trim crd structs

sadlerap:
We're spending a lot of time deserializing crds.  As an attempt to reduce time and memory spent here, we should trim down the fields stored in the CRDs we need in the crd controller.

# Changes

/kind bug

# Submitter Checklist


sadlerap:
/hold

Awaiting acceptance & conformance tests; this may not actually get us anything.

pmacik:
@sadlerap I guess it is not our own code that holds the data (while I find it a good idea to limit the amount of data to only what we need)... From what I saw in the cumulative allocation tree it was the `client-go`'s `dynamic` client who does (its JSON unmarshallers/decoders down the line).

PR #1244, opened 2022-09-29
Allow enabling/disabling olm descriptors

sadlerap:
# Changes

This introduces a new flag, `--olm-descriptors`, that toggles whether olm descriptors are used as binding sources.  The default value is `false`, i.e. olm descriptors are not checked by default.  To re-enable, run the following:

```bash
kubectl patch deployments.apps -n service-binding-operator service-binding-operator --type=json -p='[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--olm-descriptor"}]'
```

There's probably some more test coverage I need to do for this, but I'd like to see how performance looks.

/kind enhancement

# Submitter Checklist


pmacik:
/lgtm

@sadlerap The OpenShift acceptance tests fail because the OLM descriptors are not explicitelly enabled for them and only the OLM descriptor related tests fails.

I have identified those scenarios and sent a PR (#1245) to mark them with the `@olm-descriptors` tag so we can disable them on demand (openshift/release/pull/32750) or deprecate it later.

With the second PR I have temporarily disable them because I cannot use the code to patch SBO deployment to enable olm descriptors on SBO like you do until this PR is actually merged. (otherwise other PR's CI would fail complaining that the flag is invalid)

Once the above PRs are merged to the respecive repos, you can rebase your PR and the OpenShift acceptance tests should pass again.

.github/workflows/pr-checks.yaml
```diff
@@ -107,14 +107,34 @@ jobs:
...
+          make SKIP_REGISTRY_LOGIN=true -o registry-login deploy-from-index-image
+
+      - name: Patch SBO to accept OLM annotations
+        if: ${{ steps.check-skip-acceptance.outputs.can_skip != 'true' }}
+        run: |
+          echo "##[group]kubectl describe deployments.apps -n operators service-binding-operator"
+            kubectl describe deployments.apps -n operators service-binding-operator
+          echo "##[endgroup]"
+          echo "##[group]Patch operator"
+            kubectl patch deployments.apps -n operators service-binding-operator --type=json -p='[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--olm-descriptor"}]'
```
pmacik, line 126: @sadlerap are you sure this change to the deployment is not reverted back by the OLM immediatelly after it is patched?

main.go
```diff
@@ -86,11 +86,15 @@ func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
+	var enableOLMDescriptor bool
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&enableOLMDescriptor, "olm-descriptor", false,
```
pmacik, line 95: I have doubts with a cli parameters to be used to set this flag.

What about using a config map to store such flag that SBO would look for it at start and used the values from the CM. If it does not exist, SBO would create it automatically with the default values.

This would allow setting this kind of flags at install time without the need to change the actual Deployment objects (like in case of havint a binary parameter for `--olm-descriptors`). The installation would then go like... Create a ConfigMap in the operator's namespace of a particular name with `enableOLMDescriptor: true` data in it (if users want to enable it) or leave it as before to disable it by default.

sadlerap: Yeah, I'm not sure I like the interface; a `ConfigMap` probably would work better.  Are they the conventional way of providing configuration options for an operator?

pmacik: Well, it is a conventional (a "cloud-native") way in Kubernetes in general.

sadlerap:
Closing in favor of #1251.

PR #1278, opened 2022-10-19
Revert "Bump version of Go to v1.18 and fix CVE-2022-32149"

pmacik:
Reverts redhat-developer/service-binding-operator#1275

Let's fix the CVE first and then bump Go v1.18 later


pmacik:
The security vulnerability scan failure is expected.

PR #1326, opened 2022-12-20
ci: pipeline actions

sadlerap:
We should use explicit dependencies for CI steps that need it.  This lets us save CI minutes by cutting out the wait-for-push steps that most acceptance testing steps need.

# Changes

# Submitter Checklist


pmacik:
@sadlerap The workflow is intentionally split into Build(pull_request) -> Push(pull_request_target) -> PR_checks(pull_request) because we need a different context for each workflow... Build and PR_checks workflows runs in the `pull_request` context (because we want to build the images and test with the PR changes) while the Push workflow needs to run in the `pull_request_target` context to get access to the GH secrets (where credentials for pushing images to quay.io are taken from securely). (See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)

"wait-to-build" and "wait-to-push" is a way of synchronizing the workflows. Each workflow is executed in close-to-random order and we need to make sure the tests that use the images runs after they are pushed to quay, and also that the images are pushed only after they are built. The synchronization does not add too much overhead as the "wait-for-*" steps have 60s interval of re-trying while waiting so with 2 synchronizations (Build->Push, and Push->PR_checks) the overhead is ~2-3minutes in the worst case.

